PR #76, opened 2012-02-11
Fixed window size changes when tab bar appears or disappears

RhodanV5500:
Currently the window changes its height when the tab bar appears or disappears. Now the overall window height remains the same when adding or removing the tab bar.

Its not 100% perfect though. The tab bar height is hardcoded to 22 pixels and you can see the resize animation when you look closely.

Takes "Hide tab bar when there is only one tab" option into account.

Hope you find this useful :)

### DIFF
--- a/PseudoTerminal.m
+++ b/PseudoTerminal.m
@@ -831,6 +831,22 @@ NSString *sessionsKey = @"sessions";
     [self closeTab:aTab soft:NO];
 }
 
+- (void)adjustWindowHeightBy:(float)heightChange 
+{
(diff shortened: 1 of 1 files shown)


khalidchawtany:
I would love to have an option to hide tab bar permanently **in and out** of full screen mode with the ability of pressing **command key** to see the tab pane temporarily.

liuchong:
Now we can check the option `preserve window size when tab bar shows or hides`
preserve_window_size_when_tab_bar_shows_or_hides

PR #118, opened 2013-03-22
Force new splits into their own new window when Option key is held down.

JohnSpeno:
I wanted to resubmit this patch that adds a new feature to the Sessions window. If a user holds down the Option key while pressing either of the Split buttons, then the selected sessions will open in a new window split in the requested way. I use this nearly each time I use iTerm. If there are required changes, let's discuss them here and alter the pull request as needed. Thanks.


JohnSpeno:
Yeah, it's been a while, and I've been missing this feature every day since then. 😆 

Please do consider this feature for review. Thanks, George!

gnachman:
Welcome back! Landed as commit c2fa2d05a1cb1b17e33f4e3602cd5bdf9cf3ba69. Let's try to turn things around faster next time 🤣

PR #147, opened 2013-11-06
Add a keyboard shortcut option to toggle xterm mouse-reporting

drpowell:
Using iTerm2 with tmux is great.  However, I wanted a quick way to enable/disable mouse reporting.  This patch adds a option that can be bound to a keyboard shortcut


aaron-pavlus:
I have been looking for this feature as well; in regular Mac terminal Cmd+R will toggle mouse reporting, which is handy when using tmux to turn it off real quick in order to copy something to the Mac clipboard.   

I know there are other solutions, but they can be complicated to set up, especially to get working correctly with a remote tmux session; so quick-toggle of mouse reporting is very helpful as a fallback.

I know this is an old pull request; but it's the best thing I found when trying to google for this.

dubaaron:
I have been looking for this feature as well; in regular Mac terminal Cmd+R will toggle mouse reporting, which is handy when using tmux to turn it off real quick in order to copy something to the Mac clipboard.

I know there are other solutions, but they can be complicated to set up, especially to get working correctly with a remote tmux session; so quick-toggle of mouse reporting is very helpful as a fallback.

I know this is an old pull request; but it's the best thing I found when trying to google for this.

gnachman:
Yikes! I feel bad that I forgot about this PR for so long. Merged.

PR #152, opened 2013-12-11
Adds option to clone a session by doubleclicking the tab (issue #1450, #...

fenderle:


epileftro85:
in linux if you type ctrl + shift + t clone the current tab, it would b nice if something similar is implemented in iterm an keyboard combination instead clicks.

PR #164, opened 2014-03-13
Add hidden setting (AlternateMouseScroll) for arrow keys on mouse scroll

FiloSottile:
The hidden defaults setting AlternateMouseScroll turns on "mouse wheel scrolling support when a terminal is in MOUSE_REPORTING_NONE mouse mode and is using an alternate screen by translating them to arrow up/down terminfo escape sequences. (ala MouseTerm)"

Activate it with

```
$ defaults write com.googlecode.iterm2 AlternateMouseScroll -bool true
```

This patch was submitted and adapted in https://code.google.com/p/iterm2/issues/detail?id=974 and then wrapped in a hidden setting by me

### DIFF
--- a/PTYTextView.m
+++ b/PTYTextView.m
@@ -2869,18 +2869,16 @@ NSMutableArray* screens=0;
                 break;
             case MOUSE_REPORTING_NONE:
                 if ([[PreferencePanel sharedInstance] alternateMouseScroll] &&
-                    terminal.isAlternate) {
+                    [_dataSource isAlternate]) {
                     CGFloat deltaY = [event deltaY];
+                    NSData* keyMove;
                     if (deltaY > 0) {
-                        NSData* keyMove = [terminal.output keyArrowUp:[event modifierFlags]];
-                        for (int i = (int)ceil(deltaY); i > 0; --i) {
-                            [_delegate writeTask:keyMove];
-                        }
+                        keyMove = [terminal.output keyArrowUp:[event modifierFlags]];
                     } else if (deltaY < 0) {
-                        NSData* keyMove = [terminal.output keyArrowDown:[event modifierFlags]];
-                        for (int i = (int)ceil(-deltaY); i > 0; --i) {
-                            [_delegate writeTask:keyMove];
-                        }
+                        keyMove = [terminal.output keyArrowDown:[event modifierFlags]];
+                    }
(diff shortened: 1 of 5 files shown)


mtsgrd:
I can't seem to make this work. I've run `defaults write com.googlecode.iterm2 AlternateMouseScroll -bool true`, is there anything else I need to pay attention to before trying to scroll a file in vim?

adriaant:
OP wrote: "turns on "mouse wheel scrolling support when a terminal is in MOUSE_REPORTING_NONE mouse mode and is using an alternate screen by translating them to arrow up/down terminfo escape sequences. "
so check if your terminal is in MOUSE_REPORTING_NONE and if you have arrow up/down terminfo escape sequences.

afletch:
Is this in the latest beta?

gnachman:
No, nightly only. The 2.9 branch hasn't gone beta yet because I've been distracted by having a baby :) I hope to have enough time to release a 2.9 beta this summer.

ebigelow:
adriaant commented on Jan 20:
"so check if your terminal is in MOUSE_REPORTING_NONE and if you have arrow up/down terminfo escape sequences."

I apologize for my naivety with iterm, but can someone advise me on how to do this? 

In the iterm preferences there's a checkbox for 'enable xterm mouse reporting' -- should this be unchecked? I'm also not sure what escape sequence to set for the up/down arrows, since 'ku' & 'kd' don't seem to work, and neither do 'up' & 'do'.

gnachman:
@ebigelow These are the conditions that enable this feature:
1. Run the nightly build of iTerm2
2. Ensure *enable xterm mouse reporting* is **on**
3. Ensure *disable save/restore alternate screen* is **off**
4. Run a program that switches into alternate screen mode (less, vi, emacs, etc.; these programs restore the screen contents upon exiting)
5. Mouse reporting should not be in use by that program (you can tell because the I-bar cursor looks like [this](https://github.com/gnachman/iTerm2/blob/master/images/IBarCursor%402x.png), not like [this](https://github.com/gnachman/iTerm2/blob/master/images/IBarCursorXMR%402x.png))
6. Use the scroll bar

ebigelow:
Silly mistake on my part, I wasn't using the nightly build -- I assumed this had been integrated into the stable release by now. Thank you for the prompt and thorough response!

PR #174, opened 2014-07-24
Update icon

fabiopelosin:
With Yosemite it could be a good opportunity to refresh the icon

Light theme:

screen shot 2014-07-24 at 10 08 07

Dark theme:

screen shot 2014-07-24 at 10 07 53

Link to the sketch file used to design it: http://cl.ly/2W2Y0r430v0Q


anhari:
Any chance we could reconsider @jasonlong's icon?

gnachman:
@jasonlong are you still around? I'd like to revisit this. Can you provide a 1024x1024 graphic with layers? Specifically, I need the $ and cursor in their own layer(s). By incorporating this graphic it will be licensed under GPL v2 or later.

jasonlong:
@gnachman :wave: still here!

Since my comment above, I've updated my version of the icon after being inspired by the dark header feature.

icon_128x128 2x

I can get you the old design plus any tweaks you'd like (perhaps the stop light circles in the header?). I assume the green prompt and cursor should remain since it's a brand element.

:thumbsup: on GPL licensing.

gnachman:
I think I prefer the old design as it was—really minimalistic and with just the right branding.

jasonlong:
@gnachman Are you able to work with a Sketch file? If so, the original file is available [here](https://github.com/jasonlong/iterm-replacement-icon/blob/master/iterm2-yosemite-icon.sketch). If not, let me know what format you'd prefer.

gnachman:
Hm, seems a little off when I put it in the dock. I think it doesn't downscale as well as the existing icon. Compare:

screen shot 2018-04-06 at 4 08 06 pm

I think the $ and cursor need to be bigger, and there's something about the titlebar that makes it seem to be cut off on the sides.

gnachman:
The repo still exists on github. I think the best thing to do is to fork it, add the .sketch file, and send a PR for that.

PR #183, opened 2014-10-16
Floating hotkey window

ponychicken:
This is based on an old patch of yours and fixes https://code.google.com/p/iterm2/issues/detail?id=1570

The weird behauvior you describe is nowhere to be seen, at least not in yosemite. I added some code which sucessfully detected whether the hotkeywindow was over an fullscreen app, but when I couldn't figure out how to move the window up in this case, as my origin.y was seemingly reset everytime. Maybe you know where that's coming from.


AtomicCoding:
Is this still happening?

gnachman:
A change along these lines was merged in 3.1. You can enable floating panels in **Prefs > Profiles > Keys > Configure Hotkey Window**. I'll close this out.

PR #194, opened 2015-02-01
Fix korean encodings

khwon:
I found out that this change is still in `fixKoreanEncodings` branch.
I've compiled and checked that it works with CP949 encodings well.
Is there any further progress required to merge this into master branch?


gnachman:
Sorry, I must have forgotten the last step. Thanks for catching it.

PR #195, opened 2015-02-03
Make cursor opaque

allewun:
Not sure if this was an intentional design decision, but having a transparent cursor can be a bit distracting when content bleeds through it. The text doesn't use transparency, so I figured the cursor should be the same.

Here's an animation showing the change:

iterm_animation


gnachman:
I agree this looks better. Thanks.

PR #196, opened 2015-02-03
Rename Bundle Name to “iTerm” from “iTerm2”

davidbarsky:
There are two reason for this pull request:

1. More consistent naming between what the app presents in the Finder/Dock and what is presented in the menu bar.
2. More subjectively, the absence of numbers in the menu bar is preferable. Very few Mac applications have a numeral in Bundle name.


gnachman:
The app's name, for better or worse, is iTerm2. The naming consistency issue is hopefully temporary until I can sort out how to change it for folks who auto-update.

PR #197, opened 2015-02-11
Change reset shorcut from CMD+r to CMD+R

saga64:
Developers who use vim under Mac may accidently press CMD+r when
they try to refresh the web browser when the focus was still on
vim window, it will break the buffer on vim, this change will make
user to use CMD+SHIFT+R, which could somehow resolve the conflicts


gnachman:
I'm very hesitant to change shortcuts which have been around for so long. If it's an issue for you, you can remap cmd-r to "Ignore" in prefs>keys.

PR #198, opened 2015-02-15
Show main menu when right clicking empty tabBar region

allewun:
When using the "[LSUIElement hack](https://code.google.com/p/iterm2/issues/detail?id=954)" to hide dock icon, the menu bar also disappears, leaving users with no way to reach the main menu.


gnachman:
I'm not crazy about this since it's a pretty dramatic departure from standard behavior, but I understand why it's useful. How about enabling this feature only when the LSUIElement hack is on?

allewun:
Good idea, I'll see if I can figure that out.

allewun:
Ok, please have a look again.

gnachman:
Looks good, thanks!

PR #199, opened 2015-02-15
Added bindable key action to quickly toggle hotkey window pinning

allewun:
Resolves [issue 3413](https://code.google.com/p/iterm2/issues/detail?id=3413)


gnachman:
I'm not persuaded that this feature is important enough to justify a menu item, but I'd be ok with it being a bindable key action.

allewun:
I've implemented this as a bindable key action. One additional change I made was to have the hotkey window use `NSFloatingWindowLevel` when it loses focus and we don't autohide (aka "pinning"). Pinning wouldn't be very useful otherwise.

gnachman:
Window level stuff is fraught with peril; let's leave that for another PR. I believe you will find endless issues with, for example, popup windows (e.g., autocomplete). Otherwise it looks good.

allewun:
I tested popup windows and it worked fairly well for me, but I agree that the window stuff can get messy. Window level changes have been reverted, and I've squashed my commits.

allewun:
@gnachman please have another look when you have a chance

gnachman:
Thanks and sorry for the delay, looks good.

PR #201, opened 2015-03-12
Background transparency

nathanrobinson:
This fork allows setting different transparency and blur values for background windows.


gnachman:
Wow, thanks for the big change. It will take me a little bit to get through this, but I am taking a look.

gnachman:
I haven't dug in super deep but I think these comments cover most of the issues:

- The transparency for background windows doesn't affect the margins (they seem to always use the foreground setting)
- The prefs dialog is starting to look (more) like the controls on the space shuttle. How about moving the blur and transparency controls into a modal sheet, like the settings in the prefs>advanced panel? Leave behind a button labeled "Transparency and Blur…" that opens the sheet.
- I saw a few places where "bool" and "true" and "false" were used instead of BOOL and YES/NO; looks like some old wrong code. Instead of being consistent, would you mind fixing those?
- There's an "Inactivate transparency" label below the "Inactive transparency" checkbox. The label should be removed.
- Don't use abbreviations except in extreme cases. KEY_BACKGROUND_TRANS and KEY_USE_BACKGROUND_TRANS should spell out TRANSPARENCY, even if it messes up the alignment of the source code.
- Names in code (for example, useBackgroundTransparency) should match what we see in the UI (where “inactive” is used, not “background”).
- Minor style issue: new code should use this format for methods:

   ` - (void)foo {`

rather than

    - (void)foo
    {

- Always put a space between an `if` statement and the opening paren of its condition (e.g., in `recheckBlur`)

- The use of Hungarian notation is being phased out. Avoid names like "fVal" (e.g., in `setBackgroundTransparency:`). In the case of setters, the preferred name matches the property name (so, `- (void)setBackgroundTransparency:(double)backgroundTransparency {` would be correct).
- Why the `+ 10` in `-[ProfilePreferencesViewController awakeFromNib]`? Looks fine without it to me.
- The `/////// Background transparency` comments in ProfilesWindowPreferencesViewController.m should be removed or improved (if you keep them, don't use lots of /s, just the two that are required)
- In general, avoid lines over 100 characters, unless it would compromise readability. For example, in ProfilesWindowPreferencesViewController.m:94

nathanrobinson:
Thanks for the feedback. 
This is my first foray into objective C, so for the first two bullet items I am completely lost. I will try to track them down, though. 
The + 10 is there because, at least on my screen, when a preferences screen scrolls off the bottom of the window, a scroll bar is added to the view which covers up part of the controls.
This would fix itself if I am able to move the transparency and blur controls to a popup.
The rest should be easy to fix.

gnachman:
You're doing well for a first foray! Feel free to ignore the second comment, I'll take care of that. As for the margins not being correct, I think it's because the wrong alpha value is picked in -[PTYSession textViewDrawBackgroundImageInView:viewRect:blendDefaultBackground:].

I think the +10 should stay as the advanced prefs actually look pretty ugly with always-visible scrollbars, but instead of always adding 10, add `[NSScroller scrollerWidthForControlSize:NSRegularControlSize scrollerStyle:NSScrollerStyleLegacy]`. That'll always be correct even if apple changes how it's rendered (plus it explains what's going on). I always forget to test that configuration so I'm glad you caught it.

nathanrobinson:
I think I got them all except the popup for transparency sliders on the preferences dialog.

sources/PTYSession.m
```diff
@@ -801,6 +801,8 @@ - (BOOL)setScreenSize:(NSRect)aRect parent:(id<WindowControllerInterface>)parent
      horizontalSpacing:[[_profile objectForKey:KEY_HORIZONTAL_SPACING] floatValue]
        verticalSpacing:[[_profile objectForKey:KEY_VERTICAL_SPACING] floatValue]];
     [self setTransparency:[[_profile objectForKey:KEY_TRANSPARENCY] floatValue]];
+    [self setInactiveTransparency:[[_profile objectForKey:KEY_INACTIVE_TRANSPARENCY] floatValue]];
```
gnachman: New code should not look up values in _profile directly because it bypasses the relatively new mechanism for defaults. Older code generally had default values of 0 so it happens to work, and I'm slowly moving everything the new model. Instead, write:
`[iTermProfilePreferences objectForKey:KEY_INACTIVE_TRANSPARENCY inProfile:_profile]`

sources/PTYTab.m
```diff
@@ -1855,21 +1855,16 @@ - (NSSize)_minSizeOfView:(NSView*)view
...
+- (BOOL)blur {
     NSArray* sessions = [self sessions];
     for (PTYSession* session in sessions) {
         if ([session transparency] > 0 &&
             [[session textview] useTransparency] &&
             [[[session profile] objectForKey:KEY_BLUR] boolValue]) {
-            ++y;
-        } else {
-            ++n;
+            return YES;
```
gnachman: Nice!

sources/PTYTab.m
```diff
@@ -1891,15 +1886,66 @@ - (double)blurRadius
     }
 }
 
-- (void)recheckBlur
-{
+- (BOOL)useInactiveTransparency {
+    NSArray* sessions = [self sessions];
```
gnachman: New code should look like `NSArray *sessions` with the space before the *.

sources/PTYTab.m
```diff
@@ -1891,15 +1886,66 @@ - (double)blurRadius
...
+    }
+    return NO;
+}
+
+- (BOOL)inactiveBlur {
+    NSArray* sessions = [self sessions];
+    for (PTYSession* session in sessions) {
+        if ([session inactiveTransparency] > 0 &&
+            [[session textview] useInactiveTransparency] &&
+            [[[session profile] objectForKey:KEY_INACTIVE_BLUR] boolValue]) {
```
gnachman: Better to define and use session.inactiveBlur than to peek into the profile.

sources/PTYTab.m
```diff
@@ -1891,15 +1886,66 @@ - (double)blurRadius
...
+            return YES;
+        }
+    }
+    return NO;
+}
+
+- (double)inactiveBlurRadius {
```
gnachman: There's a lot of duplicated code with -blurRadius. Please refactor into a common method like
`- (double)averageBlurRadiusForInactive:(BOOL)inactive`

sources/PTYTextView.h
```diff
@@ -199,11 +199,16 @@ typedef enum {
 // Transparency level. 0 to 1.
 @property(nonatomic, assign) double transparency;
 
+// Transparency level. 0 to 1.
```
gnachman: *Inactive window* transparency level. 0 to 1.

sources/PTYTextView.h
```diff
@@ -317,6 +322,9 @@ typedef enum {
 // Updates the preferences for semantic history.
 - (void)setSemanticHistoryPrefs:(NSDictionary *)prefs;
 
+// Is the main window active?
+- (void)setIsBackground:(BOOL)isBackground;
```
gnachman: I'm trying to break direct dependencies between `PseudoTerminal` and `PTYTextView`. Rather than have it call `setIsBackground:` directly from `windowDidBecomeKey:` and `windowDidResignKey:`, we can use notifications to keep this decoupled.

You can register for notifications like this, in -[PTYTextView initWithFrame:colorMap:]:

      [[NSNotificationCenter defaultCenter] addObserver:self
                                               selector:@selector(windowDidBecomeKey:)
                                                   name:NSWindowDidBecomeKeyNotification
                                                 object:nil];

      [[NSNotificationCenter defaultCenter] addObserver:self
                                               selector:@selector(windowDidResignKey:)
                                                   name:NSWindowDidResignKeyNotification
                                                 object:nil];

And add methods like these:

      - (void)windowDidBecomeKey:(NSNotification *)notification {
          if ([notification object] == [self window]) {
              [self setIsBackground:NO];
          }
      }

      - (void)windowDidResignKey:(NSNotification *)notification {
          if ([notification object] == [self window]) {
              [self setIsBackground:YES];
          }
      }

And then -setIsBackground can be a private method.

sources/PTYTextView.h
```diff
@@ -419,5 +427,11 @@ typedef enum {
                          prefix:(NSString *)prefix
                          suffix:(NSString *)suffix;
 
+// Get the transparency value for foreground or background windows.
```
gnachman: I think the comment means it gets the transparency value for this view, which may vary depending on whether the window is active or not. Can you clarify?

sources/PTYTextView.m
```diff
@@ -6376,7 +6397,11 @@ - (BOOL)_drawLine:(int)line
     int WIDTH = [_dataSource width];
     screen_char_t* theLine = [_dataSource getLineAtIndex:line];
     BOOL hasBGImage = [_delegate textViewHasBackgroundImage];
+    
     double selectedAlpha = 1.0 - _transparency;
+    if(_isBackground){
```
gnachman: Always put a space after `if`.

sources/PTYTextView.m
```diff
@@ -7163,7 +7188,13 @@ - (NSColor *)smartCursorColorForChar:(screen_char_t)screenChar
 }
 
 - (double)transparencyAlpha {
-    return [self useTransparency] ? 1.0 - _transparency : 1.0;
+    if([self useTransparency]){
```
gnachman: Needs a space after `if`.

sources/ProfilesWindowPreferencesViewController.m
```diff
@@ -70,6 +74,31 @@ - (void)awakeFromNib {
...
+                    key:KEY_INACTIVE_TRANSPARENCY
+                   type:kPreferenceInfoTypeSlider];
+    
+    info = [self defineControl:_useInactiveBlur
+                           key:KEY_INACTIVE_BLUR
+                          type:kPreferenceInfoTypeCheckbox];
+    info.observer = ^() {
+        _inactiveBlurRadius.enabled =
+        (_useInactiveTransparency.state == NSOnState
+         && _useInactiveBlur.state == NSOnState);
```
gnachman: Move the `&&` to the end of the preceding line.

sources/PseudoTerminal.m
```diff
@@ -5596,6 +5603,7 @@ - (void)setDimmingForSession:(PTYSession *)aSession
         // broadcasting to the current session.
         [[aSession view] setDimmed:YES];
     }
+    [[aSession tab] recheckBlur];
```
gnachman: This method is only meant to adjust dimming. I see that it gets called from several places, so to avoid having recheckBlur calls littered all over the place, the simplest thing might be to rename it from `setDimmingForSession:` to `updateDimAndBlurForSession:`, and rename callers as needed (e.g., `setDimmingForSessions`). XCode's "refactoring" feature makes this pretty easy to do.

sources/SessionView.m
```diff
@@ -164,6 +164,8 @@ - (void)setDimmed:(BOOL)isDimmed {
     } else {
         _dim = isDimmed;
     }
+    
+    [[[self session] tab] recheckBlur];
```
gnachman: Is this really needed? It only has two callers, `setDimmingForSession:` (which already calls `recheckBlur`) and `-[PTYTab initWithSession:]`. It could be moved into `initWithSession:`, at least.

sources/SessionView.m
```diff
@@ -176,6 +178,8 @@ - (void)setBackgroundDimmed:(BOOL)backgroundDimmed {
     if (_backgroundDimmed != orig) {
         [self updateDim];
     }
+    
+    [[[self session] tab] recheckBlur];
```
gnachman: This is called from four places in `PseudoTerminal`. I'm not sure it's necessary to do this in all four cases--in fact, I believe only `_refreshTerminal` actually needs it. Can you move it up into the necessary call-sites? I'd like to keep this function true to its name.

gnachman:
I added a bunch of niggling line notes, mostly style stuff. Thanks for taking the time to do this!

nathanrobinson:
Well, I think I got it, but you'll want to double check me.

sources/PTYSession.m
```diff
@@ -2669,6 +2673,42 @@ - (void)setTransparency:(float)transparency
...
+
+- (BOOL)blur {
+    return [iTermProfilePreferences boolForKey:KEY_BLUR inProfile:_profile];
+}
+
+- (float)inactiveBlurRadius {
+    return [iTermProfilePreferences floatForKey:KEY_INACTIVE_BLUR_RADIUS inProfile:_profile];
+}
+
+- (float)blurRadius {
```
gnachman: Both float and double are used here; I'm sure that's due to some ancient folly. Let's use double for all new code.

sources/PTYTab.m
```diff
@@ -1855,51 +1856,86 @@ - (NSSize)_minSizeOfView:(NSView*)view
...
+    return [self averageBlurRadiusForInactive:YES];
+}
+
+- (void)recheckBlur {
     PtyLog(@"PTYTab recheckBlur");
     if ([realParentWindow_ currentTab] == self &&
         ![[realParentWindow_ window] isMiniaturized]) {
-        if ([self blur]) {
-            [parentWindow_ enableBlur:[self blurRadius]];
+        if ([[[self realParentWindow] window] isKeyWindow]){
```
gnachman: Needs a space between `)` and `{`

sources/PTYTextView.m
```diff
@@ -5221,6 +5231,27 @@ - (void)setTransparency:(double)fVal
     [self setNeedsDisplay:YES];
 }
 
+- (void)setInactiveTransparency:(double)inactiveTransparency
+{
```
gnachman: New code should put the `{` starting a method on the preceding line.

sources/PTYTextView.m
```diff
@@ -5221,6 +5231,27 @@ - (void)setTransparency:(double)fVal
...
+}
+
+- (void)setUseInactiveTransparency:(BOOL)useInactiveTransparency
+{
+    _useInactiveTransparency = useInactiveTransparency;
+    [_colorMap invalidateCache];
+    [self setNeedsDisplay:YES];
+}
+
+- (void)setIsBackground:(BOOL)isBackground
```
gnachman: For consistency, this should be setIsInactive: and the backing variable should be _isInactive.

sources/ProfilePreferencesViewController.m
```diff
@@ -155,7 +155,12 @@ - (void)awakeFromNib {
         static const CGFloat kMaxHeight = 438;
         if (view.frame.size.height > kMaxHeight) {
             // If the view is too tall, wrap it in a scroll view.
-            NSRect theFrame = NSMakeRect(0, 0, view.frame.size.width, kMaxHeight);
+            
+            CGFloat widthWithScrollBar = view.frame.size.width +
+            [NSScroller scrollerWidthForControlSize:NSRegularControlSize scrollerStyle:NSScrollerStyleLegacy];
```
gnachman: Break this into two lines, aligning the colons (when readability isn't hurt, lines are kept under 100 characters)

sources/ProfilesWindowPreferencesViewController.m
```diff
@@ -70,6 +74,31 @@ - (void)awakeFromNib {
...
+    [self defineControl:_inactiveTransparency
+                    key:KEY_INACTIVE_TRANSPARENCY
+                   type:kPreferenceInfoTypeSlider];
+    
+    info = [self defineControl:_useInactiveBlur
+                           key:KEY_INACTIVE_BLUR
+                          type:kPreferenceInfoTypeCheckbox];
+    info.observer = ^() {
+        _inactiveBlurRadius.enabled =
+        (_useInactiveTransparency.state == NSOnState &&
```
gnachman: Indent this line and the next by 4 more spaces.

sources/PseudoTerminal.m
```diff
@@ -5596,13 +5597,14 @@ - (void)setDimmingForSession:(PTYSession *)aSession
...
     [[aSession view] setNeedsDisplay:YES];
 }
 
 - (void)setDimmingForSessions
```
gnachman: This should be renamed to `updateDimAndBlurForSessions` to reflect its change.

sources/PseudoTerminal.m
```diff
@@ -5686,11 +5688,14 @@ - (void)_refreshTerminal:(NSNotification *)aNotification {
 
         // Update dimmed status of inactive sessions in split panes in case the preference changed.
         for (PTYSession* aSession in [aTab sessions]) {
-                        [self setDimmingForSession:aSession];
+                        [self updateDimAndBlurForSession:aSession];
```
gnachman: Would you mind fixing the indentation of this line while you're in here?

sources/PseudoTerminal.m
```diff
@@ -5686,11 +5688,14 @@ - (void)_refreshTerminal:(NSNotification *)aNotification {
...
             [[aSession view] setBackgroundDimmed:![[self window] isKeyWindow]];
 
             // In case dimming amount slider moved update the dimming amount.
             [[aSession view] updateDim];
+            
+
```
gnachman: Please remove one of the empty lines (two consecutive newlines aren't used except between methods)

sources/PTYTab.m
nathanrobinson: Does Xcode have a setting to make spacing consistent? I'm used to having reshaper do that for me in visual studio.

gnachman: Not that I'm aware of. XCode is bad in surprising ways.

nathanrobinson:
I [tested out](https://github.com/nathanrobinson/iTerm2/commit/e64105e9310a40be6a83b7a507dac4615657bc81) appCode’s reformatting and wanted to get your thoughts. It
handles most of your styles, but seems to add extra indents on some
multi-line constructs.

sources/PTYTab.m
```diff
@@ -1855,51 +1856,86 @@ - (NSSize)_minSizeOfView:(NSView*)view
...
-    return y > 0;
+    return NO;
 }
 
-- (double)blurRadius
-{
+- (double)averageBlurRadiusForInactive:(BOOL)inactive {
     double sum = 0;
     double count = 0;
     NSArray* sessions = [self sessions];
```
gnachman: Space before *, also on next line

sources/PTYTextView.h
```diff
@@ -199,11 +199,16 @@ typedef enum {
...
+// Inactive Window Transparency level. 0 to 1.
+@property(nonatomic, assign) double inactiveTransparency;
+
 // Blending level for background color over background image
 @property(nonatomic, assign) double blend;
 
 // Should transparency be used?
 @property(nonatomic, readonly) BOOL useTransparency;
+@property(nonatomic, readonly) BOOL useInactiveTransparency;
+@property(nonatomic, readonly) BOOL windowInactive;
```
gnachman: Since this isn't used externally it can be removed.

gnachman:
Please revert the automatic style changes to PTYTextView and let's revisit that later. I'm working on a significant refactor of that file in a feature branch and this will introduce thousands of merge conflicts.

nathanrobinson:
I'm not sure what happened, but now the original blur radius settings aren't working. I'll see if I can track that down.

PR #202, opened 2015-03-20
Use handleEvent instead of interpretKeyEvents.

mzp:
This patch is related with https://code.google.com/p/iterm2/issues/detail?id=2629 but not same.

## What steps will reproduce the problem?
1. Install AquaSKK.
2. Give focus a window of iTerm2 and make AquaSKK enable.
3. Change the mode to Japane Hirakana Input mode.
4. Press l to turn on ASCII input mode.

## What is the expected output?
The [A] icon with orange background is displayed at the location by the cursor in an iTerm window, and any characters are not inputted in the iTerm window.

## What do you see instead?
The [あ] icon with orange background is displayed at the location by the cursor in an iTerm window, and a newline character is inputted in the iTerm window.

## What OS version are you using?
MacOS X 10.10.2

## Solution
Because some IME(e.g. AquaSKK) consume event without insert any text, `_inputMethodIsInserting` is not enough.

So, we need use handleEvent instead of interpretKeyEvents.

### DIFF
--- a/sources/PTYTextView.m
+++ b/sources/PTYTextView.m
@@ -116,6 +116,7 @@ static const int kBadgeRightMargin = 10;
     BOOL _nonasciiAntiAlias;  // Only used if self.useNonAsciiFont is set.
 
     // NSTextInputClient support
+    NSEvent* _keydownEvent;
     BOOL _inputMethodIsInserting;
     NSRange _inputMethodSelectedRange;
     NSRange _inputMethodMarkedRange;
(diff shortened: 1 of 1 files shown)


gnachman:
I have to warn you that this is going to be a very hard change to do correctly. This code is very complex.

I don't know why, but as written, it breaks key repeat. If you press and hold "x" (as an example) with a US keyboard, `[self.inputContext handleEvent:event]` returns YES and `[delegate keyDown:event]` never gets called.

gnachman:
Fixed in commit 71aaf5939f73023691ba6b6759f154f4901d12ba

PR #203, opened 2015-03-22
Update preference pane icons for Yosemite

peteschaffner:
I love the current pref icons by @kaishin, but they were starting to feel a little out-of-date on Yosemite, so I riffed on them a bit. Let me know your thoughts.

screen shot 2015-03-22 at 8 48 59 pm


gnachman:
Awesome! This is a much flatter look. The only thing that doesn't look quite right to me is the green background on Profiles. It draws too much attention, IMO. I love the green cursor in Appearance, though; it's a subtle callback to the icon.

Would you mind adding the source files you used (from Photoshop, Sketch, etc.) in case we need to tweak them later (e.g., add 3x assets).

peteschaffner:
Yeah I agree, all that green was a bit much. I was inspired by the blue "@" account icons in a lot of Yosemite pref panels, but I think having a more uniform set feels better. Let me know what you think about this grey version.

image

peteschaffner:
I cheated a  bit with the advanced icon, and just utilized the 64 and 32x32 cuts of the ToolbarAdvanced.icns from the CoreTypes.bundle. I would like to draw it eventually though, in case we want to tweak it some.

peteschaffner:
A couple things I would like to work on before this gets merged:
- Flatten most icons out a bit more by toning down shadows and gradients
- Improve selection state on "Arrangements" icon (pixels with alpha channel don't get included in the highlight)
- Draw "Advanced" icon at a slightly smaller size to better fit with the scale of our set (or vice versa)

gnachman:
Much better! I agree with your TODO list as well.

kaishin:
Nice work @peteschaffner. I'd probably stick the Advanced icon used by other apps. Too bad the one that can be set in Xcode is not retina-optimized, so you have to include the asset in the app.

peteschaffner:
Thanks @kaishin. Yeah, I noticed that too about the Advanced Xcode default icon. I agree, it is probably better to stick with the system standard as people are accustomed to it, but just for funs I drew a new one. I'm fine with reverting it though :)

Now with moar flat:
image

gnachman:
Looking good! The old icons make me cringe when I look at them now. A
slightly scaled-down replica of the standard Advanced icon would be ideal,
but I like the one you drew too.

peteschaffner:
Ok, I drew the system ToolbarAdvanced icon at a slightly scaled down size. I think I prefer it as well. I also played with the scale of the two circular icons a lot, but finally settled on this. Let me know your thoughts.
image

kaishin:
This looks good to merge as far as I am concerned.

> The old icons make me cringe when I look at them now.

Mavericks as a whole does that to me now, including a lot of icons I designed.

gnachman:
I loved your icons for a long time, @kaishin , didn't mean to slag on them :). But it's time to get flat up in here.

PR #204, opened 2015-03-29
Don't forget about processes that were alive when we tried to reap them

ridiculousfish:
Pids are added to a "deadpool" in TaskNotifier when we expect them to
die soon. When reaping processes, it's possible that waitpid() will
return 0, meaning that the process is not yet dead. In that case, put
its pid back into the deadpool so that we attempt to reap it again.
Otherwise the process will become a zombie.

Fixes https://code.google.com/p/iterm2/issues/detail?id=3471


turing-complete:
Brilliant! Thanks a lot for this!

gnachman:
Thanks, that was an ugly little bug.

PR #205, opened 2015-03-30
Strip more bracketed delimiters

grddev:
The main motivation for this change is that git when presenting merge conflicts sometimes shows the file paths in single quotes. Instead of just adding single quotes, I thought it was just as well to add some other delimiters that are unlikely to appear in file names

### DIFF
--- a/sources/iTermSemanticHistoryController.m
+++ b/sources/iTermSemanticHistoryController.m
@@ -40,19 +40,6 @@ NSString *const kSemanticHistoryWorkingDirectorySubstitutionKey = @"semanticHist
 @synthesize prefs = prefs_;
 @synthesize delegate = delegate_;
 
-static unsigned int oppositeDelimiter(unichar start) {
-    switch(start) {
-        case '(': return ')';
-        case '<': return '>';
-        case '[': return ']';
-        case '{': return '}';
-        case '\'':
-        case '"':
-            return start;
-        default:
-            return (unsigned int)-1; // unsigned int to prevent it from matching any unsigned short
-    }
-}
 
 - (NSString *)getFullPath:(NSString *)path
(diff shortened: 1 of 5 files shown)


sources/iTermSemanticHistoryController.m
```diff
@@ -73,9 +87,13 @@ - (NSString *)getFullPath:(NSString *)path
         return nil;
     }
 
-    // If it's in parens, strip them.
-    if (path.length > 2 && [path characterAtIndex:0] == '(' && [path hasSuffix:@")"]) {
-        path = [path substringWithRange:NSMakeRange(1, path.length - 2)];
+    // If it's in any form of bracketed delimiters, strip them
```
gnachman: I think this implementation chooses performance over clarity, which isn't the right tradeoff here (in the midst of filesystem access). I'd suggest doing something like this:

    while ([path isBracketed]) {
        path = [path stringByRemovingBrackets];
    }

and add -isBracketed and -stringByRemovingBrackets methods to NSStringIterm, like:

    - (BOOL)isBracketed {
        if (self.length < 2) return NO;
        if ([self hasPrefix:@"("] && [self hasSuffix:@")"]) return YES;
        if ([self hasPrefix:@"<"] && [self hasSuffix:@">"]) return YES;
        etc.
        return NO;
    }

grddev:
@gnachman: I moved the implementation into NSStringITerm, but only exposed a method for removing all delimiters, instead of peeling one delimiter at a time. I guess one could split into further methods if need arises, but the method just below it (stringByRemovingTerminatingPunctuation) seemed to similarly remove all punctuation and not just one symbol.

gnachman:
Thanks!

PR #206, opened 2015-04-10
Fix for setting BBEdit as default for semantic history

fingolfin:
Without this, if a user has both BBEdit and TextMate installed and
chooses BBEdit as default for semantic history, iTerm2 would still
launch TextMate. This is because both were configured to use the
same URL scheme.
using that has the problem that when both BBEdit

(Also fix a typo in a debug message :)

Note: Unfortunately, I am unable to build iTerm2, so this has *not* been verified to work -- things, is, on my OS X 10.8.5 machine, ibtoold segfaults (on NumberOfSpacesAccessoryView.xib, I think). Still, this fix should be correct, I hope. I'll be happy to test a nightly build with this, or a custom test binary if somebody is willing to provide one.

Thanks for making iTerm2!

### DIFF
--- a/sources/iTermSemanticHistoryController.m
+++ b/sources/iTermSemanticHistoryController.m
@@ -295,7 +295,7 @@ NSString *const kSemanticHistoryWorkingDirectorySubstitutionKey = @"semanticHist
     }
 
     if ([prefs_[kSemanticHistoryActionKey] isEqualToString:kSemanticHistoryCoprocessAction]) {
-        DLog(@"Launch coproress with script %@", script);
+        DLog(@"Launch coprocess with script %@", script);
         assert(delegate_);
         [delegate_ semanticHistoryLaunchCoprocessWithCommand:script];
         return YES;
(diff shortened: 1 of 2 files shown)


gnachman:
How long has x-bbedit been around? Will this break a lot of users?

fingolfin:
The oldest BBEdit release I can easily get my hands on is 8.7.2, released on 12/19/2007. It already supports x-bbedit.

Anyway, if you are concerned about that, I wonder why you use the approach via URL schemes in the first place -- naively spoken, wouldn't using the bundle ID plus `[NSWorkspace - openURLs:withAppBundleIdentifier:options:additionalEventParamDescriptor:launchIdentifiers:]` be the better choice then? That said, I didn't test it, so perhaps there is some "obvious" flaw with that idea :).

gnachman:
Hm, I didn't know about that NSWorkspace method. I'm sure I searched for an alternate URL scheme but somehow missed x-bbedit. Anyway, this looks fine. Thanks for doing the research.

gnachman:
I had to revert this--I couldn't get it to work. Let me take a stab at using the NSWorkspace method you found.

gnachman:
Opening by bundle ID done in commit f2dc34f

fingolfin:
Sorry, didn't see your updates. I am glad to see `openURLs:withAppBundleIdentifier` worked out, that seems like the better (and easier to extend) method anyway!

fingolfin:
(Oh, and I meant to say: I tested a nightly build, and it works like a charm, great!)

PR #208, opened 2015-04-16
Fix search direction for Find Next and Find Previous actions

allewun:
⌘G and ⌘⇧G were going the wrong directions when searching within iTerm2.

The comment in the code mentioned something about the search UI being reversed, but I couldn't reproduce this. I tested the left / right buttons in the Find View and they work as expected.

### DIFF
--- a/sources/PseudoTerminal.m
+++ b/sources/PseudoTerminal.m
@@ -7209,15 +7209,12 @@ static const CGFloat kHorizontalTabBarHeight = 22;
     [[self currentSession] showFindPanel];
 }
 
-// findNext and findPrevious are reversed here because in the search UI next
-// goes backwards and previous goes forwards.
-// Internally, next=forward and prev=backwards.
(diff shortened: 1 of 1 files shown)


gnachman:
Cmd-G goes backwards by default because search goes backwards. It's more useful to find results from the bottom up, in general, and especially if you have a very long scrollback history. So this is working as intended. Users can always change Cmd-G and Cmd-Shift-G in Prefs>Keys.

allewun:
I agree with the idea, but these defaults are inconsistent with pretty much all apps that use the shortcuts, Terminal.app included.

I think users should be able to expect standard behaviors of these shortcuts across their apps. Given that you can remap the keys as you pointed out, my vote stands for keeping the native direction and allowing power users to remap.

adriaant:
I'm with @gnachman The terminal isn't a text-editor or webpage. Search has to go backwards, since older content is on top.

PR #209, opened 2015-04-16
Fix a bookmarks bug caused by a nil value

ConnorCimowsky:
In the case where no migration was needed from old-style bookmarks to new-style bookmarks, no value would be set for `KEY_NEW_BOOKMARKS`. This resulted in the bookmarks for `KEY_DEFAULT_GUID` never getting loaded.

This fixes [Issue 3491](https://code.google.com/p/iterm2/issues/detail?id=3491).


gnachman:
See commit acf0329, I think it's a simpler solution. Let me know if it gives you any trouble.

PR #211, opened 2015-04-23
Move badge to bottom right and half the size as default.

gwydirsam:
Here is a screenshot:

iterm2shot


gnachman:
I put the badge on the top right because there's rarely anything interesting up there. On the bottom, you'll tend to run into status bars. Based on the screenshot, I'm guessing vi's status bar would touch the bottom of that badge. Maybe placement should be a configuration option, though.

gwydirsam:
You are right. I have mine set to disable the badge for vim, mutt, irssi things like that because of that exact reason. Here is a shot of that disabled:

itermsmall

gnachman:
Quite reasonable; please file a feature request for configurable position and font.

PR #212, opened 2015-04-24
Make the find bar first responder immediately upon opening

ridiculousfish:
More aggressive fix for issue 3496. The original fix allowed
a possibility for enqueued events to be processed before
the text field gained focus, so that for example typing
cmd-F followed by 'a', the 'a' would go to the iTerm window.
This fix ensures that enqueued events go to the search field.

### DIFF
--- a/sources/FindViewController.m
+++ b/sources/FindViewController.m
@@ -476,12 +476,8 @@ const CGFloat kEdgeWidth = 3;
 
     [NSAnimationContext endGrouping];
 
-    dispatch_async(dispatch_get_main_queue(), ^{
-        DLog(@"Grab focus for find view %@", self.view);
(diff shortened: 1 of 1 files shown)


ridiculousfish:
I found that after pressing cmd-F and beginning to type, sometimes characters would go to the main document instead of the search field. I propose to fix this by making it the first responder immediately.

I found that the call to `-findViewControllerMakeDocumentFirstResponder` boils down to a call to `makeFirstResponder`, so I believe it can be omitted. Feedback very welcome.

gnachman:
I was so pissed that I had to do a delayed perform here, didn't even notice the delegate call!

ridiculousfish:
Coolio :)

PR #219, opened 2015-07-07
take midpoint of min/max when estimating points for badge text font

rolandwalker:
This makes no substantive difference to the calculation, but is likely what was intended.


gnachman:
Good catch, thanks.

PR #220, opened 2015-07-08
configurable tab label truncation style

rolandwalker:
Choice of Right, Middle, or Left elision to make label text fit on tab GUI element.

Setting only visible in Preferences / Advanced.

Exposing a bare `int` for the setting is crude.  I'm hazily assuming you may add an `NSDictionary` type to the advanced settings in the future (mapping to a dropdown in the GUI), which could be made backward-compatible to the `int` value.

### DIFF
--- a/ThirdParty/PSMTabBarControl/source/PSMYosemiteTabStyle.m
+++ b/ThirdParty/PSMTabBarControl/source/PSMYosemiteTabStyle.m
@@ -9,6 +9,7 @@
 #import "PSMYosemiteTabStyle.h"
 #import "PSMTabBarCell.h"
 #import "PSMTabBarControl.h"
+#import "iTermAdvancedSettingsModel.h"
 
 #define kPSMMetalObjectCounterRadius 7.0
 #define kPSMMetalCounterMinWidth 20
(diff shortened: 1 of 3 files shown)


ThirdParty/PSMTabBarControl/source/PSMYosemiteTabStyle.m
```diff
@@ -9,6 +9,7 @@
 #import "PSMYosemiteTabStyle.h"
 #import "PSMTabBarCell.h"
 #import "PSMTabBarControl.h"
+#import "iTermAdvancedSettingsModel.h"
```
gnachman: I don't want iTerm code leaking into PSM code. It really should be in a separate target. This should go up the delegate chain, instead.

smikheiev:
hi! what's the status of this PR? it would be nice to get this feature in

gnachman:
@smikheiev This has gotten more complex since 2015. Can you elaborate on your use case? It's generally better to do things automatically if possible, although this might be hard depending on what's going on.

rolandwalker:
I'd forgotten about this, and would be fine with closing, since I didn't know how to follow through.  But it is a desirable feature.

PR #221, opened 2015-07-09
V2 applescript newline

wfaulk:
Implementation for [gitlab issue 3662](https://gitlab.com/gnachman/iterm2/issues/3662)

### DIFF
--- a/English.lproj/iTerm.scriptTerminology
+++ b/English.lproj/iTerm.scriptTerminology
@@ -258,6 +258,13 @@
 					<key>Name</key>
 					<string>text</string>
 				</dict>
+				<key>newline</key>
+				<dict>
+					<key>Description</key>
+					<string>If newline should be appended to text</string>
+					<key>Name</key>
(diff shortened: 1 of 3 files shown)


gnachman:
Thanks for the PR! You don't want to mess with the v2 branch, though: it's a dead end. Work on master, where applescript is generally much cleaner. Also, let's make the argument "without newline". The common case is writing out simple shell commands, so the default is appropriate.

wfaulk:
Fair enough.  I'll close this and make the same update in master.

That said, I'm not sure what you mean by "make the argument 'without newline'".  The way I implemented it, the argument could have been either "with newline" or "without newline", both doing the obvious thing.  Leaving the argument out altogether reverted to the old "with newline unless ending in space" method, so as not to break existing scripts.

I'm going to assume we don't need the backward compatibility in a new major version, so I'll drop that last part and just make the default be straightforward "with newline" method.  That is, drop the ending-in-space special case.

PR #222, opened 2015-07-11
Add "with/without newline" argument to Applescript write verb

wfaulk:
I've completely replaced the old notion of not sending a newline if the text ends in a space.

I think this works, but I'm currently unable to see how to use Applescript with iTerm2 v3.

With iTerm2 v2, I could do something like:

	set term to (make new terminal)
	tell term
		set sess to (launch session "default")
	end tell
	tell sess to write text "echo"

With iTerm2 v3, I simply don't see how to get a handle on a session, new or old, to use `write` with.

As such, I cannot test my changes. If you let me know how to get a session handle, I'll be happy to test and fix, if needed.

### DIFF
--- a/iTerm2.sdef
+++ b/iTerm2.sdef
@@ -288,6 +288,10 @@
       <parameter name="text" code="Text" type="text"
           description="Text to send" optional="yes">
         <cocoa key="text"/>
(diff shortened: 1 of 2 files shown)


gnachman:
I modified all the create window, create tab, and split session commands to return the new window/tab/session so you can work the way you used to.

PR #223, opened 2015-07-15
Allow substring hostname matching.

dadowns:
This adds in the ability to only specify part of the full hostname for APS. For example instead of having to add:
someservice01.stage.company.com
someservice02.stage.company.com
someservice03.stage.company.com
otherservice01.stage.company.com
otherservice02.stage.company.com

you can just add:
.stage.company.com

and for prod:
.prod.company.com

In my use case I wanted two ssh profiles, one for stage servers and one for production and didn't want to have to type in over 50 server names into each.

I left exact match in place to take precedence over a partial match.

### DIFF
--- a/sources/iTermRule.m
+++ b/sources/iTermRule.m
@@ -93,16 +93,43 @@
   int score = 0;
 
   if (self.hostname != nil) {
-    NSRange containsHost = [hostname rangeOfString:self.hostname options:NSCaseInsensitiveSearch];
-    if (containsHost.location != NSNotFound) {
-      if (containsHost.location == 0 && containsHost.length == hostname.length) {
+    // Fixup hostname for regex match
+    NSString *fixedHostname = [self.hostname stringByReplacingOccurrencesOfString:@"." withString:@"\\."];
+    int prevLength = [fixedHostname length];
+    fixedHostname = [fixedHostname stringByReplacingOccurrencesOfString:@"*" withString:@".*"];
+    int postLength = [fixedHostname length];
+
+    // If no wildcards just do exact match
+    if (prevLength == postLength) {
+      if ([hostname isEqualToString:self.hostname]) {
         score |= kHostExactMatchScore;
+      }
+    } else {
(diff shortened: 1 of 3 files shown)


dadowns:
Cleaned this up to only do a single string compare and guarded against a nil hostname.

gnachman:
Maybe it would make more sense to use wildcards to match partial host names? There's already a precedent here for `username@*:/path`. Also, I added a unit test--please sync and update the test appropriately.

dadowns:
Ok, updated to use wildcards instead. This supports multi wildcard matching, example:
*.stage.company.com
service*.*.company.com
*.stage.*

Exact matches still have a higher score. Also tweaked the best score logic for cases like this:
Actual host: 
service01.stage.company.com

Rules:
*.stage.company.com // matches
service*.stage.company.com // matches better because the hostname is longer

I figured the more explicitly provided hostname should have higher precedence.

sources/iTermRule.m
```diff
@@ -85,17 +85,51 @@ - (NSString *)description {
 - (int)scoreForHostname:(NSString *)hostname
                username:(NSString *)username
                    path:(NSString *)path {
-  const int kHostMatchScore = 4;
-  const int kUserMatchScore = 2;
-  const int kPathMatchScore = 1;
+  const int kHostExactMatchScore = 8;
```
gnachman: Minor style issue: don't try to keep the = signs aligned.

dadowns: fixed

gnachman:
I've added a new method in commit c268e37, -[NSString stringMatchesCaseInsensitiveGlobPattern:], which you should use instead of regular expressions. Regexes are trouble because there are so many special characters that you can't escape them all.

dadowns:
Fixed. When I originally started I contemplated writing a state machine to match ? and *, but that seemed a bit much and I like your solution better. Thanks for your feedback and consideration on this, it'll be a big help for a bunch of our engineers here.

PR #224, opened 2015-07-17
Fixed titlebar separator

fjolnir:
Titlebar separator is now slightly lighter in color, and adjusts itself when the window is in the background.

It's a small change, but it's a lot nicer this way.

### Before (View at full size)
before

### After (View at full size)
after


gnachman:
This is a good idea but the color of the division view should be changed in response to -[PseudoTerminal windowDidBecomeKey:] and  -windowDidResignKey: rather than modifying SolidColorView.

fjolnir:
How's this?

gnachman:
Thanks!

PR #225, opened 2015-07-17
Make small scroll events move the terminal by one line.

ojotoxy:
This makes the scrolling behavior a bit nicer so it doesn't effectively ignore small scroll events. It matches the behavior of Terminal.app a bit better in this regard.

### DIFF
--- a/sources/PTYScrollView.m
+++ b/sources/PTYScrollView.m
@@ -116,7 +116,21 @@
     NSRect scrollRect;
 
     scrollRect = [self documentVisibleRect];
-    scrollRect.origin.y -= [theEvent deltaY] * [self verticalLineScroll];
+
+    CGFloat delta = [theEvent deltaY];
+    // Make sure that a very small scroll event moves by at least one line.
+    if (fabs(delta) < 1) {
+        if (delta > 0) {
+            delta = 1;
+        } else if (delta < 0) {
+            delta = -1;
+        } else {
+            // The delta could be 0 in case of touchpad scrolling.
+            delta = 0;
+        }
+    }
+
+    scrollRect.origin.y -= delta * [self verticalLineScroll];
     [[self documentView] scrollRectToVisible:scrollRect];
 
     [self detectUserScroll];


apetresc:
I haven't tested out the patch, but I'm strongly in favor of the concept!

wladston:
Also didn't test, also +1 for the concept :)

gnachman:
I haven't been able to generate a scroll event with a small nonzero value. But maybe you can't do it with a trackpad.

Also, this change limits deltaY to 1 or -1. Was that intentional?

ojotoxy:
It shouldn't limit the size of large scroll events due to this line:

if (fabs(delta) < 1) {

I just tested with a trackpad and some events about .5 in size were generated. But they weren't very common. They are very common with a mouse scroll wheel.

gnachman:
Of course, you're right. Seems OK, let's give it a try and see if anyone complains.

KevinBaragonaIJKCO:
Just curious what became of this. Scrolling is slower again.

ojotoxy:
Just curious what became of this. It's back to how it was originally :)

gnachman:
I haven't changed anything. Can you tell in which version it came back?

ojotoxy:
Seems like this commit: https://github.com/gnachman/iTerm2/commit/0809386339b8f61c68a141fc753cbd93d1dc7478#diff-ea9bd756a249eb6f04d1cf7c3408f902 <https://github.com/gnachman/iTerm2/commit/0809386339b8f61c68a141fc753cbd93d1dc7478#diff-ea9bd756a249eb6f04d1cf7c3408f902>

Discussion... https://gitlab.com/gnachman/iterm2/issues/3209 <https://gitlab.com/gnachman/iterm2/issues/3209>

gnachman:
Thanks for the clarifying links, @ojotoxy. The current behavior appears to match Terminal.app pretty well. Commit 588845f3 adds an advanced preference to use the older logic.

PR #226, opened 2015-07-20
Fix pattern matching for user@host and user@host:path

beatrohrer:
- the colon from user@host:path has been included in the hostname
- user@host has not been recognized


gnachman:
Thanks. I added unit tests and fixed a few more bugs.

PR #228, opened 2015-07-30
added check for hotkeywindow before setting alphavalues

reeselevine:
See https://gitlab.com/gnachman/iterm2/issues/3724


gnachman:
Thanks!

PR #231, opened 2015-08-11
Fixed wording in settings for triple-click setting.

Nirma:
While using iTerm I noticed that the wording was a little odd for the triple click setting.

It originally said "Triple-click selects full wrapped lines" but perhaps something like "Triple-click selects entire wrapped lines" would be clearer. 

Hope this isn't too much of a nitpick!


gnachman:
Agreed, I never liked the wording of this one.

PR #232, opened 2015-08-16
Resolved a few open issues w/imgls.pl

mikecappella:
1. Fixed the overlapping image issues when height=1.  This was resolved
by adding a null byte after the closing ^G sequence (this seems
required to allow the ^G to be accepted).
2. Removed the Cursor Up after the image was drawn.  This doesn’t work
reliably enough with various height/width/aspect ratio settings.
3. Fixed in issue when a file was not readable.


gnachman:
I was able to reproduce the bug where the token after the image was dropped, and it'll be fixed in tomorrow's nightly build.

PR #233, opened 2015-09-04
Added option to make windows borderless and rounded

DrabWeb:
After submitting a screenshot of my custom iTerm to the r/UnixPorn subreddit, people liked it and asked me to commit it to the master branch. This adds the option in the profile menu for a user to select the borderless rounded window style, which looks like the picture above.


gnachman:
This looks interesting--would you mind making a few changes so I can pull it?

1. Clean up the unrelated changes to xib files
2. We don't need a separate rounded vs non-rounded setting. Just replace the existing title-less mode with the rounded title-less mode.
3. (the hard part) Preserve the old logic for folks on OSes older than 10.10.

I'm happy to help answer any questions to help w/ this. Thanks!

perj:
I just want to point out that there is one more difference than the roundness of the window. The current title-less window doesn't have any shadow, and this one does. Personally I prefer shadows, so it isn't really an issue for me, just something to note.

DrabWeb:
@gnachman I don't know how I made those changes to the XIB, but I will revert them. This method also does not work on < 10.10. If it were to be kept in older version of OSX, I would need to do a borderless window, then apply a rounded window mask, which would be a custom bitmap. I do not know how to do this, and i've tried looking before but people haven't even made examples. Some people also like having the window with square corners, so I don't know if replacing it would be a good idea, but if you want me to, I will.

@perj I will look into this for a different pull request, but for now i'm focusing on rounded

perj:
I didn't mean for you to have to change anything. Sorry if that was confusing.

gnachman:
I think I like that it has a shadow but I can see how people would tile these and prefer that it not have one. Good news is it's easy to remove a window's shadow, the only hard part is figuring out the right default.

DrabWeb:
@gnachman I have cleaned up the XIBs, after some fiddling around (I never do pull requests). Just to clarify, do you still want me to replaced the squared borderless with this?

gnachman:
Maybe you need to rebase? I see all kinds of bogus diffs still.

I do think this should replace the square borderless window style.

DrabWeb:
@gnachman Yeah, I guess I will just re-branch the repository and make the changes again, then do a new pull request. You can just close this one.

DrabWeb:
Here is a link to the new pull request

https://github.com/gnachman/iTerm2/pull/237

PR #234, opened 2015-09-06
force encoding value to 32bit

khwon:
NSStringEncoding is defined as NSUInteger, which is defined as unsigned int in 32bit-env & unsigned long in 64bit-env.
However iTerm saves value in internal dictionary as 32bit signed int, so casting 32bit signed int to 64bit unsigned int occurs, making unintended sign expansion (0xffffffff80000422 instead of 0x0000000080000422). So we need to clear high 32bit of encoding value to correctly match encoding values with predefined constants.

Also we can handle encoding option as 64bit value in whole application but that might break 32-bit build environment.


gnachman:
Thanks for letting me know. A more comprehensive fix is in commit 580ce68.

PR #239, opened 2015-10-04
Setting to allow hot key window to float above windows even in background

aleffert:
This is useful for watching the status of a terminal even when you're
doing other things.


gnachman:
Merged with fixes to popup windows.

aleffert:
Thanks!

PR #241, opened 2015-10-10
Modernization

MaddTheSane:
These changes are to help make the code more future-proof, and integrate with Swift better.

### DIFF
--- a/sources/AdvancedWorkingDirectoryWindowController.h
+++ b/sources/AdvancedWorkingDirectoryWindowController.h
@@ -14,6 +14,6 @@
 @property(nonatomic, copy) Profile *profile;
 
 // An array of keys that may be changed in self.profile. All take string values.
-@property (readonly) NSArray<NSString*> *allKeys;
+@property(nonatomic, readonly) NSArray<NSString *> *allKeys;
 
 @end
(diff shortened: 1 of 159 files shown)


gnachman:
Wow.

Thank you for taking this on. It will probably take me some time to get through this all.

sources/AdvancedWorkingDirectoryWindowController.h
```diff
@@ -14,6 +14,6 @@
 @property(nonatomic, copy) Profile *profile;
 
 // An array of keys that may be changed in self.profile. All take string values.
-- (NSArray *)allKeys;
+@property (readonly, strong) NSArray<NSString*> *allKeys;
```
gnachman: Why strong? IIRC that's ARC only and not needed for a readonly property.

MaddTheSane: Ah, you are right.
I was trying to make it more ARC and Swift friendly.

Should I revert it?

gnachman:
I've started a branch **MaddTheSane-modernization** where I'm making some changes for style. There's a lot of testing to do, but it looks good so far.

gnachman:
Merged. Thanks!

PR #247, opened 2015-12-11
Fill in x/ypixel in struct winsize for the TIOCGWINSZ ioctl.

0x09:
https://developer.apple.com/library/mac/documentation/Darwin/Reference/ManPages/man4/tty.4.html

iTerm currently fills in the row/column fields, but not the pixel density. This commit brings it in line with Terminal's behavior.

I was actually using this in combination with terminfo commands to play gifs before gif support was added.

### DIFF
--- a/sources/PTYSession.m
+++ b/sources/PTYSession.m
@@ -1169,7 +1169,7 @@ static NSTimeInterval kMinimumPartialLineTriggerCheckInterval = 0.5;
 {
     DLog(@"Set session %@ to %dx%d", self, width, height);
     [_screen resizeWidth:width height:height];
-    [_shell setWidth:width height:height pixelWidth:width*_textview.charWidth pixelHeight:height*_textview.lineHeight];
+    [_shell setWidth:width height:height pixelWidth:width * _textview.charWidth pixelHeight:height * _textview.lineHeight];
     [_textview clearHighlights];
     [[_tab realParentWindow] invalidateRestorableState];
 }
@@ -3586,6 +3586,7 @@ static NSTimeInterval kMinimumPartialLineTriggerCheckInterval = 0.5;
             [[[self tab] parentWindow] fitWindowToTab:[self tab]];
         }
     }
+
     // If the window isn't able to adjust, or adjust enough, make the session
     // work with whatever size we ended up having.
     if ([self isTmuxClient]) {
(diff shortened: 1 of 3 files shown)


sources/PTYTask.m
```diff
@@ -43,6 +43,8 @@
                 struct winsize* win,
                 int width,
                 int height,
+                int pixelWidth,
```
gnachman: Should be an `unsigned short` to match the fields in the winsize struct.

sources/PTYSession.m
```diff
@@ -1169,7 +1169,7 @@ - (void)setWidth:(int)width height:(int)height
 {
     DLog(@"Set session %@ to %dx%d", self, width, height);
     [_screen resizeWidth:width height:height];
-    [_shell setWidth:width height:height];
+    [_shell setWidth:width height:height pixelWidth:width*_textview.charWidth pixelHeight:height*_textview.lineHeight];
```
gnachman: Please put spaces around the `*` operators.

gnachman:
The tty should be notified when the session size changes because of a font change that does not adjust the session's row/col size. I'd do it in this method:
````
- (void)setFont:(NSFont*)font
     nonAsciiFont:(NSFont*)nonAsciiFont
    horizontalSpacing:(float)horizontalSpacing
    verticalSpacing:(float)verticalSpacing
````

at the end of this if statement's block:
````
        if ([iTermPreferences boolForKey:kPreferenceKeyAdjustWindowForFontSizeChange]) {
            [[[self tab] parentWindow] fitWindowToTab:[self tab]];
````

sources/PTYTask.m
0x09: Shall I apply this to the existing width/height params? They aren't used inside these functions besides to assign ws_row/col.
Same regarding `launchWithPath` and `setWidth:...`

gnachman: If you're feeling ambitious, yes please! This code is old and crufty and I like to clean it up as I go, but I won't hold up this PR for that.

shabbyrobe:
This would help me a lot; is there anything specific (beyond a bit of a merge/rebase) that needs to happen to resurrect this PR?

gnachman:
Implemented in commit ccf7ed6cd

shabbyrobe:
Amazing! Thank you so much!

PR #248, opened 2015-12-16
Fix find view invocation/dismissal animation

roandelyf:
We were over-shooting the initial find view bezier curve drawing, making
the curve->line->curve paths discontinuous on both sides.

In the animation, this would appear as 2 jagged triangles on both side of the find view,
whenever the view's height was smaller than 2*radius of the bezier curve.


roandelyf:
Hi gnachman, I'd like to start contributing some patches to iTerm2.
Are pull requests the preferred way for doing so?

gnachman:
Yes, pull requests are great! Sorry for being slow, I have family visiting this week and haven't had time to do much work.

PR #249, opened 2015-12-22
Preserve multiline prompt on 'clear buffer' gnachman/iterm2#1330

dreadatour:
"Clear Buffer" clears all output except last line, which supposed to be a command prompt. But prompt can consist of more then one line. It would be great to set number of lines that isn't clear by "Clear Buffer".

https://gitlab.com/gnachman/iterm2/issues/1330


sources/VT100Grid.m
```diff
@@ -388,12 +388,12 @@ - (int)scrollUpIntoLineBuffer:(LineBuffer *)lineBuffer
...
         unlimitedScrollback:(BOOL)unlimitedScrollback
-        preserveCursorLine:(BOOL)preserveCursorLine {
+        preservePromptLines:(int)preservePromptLines {
     self.scrollRegionRows = VT100GridRangeMake(0, size_.height);
     self.scrollRegionCols = VT100GridRangeMake(0, size_.width);
     int numLinesToScroll;
-    if (preserveCursorLine) {
-        numLinesToScroll = cursor_.y;
+    if (preservePromptLines > 0) {
+        numLinesToScroll = cursor_.y - preservePromptLines + 1;
```
gnachman: Must ensure this does not go negative. `MAX(0, cursor_.y - preservePromptLines + 1)`.

sources/VT100Grid.m
```diff
@@ -405,7 +405,7 @@ - (int)resetWithLineBuffer:(LineBuffer *)lineBuffer
     }
     self.cursor = VT100GridCoordMake(0, 0);
 
-    [self setCharsFrom:VT100GridCoordMake(0, preserveCursorLine ? 1 : 0)
+    [self setCharsFrom:VT100GridCoordMake(0, preservePromptLines)
```
gnachman: This will probably explode if preservePromptLines >= size_.height, so constrain it to not exceed `size_.height - 1`.

sources/VT100Screen.m
```diff
@@ -848,13 +848,22 @@ - (void)clearBuffer {
 // This clears the screen, leaving the cursor's line at the top and preserves the cursor's x
 // coordinate. Scroll regions and the saved cursor position are reset.
 - (void)clearAndResetScreenPreservingCursorLine {
+    int numberOfLinesInPrompt = [iTermAdvancedSettingsModel numberOfLinesInPrompt];
```
gnachman: Rather than using an advanced setting, it would be better to figure out how tall the prompt is from shell integration. The prompt starts at the line `currentGrid_.cursor.y` when `-terminalPromptDidStart` gets called and ends at the line `currentGrid_.cursor.y` when `-terminalCommandDidStart` gets called. I would keep track of the last two values. If shell integration is not installed, then you could fall back to the advanced setting.

sources/VT100Grid.m
dreadatour: Agreed, fix this in line 395 above - add check for preservePromptLines < size_.height.

sources/VT100Screen.m
dreadatour: This is very good remark. I didn't use shell integration, and I think a lot about how to automate this (to do not use `numberOfLinesInPrompt` advanced setting).

I will add shell integration support, thank you!

PR #250, opened 2015-12-29
add shrink and expand function to menubar and keyboard shortcut

salexkidd:
Dear iTerm2 developers.

I added shrink and expand window function to Menu bar(and keyboard short cut preference).
Please merge If you are satisfied.


salexkidd:
Please off the "Disable session-initiated window resizing" (Preferences->Profile->Terminal Tab)checkbox when test.

sources/PseudoTerminal.m
```diff
@@ -5637,6 +5637,16 @@ - (IBAction)moveTabRight:(id)sender
     [self tabsDidReorder];
 }
 
+- (IBAction)expandRow:(id)sender {
+    [self.currentSession.screen terminalSetRows:self.currentSession.rows + 1
```
gnachman: If you call `[self sessionInitiatedResize:self.currentSession width:self.currentSession.columns height:self.currentSession.rows+1]` then it'll work even if terminal-initiated resizing is disabled. Same in `shrinkRow`.

Interfaces/MainMenu.xib
```diff
@@ -957,6 +958,19 @@ DQ
                             <menuItem isSeparatorItem="YES" id="1405">
                                 <modifierMask key="keyEquivalentModifierMask" command="YES"/>
                             </menuItem>
+                            <menuItem title="Shrink Row" keyEquivalent="" id="E89-Mw-yOH">
```
gnachman: How about `Increase Height` and `Decrease Height` instead of `Shrink/Expand Row`?

If we're going to add commands for changing height then we need to add commands for changing the width as well.

sources/PseudoTerminal.m
```diff
@@ -6222,7 +6232,7 @@ - (void)safelySetSessionSize:(PTYSession*)aSession rows:(int)rows columns:(int)c
     }
     PtyLog(@"safelySetSessionSize");
     BOOL hasScrollbar = [self scrollbarShouldBeVisible];
-    if (windowType_ == WINDOW_TYPE_NORMAL || windowType_ == WINDOW_TYPE_NO_TITLE_BAR) {
+    if (windowType_ == WINDOW_TYPE_NORMAL || windowType_ == WINDOW_TYPE_NO_TITLE_BAR || windowType_ == WINDOW_TYPE_TOP) {
```
gnachman: I think you found a bug. This line should actually say:
````
if (![self anyFullScreen]) {
````
Since all window types besides full screen are now resizable.

Comment (salexkidd) [inline sources/PseudoTerminal.m]: Thx for Fix it! :+1: 
I pulled a master(75da04e5c55e68dfa128bf5fab48af0e9374ddc0)

File: Interfaces/MainMenu.xib (hunk 2 above)
Comment (salexkidd) [inline Interfaces/MainMenu.xib]: I think that it is a good idea! and I'll add Add changing width function.

**Comment by salexkidd** [comment]
@gnachman I pushed. Please check my code. thanks. (Sorry, My english is so poor...!)

File: sources/PseudoTerminal.m [hunk 4]
```diff
@@ -5637,6 +5637,34 @@ - (IBAction)moveTabRight:(id)sender
     [self tabsDidReorder];
 }
 
+- (IBAction)increaseRow:(id)sender {
+    if (![iTermProfilePreferences boolForKey:KEY_DISABLE_WINDOW_RESIZING inProfile:self.currentSession.profile]) {
```
gnachman: You don't want this if statement--the feature should work even if terminal-initiated resizing is off

sources/PseudoTerminal.m
```diff
@@ -5637,6 +5637,34 @@ - (IBAction)moveTabRight:(id)sender
...
+        [self.currentSession.screen terminalSetRows:self.currentSession.rows + 1
```
gnachman: Why not call `[self sessionInitiatedResize:self.currentSession width:self.currentSession.columns height:self.currentSession.rows+1]`? It saves several levels of indirection.

sources/iTermKeyBindingMgr.m
```diff
@@ -471,6 +471,22 @@ + (NSString *)formatAction:(NSDictionary *)keyInfo
                             [self stringForSelectionMovementUnit:auxText.integerValue]];
             break;
 
+        case KEY_ACTION_DECREASE_ROW:
+            actionString = @"Decrease Row";
```
gnachman: Please rename the actions, constants, and methods to use this wording:

 * Decrease Height
 * Increase Height
 * Decrease Width
 * Increase Width

salexkidd:
Change constant and method name, and call sessionInitiatedResize.

gnachman:
Looks good. Thanks!

PR #251, opened 2015-12-29
Enq

aljex:
[edit] This is a "real" pull request now. [/edit]

Hi, this is not a "real" pull request, yet. I would like anyone who can, to tell me what am I missing?

I have added a hard-coded ENQ Answerback to my own fork of iTerm2. It seems to be working fine so far, although I haven't tested extensively, such as testing if it corrupts print jobs or downloads etc.

Now I am trying to make it into a proper config option in Profiles -> Terminal instead of hard-coded.

I've managed to get as far as getting the Profiles -> Terminal screen to accept, store, and display a stored value as expetcted, and the code builds and runs without error, but the value is not actually output in response to ENQ.

In VT100Terminal.m, under "case VT100CC_ENQ:" the commented-out line with a hard-coded value DOES work. The next line which is currently uncommented and tries to use the saved config value instead of a hard-coded value, outputs nothing.

So I think I'm pretty close, but I guess I've reached the limit of my ability to clomp around without actually knowing what I'm doing. WHat am I missing?

Sorry about the big diffs in the .xib file. xcode actually made over 500!!! edits all by itself which I painfully manually reverted one at a time to get the diff down to what it is now. The real changes in that file are basically adding the new Answerback String field to the Profiles Terminal view, and shifting the bottom half of that view down a little to make room for it. But xcode seems to have also relocated chunks of xml around which made large diffs.

Thanks I hope!

iterm2_enq_cfg

iterm2_enq_run


gnachman:
Look in `PTYSession.m` where `setTermVariable:` is called. You need to do the same thing to set the answerback string in the `VT100Terminal` instance. In other words, something like this:

````
[self setTermVariable:[iTermProfilePreferences stringForKey:KEY_TERMINAL_TYPE inProfile:aDict]];  // existing line
_terminal.answerBackString = [iTermProfilePreferences stringForKey:KEY_ANSWERBACK_STRING inProfile:aDict]
````

**Comment by aljex** [comment]
Thanks to your hint I got 99% of the way. It is responding to ENQ now, and it is sending the value stored in prefs.

The only thing left is, the hard-coded version interprets that &quot;&#92;n&quot; I have in the string, and sends a newline, while the code in this branch sends backslash n.

It means I can't use the shell's built-in read command to read the string.
I do have a standalone answerback binary which I can use to read the answerback without a trailing newline, but really the client should be configurable to send it.

What I added (beyond this pull request) to get to this point was:

in PTYSession.h:

```obj-c
[self setTermVariable:[iTermProfilePreferences stringForKey:KEY_TERMINAL_TYPE inProfile:aDict]];  // existing line for context
[self setAnswerBackString:[iTermProfilePreferences stringForKey:KEY_ANSWERBACK_STRING inProfile:aDict]];  // added this line
...
- (void)setAnswerBackString:(NSString *)s
{
    [_answerBackString autorelease];
    _answerBackString = [s copy];
    [_terminal setAnswerBackString:_answerBackString];
}
```

in VT100Terminal.m:
```obj-c
- (void)setAnswerBackString:(NSString *)s
{
    _answerBackString = s;
}
```

I guess I just need to insert something like a snprintf() or similar somewhere?

Maybe something like this?<br>
http://stackoverflow.com/questions/7860867/converting-escaped-utf8-characters-back-to-their-original-form/7861345#7861345

Or this?<br>
http://stackoverflow.com/a/11615076

Or maybe I need to change something about the way the value is stored into prefs in the first place?<br> &quot;defaults read ...&quot; outputs &quot;iTerm2&#92;&#92;&#92;&#92;n&quot; , four backslashes for the single one entered into the text field. The defaults utility might be changing it for display too. Only a single backslash comes through from the client to the host though. Maybe I'm mis-using NSString somewhere and should be using NSData? Take a c string from the user via the NSTextField, but immediately treat it as NSData the rest of the time?

I don't care what exact format has to be written into the config field in order to get the desired output, backslash codes, carat codes, even urlencoding, just that there is some way to do it.

I can't quite follow any of the examples I've googled up so far like above. They seem like they might almost do what I want, but I don't know how to take the examples and do it for real. I'm not actually a c coder or an obj-c coder at all. I am amazed I got this far. I have been doing 100% pure monkey see monkey do here by copying other parts of the existing code. Which means also, if there is anything about this code that doesn't seem right or optimal or conforming to best practices, do not assume I had some reason for it. Assume I am utterly ignorant.

aljex:
OK I have it interpreting a few common shell escapes, and escaped hex values, using the handy dandy stringByReplacingEscapedChar and stringByReplacingEscapedHexValuesWithChars, which already existed in the code.

I guess this is a "real" pull request now!

Ah, well maybe I should put something about this in the help and whatever other docs are part of the package, and *then* it's a "real" pull request.

aljex:
I'd also like to know how to make changes to PreferencePanel.xib without xcode making eleventy-seven berjillion changes and rearranging big chunks of the file and basically making a huge unreadable diff that completely hides the changes that I actually made. How the heck can you or anyone else audit a diff like that? The diff for that file in this pull request is insane.

xcode does the same thing to the plist files sometimes. If I edit the file at all, then all the elements get some change like 0's become 0.0's etc. I have had to manually discard all changes a few times to keep the commits clean. Luckily for those files there are no changes I actually want to make (outside of my own other fork), but what if I had?

This seems like a real poor interaction between xcode and any RCS.

aljex:
Sample crazy answerback string:
iterm2_aljex_prefs_enq_crazy_in

Sample normal answerback string:
iterm2_aljex_prefs_enq_normal_in

Output/results of both:
iterm2_aljex_prefs_enq_out

Note: Default value is intentionally still empty, not "iTerm2\n".

gnachman:
> I'd also like to know how to make changes to PreferencePanel.xib without xcode making eleventy-seven berjillion changes and rearranging big chunks of the file and basically making a huge unreadable diff that completely hides the changes that I actually made. How the heck can you or anyone else audit a diff like that? The diff for that file in this pull request is insane.
> xcode does the same thing to the plist files sometimes. If I edit the file at all, then all the elements get some change like 0's become 0.0's etc. I have had to manually discard all changes a few times to keep the commits clean. Luckily for those files there are no changes I actually want to make (outside of my own other fork), but what if I had?
> This seems like a real poor interaction between xcode and any RCS.

Everyone who uses .xib files and .plists feels the same way. Apple doesn't seem interested in solving this problem, unfortunately.

Interfaces/iTerm2 Help/index.html
```diff
@@ -1243,6 +1243,12 @@
 
 <p><br/><br/></p>
 
+<h4>Answerback String</h4>
```
gnachman: This file is auto-generated from `source/_includes/documentation-preferences.md` in the repo https://github.com/gnachman/iterm2-website. Please send a separate PR to update that file.

plists/DefaultBookmark.plist
```diff
@@ -560,6 +560,8 @@
 	<false/>
 	<key>Terminal Type</key>
 	<string>xterm</string>
+	<key>Answerback String</key>
```
gnachman: No need to specify this; the default will be taken from code. This file could probably go away with a little work, but it definitely doesn't need to grow.

plists/MigrationMap.plist
```diff
@@ -718,6 +718,18 @@
       <string>xterm</string>
     </dict>
     <dict>
+        <key>Prefix</key>
```
gnachman: Revert changes to this file. It's only needed for prefs that existed before iTerm2 existed.

sources/PTYSession.m
```diff
@@ -205,6 +205,8 @@ @implementation PTYSession {
     VT100Terminal *_terminal;
 
     NSString *_termVariable;
+    
+    NSString *_answerBackString;
```
gnachman: PTYSession doesn't need to keep a copy of `answerBackString`. It's ok for VT100Terminal to keep the only copy of it. So you can get rid of this ivar and get rid of `setAnswerbackString:` and just set it directly in the VT100Terminal.

sources/VT100Terminal.m
```diff
@@ -269,6 +270,18 @@ - (void)setTermType:(NSString *)termtype
     [delegate_ terminalTypeDidChange];
 }
 
+- (void)setAnswerBackString:(NSString *)s
+{
```
gnachman: Move the open brace to the end of the preceding line

sources/VT100Terminal.m
```diff
@@ -269,6 +270,18 @@ - (void)setTermType:(NSString *)termtype
...
+    s = [s stringByReplacingEscapedChar:'a' withString:@"\x07"];
```
gnachman: I suggest using `[s stringByExpandingVimSpecialCharacters]` instead. It's a bit more capable and already in use elsewhere.

sources/VT100Terminal.m
```diff
@@ -1234,7 +1247,10 @@ - (void)executeToken:(VT100Token *)token {
 
         //  VT100 CC
         case VT100CC_ENQ:
-            // TODO: Add support for an answerback string here.
+            // This works, and sends a newline.
```
gnachman: Delete the commented-out code.

sources/VT100Terminal.m
```diff
@@ -1234,7 +1247,10 @@ - (void)executeToken:(VT100Token *)token {
...
+            //[delegate_ terminalSendReport:[@"iTerm2\n" dataUsingEncoding:NSUTF8StringEncoding]];
+            // This works, and sends a literal backslash n when the string has \n in it.
+            [delegate_ terminalSendReport:[_answerBackString dataUsingEncoding:NSUTF8StringEncoding]];
```
gnachman: Use `self.encoding` instead of `NSUTF8StringEncoding`.

sources/VT100Terminal.m
```diff
@@ -2408,6 +2425,7 @@ - (void)setStateFromDictionary:(NSDictionary *)dict {
         return;
     }
     self.termType = dict[kTerminalStateTermTypeKey];
+    self.answerBackString = dict[kTerminalStateAnswerBackStringKey];
```
gnachman: You don't want to pull an NSNull out here. I really should have done this for termType too.
````
if ([self.answerBackString isKindOfClass:[NSNull class]]) {
  self.answerBackString = nil;
}
````

File: sources/PTYSession.m (hunk 4 above)
Comment (aljex) [inline sources/PTYSession.m]: I failed to figure out how to do this.

**Comment by aljex** [comment]
I got only a few of those things.

File: sources/PTYSession.m (hunk 4 above)
Comment (gnachman) [inline sources/PTYSession.m]: Instead of doing
````
[self setAnswerBackString:[iTermProfilePreferences stringForKey:KEY_ANSWERBACK_STRING inProfile:aDict]];
````
do:
````
[_terminal setAnswerBackString:[iTermProfilePreferences stringForKey:KEY_ANSWERBACK_STRING inProfile:aDict]];
````

Then you can remove _answerBackString from this file.

File: sources/VT100Terminal.m [hunk 10]
```diff
@@ -269,6 +270,12 @@ - (void)setTermType:(NSString *)termtype
     [delegate_ terminalTypeDidChange];
 }
 
+- (void)setAnswerBackString:(NSString *)s
+{
```
gnachman: Please move the brace up to the end of the preceding line.

sources/VT100Terminal.m
```diff
@@ -2408,6 +2416,10 @@ - (void)setStateFromDictionary:(NSDictionary *)dict {
         return;
     }
     self.termType = dict[kTerminalStateTermTypeKey];
+
+    self.answerBackString = dict[kTerminalStateAnswerBackStringKey];
+    if ([self.answerBackString isKindOfClass:[NSNull class]]) self.answerBackString = nil;
```
gnachman: Please use braces for the `if` statement's body.

gnachman:
Let me know if you need more clarification on anything. This is very close to being ready to pull.

aljex:
I think I'm just about there except for the help index.html

gnachman:
Merged, thanks! Please update the docs as we discussed. Also, please add an example of setting the answerback to the applescript docs (https://github.com/gnachman/iterm2-website/blob/master/source/applescript.md)

PR #252, opened 2015-12-30
Focus ring drawing fix, and some code refactoring

roandelyf:
#### 4 commits

**Use dispatch_once for one-shot actions**
Unless there's a specific reason not to use this (for perf, if all static variable accesses are guaranteed to
occur on the main thread), dispatch_once is usually a cleaner pattern.

**focus ring drawing**
this is something that has always irked me in AppKit. The focus ring is appearing above the title bar when sliding down, even though the search field is behind that bar.

**Align static path strings**
some variable definition alignment

**Refactor system version check code**
refactored duplicate code

### DIFF
--- a/sources/HotkeyWindowController.m
+++ b/sources/HotkeyWindowController.m
@@ -712,18 +712,16 @@ static CGEventRef OnTappedEvent(CGEventTapProxy proxy, CGEventType type, CGEvent
 }
 
 - (void)requestAccessibilityPermissionMavericks {
-    static BOOL alreadyAsked;
-    if (alreadyAsked) {
-        return;
-    }
-    alreadyAsked = YES;
 #if MAC_OS_X_VERSION_MAX_ALLOWED >= 1090
-    NSDictionary *options = [NSDictionary dictionaryWithObject:[NSNumber numberWithBool:YES]
-                                                        forKey:(NSString *)kAXTrustedCheckOptionPrompt];
-    // Show a dialog prompting the user to open system prefs.
-    if (!AXIsProcessTrustedWithOptions((CFDictionaryRef)options)) {
-        return;
-    }
+    static dispatch_once_t onceToken;
+    dispatch_once(&onceToken, ^{
+        NSDictionary *options = [NSDictionary dictionaryWithObject:[NSNumber numberWithBool:YES]
+                                                            forKey:(NSString *)kAXTrustedCheckOptionPrompt];
+        // Show a dialog prompting the user to open system prefs.
+        if (!AXIsProcessTrustedWithOptions((CFDictionaryRef)options)) {
+            return;
+        }
+    });
 #endif
 }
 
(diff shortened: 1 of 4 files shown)


sources/HotkeyWindowController.m
```diff
@@ -712,18 +712,16 @@ - (void)carbonHotkeyPressed:(id)handler {
 }
 
 - (void)requestAccessibilityPermissionMavericks {
-    static BOOL alreadyAsked;
-    if (alreadyAsked) {
-        return;
-    }
-    alreadyAsked = YES;
 #if MAC_OS_X_VERSION_MAX_ALLOWED >= 1090
```
gnachman: Would you mind getting rid of this #if and #endif while you're in here?

sources/iTermController.m
```diff
@@ -60,18 +60,18 @@ - (void)_cycleWindowsReversed:(BOOL)back;
 @end
 
 // Constants for saved window arrangement key names.
-static NSString* APPLICATION_SUPPORT_DIRECTORY = @"~/Library/Application Support";
-static NSString *SUPPORT_DIRECTORY = @"~/Library/Application Support/iTerm";
-static NSString *SCRIPT_DIRECTORY = @"~/Library/Application Support/iTerm/Scripts";
+static NSString *APPLICATION_SUPPORT_DIRECTORY = @"~/Library/Application Support";
+static NSString *SUPPORT_DIRECTORY             = @"~/Library/Application Support/iTerm";
```
gnachman: The coding style used in iTerm2 is to not align equal signs and such.

sources/iTermController.m
```diff
@@ -124,16 +111,15 @@ @implementation iTermController {
     id runningApplicationClass_;
 }
 
-static iTermController* shared = nil;
-static BOOL initDone = NO;
+static iTermController* shared;
```
gnachman: This code is so old! `shared` ought to go inside `sharedInstance`.

sources/HotkeyWindowController.m
```diff
@@ -712,18 +712,16 @@ - (void)carbonHotkeyPressed:(id)handler {
...
-    NSDictionary *options = [NSDictionary dictionaryWithObject:[NSNumber numberWithBool:YES]
-                                                        forKey:(NSString *)kAXTrustedCheckOptionPrompt];
-    // Show a dialog prompting the user to open system prefs.
-    if (!AXIsProcessTrustedWithOptions((CFDictionaryRef)options)) {
-        return;
-    }
+    static dispatch_once_t onceToken;
+    dispatch_once(&onceToken, ^{
+        NSDictionary *options = [NSDictionary dictionaryWithObject:[NSNumber numberWithBool:YES]
```
gnachman: Please update this to `@{ (NSString *)kAXTrustedCheckOptionPrompt: @YES }`

gnachman:
Thanks for this; just a few minor changes and I'll be glad to pull it.

sources/iTermController.m
ronyfadel: It's used by +sharedInstanceRelease, that's why it's outside of sharedInstance

gnachman:
Looks like there are merge conflicts. Please rebase and I can pull.

roandelyf:
should be mergeable now

PR #253, opened 2015-12-30
Store Character Encoding pref as NSUInteger instead of int.

aljex:
Fixes: https://gitlab.com/gnachman/iterm2/issues/4105


sources/NSTextField+iTerm.m
```diff
@@ -38,4 +38,10 @@ - (int)separatorTolerantIntValue {
     }
 }
 
+- (NSUInteger)separatorTolerantUIntValue {
```
gnachman: Call it `UnsignedInteger` instead of `UInt` to hint that it's 64-bit (to match the `NSNumber` convention).

sources/iTermPreferences.h
```diff
@@ -153,6 +153,9 @@ extern NSString *const kPreferenceKeyShowFullscreenTabBar;
 + (int)intForKey:(NSString *)key;
 + (void)setInt:(int)value forKey:(NSString *)key;
 
++ (NSUInteger)uintForKey:(NSString *)key;
```
gnachman: unsignedIntegerForKey

sources/PreferenceInfo.h
```diff
@@ -12,8 +12,10 @@ typedef NS_ENUM(NSInteger, PreferenceInfoType) {
     kPreferenceInfoTypeCheckbox,
     kPreferenceInfoTypeInvertedCheckbox,  // true=checked, false=unchecked. Handy when inverting a checkbox's text, but the user defaults key can't be changed.
     kPreferenceInfoTypeIntegerTextField,
+    kPreferenceInfoTypeUnsignedIntegerTextField,
     kPreferenceInfoTypeStringTextField,
     kPreferenceInfoTypePopup,
+    kPreferenceInfoTypeUPopup,
```
gnachman: Let's avoid abbreviations and call it UnsignedIntegerPopup. Also, add a comment to Popup saying it is deprecated in favor of UnsignedIntegerPopup.

sources/ToolJobs.m
```diff
@@ -93,6 +93,20 @@ - (void)setIntValue:(int)i {
     [self setToolTip:[NSString stringWithFormat:@"SIG%@ (%d)", signalNames[i], i]];
 }
 
+- (NSUInteger)uintValue {
```
gnachman: No need to change `ToolJobs.m`. The only caller of `setIntValue:` in this class is `textDidEndEditing:` just below.

sources/iTermPreferences.m
```diff
@@ -279,9 +279,13 @@ + (BOOL)defaultValueForKey:(NSString *)key isCompatibleWithType:(PreferenceInfoT
     id defaultValue = [self defaultValueMap][key];
     switch (type) {
         case kPreferenceInfoTypeIntegerTextField:
+        case kPreferenceInfoTypeUnsignedIntegerTextField:
         case kPreferenceInfoTypePopup:
             return ([defaultValue isKindOfClass:[NSNumber class]] &&
                     [defaultValue doubleValue] == ceil([defaultValue doubleValue]));
+        case kPreferenceInfoTypeUPopup:
+            return ([defaultValue isKindOfClass:[NSNumber class]] &&
+                    [defaultValue unsignedLongValue] == ceil([defaultValue unsignedLongValue]));
```
gnachman: Use `unsignedIntegerValue`, not `unsignedLongValue` for consistency. Doesn't matter in practice as both are 64 bit.

sources/iTermPreferences.m
```diff
@@ -391,6 +395,14 @@ + (void)setInt:(int)value forKey:(NSString *)key {
     [self setObject:@(value) forKey:key];
 }
 
++ (NSUInteger)uintForKey:(NSString *)key {
+    return [(NSNumber *)[self objectForKey:key] unsignedIntegerValue];
```
gnachman: Does this do the right thing if a negative value was saved in an older version of iTerm2?

sources/iTermPreferences.m
```diff
@@ -391,6 +395,14 @@ + (void)setInt:(int)value forKey:(NSString *)key {
...
+}
+
++ (void)setUInt:(NSUInteger)value forKey:(NSString *)key {
```
gnachman: Call it `setUnsignedInteger:` instead.

sources/iTermPreferencesBaseViewController.m
```diff
@@ -157,6 +170,10 @@ - (IBAction)settingChanged:(id)sender {
                 [self setInt:[sender selectedTag] forKey:info.key];
                 break;
 
+            case kPreferenceInfoTypeUPopup:
+                [self setUInt:[sender selectedTag] forKey:info.key];
```
gnachman: `-[NSPopupButton selectedTag]` is an `NSInteger`. I think it would be a bug for it to have a negative value, so we should assert that it is positive here.

sources/iTermPreferencesBaseViewController.m
```diff
@@ -247,6 +264,14 @@ - (void)updateValueForInfo:(PreferenceInfo *)info {
             break;
         }
 
+        case kPreferenceInfoTypeUnsignedIntegerTextField: {
+            assert([info.control isKindOfClass:[NSTextField class]]);
+            NSTextField *field = (NSTextField *)info.control;
+            // FIXME: What is correct here? There is no unsignedIntegerValue in NSTextField.
+            field.doubleValue = [self uintForKey:info.key];
```
gnachman: You could do `field.stringValue = [NSString stringWithFormat:@"%lu", [self unsignedIntegerForKey:info.key]];`

sources/iTermPreferencesBaseViewController.m
```diff
@@ -270,6 +295,13 @@ - (void)updateValueForInfo:(PreferenceInfo *)info {
             break;
         }
 
+        case kPreferenceInfoTypeUPopup: {
+            assert([info.control isKindOfClass:[NSPopUpButton class]]);
+            NSPopUpButton *popup = (NSPopUpButton *)info.control;
+            [popup selectItemWithTag:[self uintForKey:info.key]];
```
gnachman: Assert that `[self unsignedIntegerForKey:info.key]` will fit in a signed NSInteger (is less than or equal to `NSIntegerMax`), since tags are signed.

sources/iTermPreferencesBaseViewController.m
```diff
@@ -344,6 +376,19 @@ - (int)intForString:(NSString *)s inRange:(NSRange)range
     return val;
 }
 
+- (NSUInteger)uintForString:(NSString *)s inRange:(NSRange)range
+{
```
gnachman: Move the open brace to the end of the preceding line (this is the newer style, and I'm updating code as I go). Also, this method should be called `unsignedIntegerForString:inRange:`.

sources/iTermPreferencesBaseViewController.m
```diff
@@ -362,6 +407,21 @@ - (void)applyIntegerConstraints:(PreferenceInfo *)info {
...
+    NSTextField *textField = (NSTextField *)info.control;
+    NSUInteger iv = [self uintForString:[textField stringValue] inRange:info.range];
+    unichar lastChar = '0';
+    int numChars = [[textField stringValue] length];
+    if (numChars) {
+        lastChar = [[textField stringValue] characterAtIndex:numChars - 1];
+    }
+    if (iv != [textField separatorTolerantUIntValue] || (lastChar < '0' || lastChar > '9')) {
+        // FIXME: What is correct here? There is no "setUnsignedIntegerValue" in NSTextField
+        [textField setDoubleValue:iv];
```
gnachman: See previous comment about setting the textField's `stringValue`.

aljex:
Thanks! Ok I got most of that stuff.  I think it's only missing:

* I don't quite know how to do the two assert()'s you want.
* And I'm testing now to see what happens when a new binary reads an old plist with a negative value in it.

gnachman:
Thanks for the changes so far! Will review them presently.

> I don't quite know how to do the two assert()'s you want.

To assert that a tag is positive:
````
const NSInteger tag = [NSPopupButton selectedTag];
assert(tag >= 0);
````

To assert that an unsigned integer will fit in an NSInteger:
````
const NSUInteger value = [self unsignedIntegerForKey:info.key];
assert(value <= NSIntegerMax);
````

File: sources/PreferenceInfo.h [hunk 13]
```diff
@@ -12,8 +12,10 @@ typedef NS_ENUM(NSInteger, PreferenceInfoType) {
     kPreferenceInfoTypeCheckbox,
     kPreferenceInfoTypeInvertedCheckbox,  // true=checked, false=unchecked. Handy when inverting a checkbox's text, but the user defaults key can't be changed.
     kPreferenceInfoTypeIntegerTextField,
+    kPreferenceInfoTypeUnsignedIntegerTextField,
     kPreferenceInfoTypeStringTextField,
     kPreferenceInfoTypePopup,
```
gnachman: Add a comment here warning that it only support 32-bit tag values.

sources/iTermProfilePreferences.m
```diff
@@ -77,9 +88,13 @@ + (BOOL)defaultValueForKey:(NSString *)key isCompatibleWithType:(PreferenceInfoT
     id defaultValue = [self defaultValueMap][key];
     switch (type) {
         case kPreferenceInfoTypeIntegerTextField:
+        case kPreferenceInfoTypeUnsignedIntegerTextField:
```
gnachman: I think this should go with UnsignedIntegerPopup since a double cannot precisely represent every possible NSUInteger.

sources/iTermProfilePreferences.m
```diff
@@ -77,9 +88,13 @@ + (BOOL)defaultValueForKey:(NSString *)key isCompatibleWithType:(PreferenceInfoT
...
         case kPreferenceInfoTypePopup:
             return ([defaultValue isKindOfClass:[NSNumber class]] &&
                     [defaultValue doubleValue] == ceil([defaultValue doubleValue]));
+        case kPreferenceInfoTypeUnsignedIntegerPopup:
+            return ([defaultValue isKindOfClass:[NSNumber class]] &&
+                    [defaultValue unsignedIntegerValue] == ceil([defaultValue unsignedIntegerValue]));
```
gnachman: Remove the call to `ceil` here, that was only needed for `doubleValue` above.

sources/PreferenceInfo.h
aljex: I don't know what you mean exactly here. What about the new value type requires a special warning exactly? Or do you mean that the old one needs the warning? Sorry.

aljex:
So I just tried 2.1.4 and it does NOT store those DOS encodings as negative values in the users local plist. It stores the same numerical values that my -enq branch does.

2.9.20160102 stores the same negative numbers I saw in 2.9.2015mmdd 

For each of these, the app is being launched from my Applications folder.
I ensure all processes are closed by doing menu>quit<br>
I wipe locale settings with defaults delete com.googlecode.iterm2 while there are no running iterms of any version anywhere.<br>
launch the app by clicking it from the applications folder.<br>
Go in to prefs and see what the encoding is out of the box first.<br>
Pick a different encoding,<br>
Use a regular osx terminal to read the same plist I just deleted.<br>
Close and re-open the same app to see if it recognizes and properly handles the data it wrote itself.<br>

2.1.4-latest 
"Unicode (UTF-8)" = "Character Encoding" = 4;  (initial)
"Western (DOS Latin 1)" = "Character Encoding" = 2147484688;
"Latin-US (DOS)" = "Character Encoding" = 2147484672;
Reads the stored values correctly on restart.

2.9.20160102
Crash log on first run with no local settings present.
Icon left in dock after crashlog. Clicking icon in doc brings up a normal working terminal window.
Default "Unicode (UTF-8)" = ""
becomes "4" only after I click on the encoding preference and put it back to utf-8.
"Western (DOS Latin 1)" = "Character Encoding" = "-2147482608"
"Latin-US (DOS)" = "Character Encoding" = "-2147482624";
Reads and those values and somehow arrives at the correct encoding that created them, even  after an app restart.

-enc-char-pref branch latest:
same utf8 default at first after wipe, stored as "4"<br>
"Western DOS Latin 1" = ""Character Encoding" = 2147484688;"
"Latin-US (DOS)" = "Character Encoding" = 2147484672;

Upgrades & downgrades:
Saving a negative value with a 2.9 binary
Reading with 2.1.4
Whoa!
2.1.4 *correctly* interprets the negative number on startup, AND changes it to the positive number for the same encoding if you go in to prefs and click on the encodings at all.
And still shows the same correct/expected encoding after the change. 2.1.4 is doing magic on the fly already?
 Brians-Air:~ bkw$ defaults read com.googlecode.iterm2 |grep Encoding
            "Character Encoding" = "-2147482608";
Brians-Air:~ bkw$ defaults read com.googlecode.iterm2 |grep Encoding
            "Character Encoding" = 2147484688;

If I leave the positive value in place,.. quit all iterms entirely, start 2.9, it reads the positive value correctly. The right encoding (western dos latin 1) is set. If I click on it the encoding popup, but abort the popup by clicking elsewhere, the defaults command still shows the negative value. If I click on the same "Western (DOS Latin 1)" that was already in the field, no changing anything, but this time closing the popup list of encodings by choosing the same one that was already there... the defaults command now shows the positive value.
Brians-Air:~ bkw$ defaults read com.googlecode.iterm2 |grep Encoding
            "Character Encoding" = "-2147482608";
Brians-Air:~ bkw$ defaults read com.googlecode.iterm2 |grep Encoding
            "Character Encoding" = 2147484688;

I have to pick this up tomorrow. I'm too tired to keep track of exactly what actions produce what results, and what pattern they all make taken together.

Save new positive value for Western DOS Latin 1 from new -enc-char-pref branch
Load with 2.9
Wrong encoding, shows Arabic
Select Wetsern DOS Latin 1 and save it. Saves as negative number.
Close and re-open 2.9. Same 2.9 that just wrote this pref.
value is still same negative
Says Arabic DOS again,
Click on Arabic DOS to choose it
... an new negative number: "Character Encoding" = "-2147482599";
(I expected 0 or 1 actually...)
That was all in 2.9

ok I'm fried for now...

sources/PreferenceInfo.h
gnachman: Yeah, the old one needs a warning.

gnachman:
Good! It works both 2147484688 and -2147482608 are encoded as is 0x80000410. I don't expect we'll have a problem then.

aljex:
It compiles and runs and still behaves correctly as far as I can see, but I actually have no idea if the assert()'s are actually right, if they're checking the things that actually need to be checked.

gnachman:
Asserts look correct to me. Thanks for doing a great job on this PR!

aljex:
Thank you!

PR #254, opened 2015-12-31
Access app delegate via -[iTermApplication delegate] instead of casti…

roandelyf:
…ng down from NSApplication APIs

### DIFF
--- a/sources/iTermApplication.m
+++ b/sources/iTermApplication.m
@@ -47,6 +47,12 @@
     [super dealloc];
 }
 
++ (iTermApplication *)sharedApplication {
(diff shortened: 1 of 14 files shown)


sources/iTermApplication.m
```diff
@@ -47,6 +47,12 @@ - (void)dealloc {
     [super dealloc];
 }
 
++ (iTermApplication *)sharedApplication {
+    __kindof NSApplication *sharedApplication = [super sharedApplication];
```
gnachman: Love this use of `__kindof`, never seen it before!

PR #255, opened 2016-01-02
Idle keepalive character period, behavior and setting.

aljex:
* Fix anti-idle period ignoring user setting.<br>
  (a couple of different things combined to make neither less than nor greater than 60 possible)
* Allow anti-idle more frequent than 60 seconds.
* Anti-idle starts counting immediately instead of waiting 60 sec before the first time.
* Anti-idle settings change now take effect immediately instead of requiring a new session.
* Remove anti-idle on/off setting. Period 0 = off , >0 = on.
* Move anti-idle period setting from app-wide advanced pref to per-profile pref.<br>
  - Most sessions do not need or want nulls.
  - Some sessions need very frequent nulls.
  - Others need only infrequent nulls.
  - User can still set their own desired app-wide default behavior for new sessions by just setting the value in Default profile.
* Put both anti-idle ui parts together in same panel.<br>
  Preferences>Profiles>Session

Default is 0 seconds, aka feature is disabled by default.

Since the diff for PreferencePanel.xib is insane, attached is a pic showing the actual change, which can just be done manually.
 * The left-hand check box was cut and replaced with a plain text label.
 * The new NSTextField on the right, "Idle Period" (seconds), is links to<br>
  Referencing Outlets: _idlePeriod - Session View Controller
iterm2_aljex_prefs_idle


sources/PTYSession.m
```diff
@@ -3070,30 +3073,15 @@ - (void)setBlend:(float)blendVal {
...
-    if (set) {
-        NSTimeInterval period = MIN(60, [iTermAdvancedSettingsModel antiIdleTimerPeriod]);
-
-        _antiIdleTimer = [[NSTimer scheduledTimerWithTimeInterval:period
-                                                           target:self
+- (void)setAntiIdle:(NSTimeInterval)period {
+    [_antiIdleTimer invalidate];
+    [_antiIdleTimer release];
+    _antiIdleTimer = nil;
+    if (period > 0) _antiIdleTimer = [[NSTimer scheduledTimerWithTimeInterval:period
```
gnachman: Use braces for if statement body.

sources/PTYSession.m
```diff
@@ -3070,30 +3073,15 @@ - (void)setBlend:(float)blendVal {
...
 }
 
-- (BOOL)antiIdle
-{
-    return _antiIdleTimer ? YES : NO;
-}
-
-- (void)setAntiIdle:(BOOL)set
-{
-    if (set == [self antiIdle]) {
```
gnachman: I want to keep the checkbox that enables/disables the feature.
 * Using 0 to disable it is more unintuitive than having a checkbox, and requires extra wording on the preference panel.
 * It really is a dangerous feature and has been the cause of many bug reports from confused users, so the warning message is important to keep.

sources/PTYSession.m
```diff
@@ -3503,10 +3491,8 @@ - (void)scheduleUpdateIn:(NSTimeInterval)timeout {
 
 - (void)doAntiIdle {
     NSTimeInterval now = [NSDate timeIntervalSinceReferenceDate];
-    if (now >= _lastInput + 60) {
-        [_shell writeTask:[NSData dataWithBytes:&_antiIdleCode length:1]];
-        _lastInput = now;
-    }
+    [_shell writeTask:[NSData dataWithBytes:&_antiIdleCode length:1]];
```
gnachman: The if statement ensured that it actually only activated during idle times. As this is written, it will write when not idle. The original behavior seems better, since  writing the anti-idle code while you're busy typing a command would (perhaps invisibly) corrupt your work.

sources/PreferenceInfo.h
```diff
@@ -12,6 +12,7 @@ typedef NS_ENUM(NSInteger, PreferenceInfoType) {
     kPreferenceInfoTypeCheckbox,
     kPreferenceInfoTypeInvertedCheckbox,  // true=checked, false=unchecked. Handy when inverting a checkbox's text, but the user defaults key can't be changed.
     kPreferenceInfoTypeIntegerTextField,
+    kPreferenceInfoTypeFloatTextField,
```
gnachman: Let's use Double, not Float. The extra precision doesn't come with a significant price.

sources/ProfilesSessionPreferencesViewController.m
```diff
@@ -72,45 +72,15 @@ - (void)awakeFromNib {
...
-    info.observer = ^() {
-        _idleCode.enabled = [self boolForKey:KEY_SEND_CODE_WHEN_IDLE];
-    };
-
     info = [self defineControl:_idleCode
                            key:KEY_IDLE_CODE
                           type:kPreferenceInfoTypeIntegerTextField];
-    info.range = NSMakeRange(0, 256);
-
+    info.range = NSMakeRange(0, 255);
```
gnachman: 256 was correct because it's the range's size, not its maximum value.

sources/iTermAdvancedSettingsModel.m
```diff
@@ -65,7 +65,6 @@ + (NSString *)name { \
 
 #pragma mark Terminal
 DEFINE_BOOL(traditionalVisualBell, NO, @"Terminal: Visual bell flashes the whole screen, not just a bell icon.");
-DEFINE_FLOAT(antiIdleTimerPeriod, 60, @"Terminal: Anti-idle interval in seconds.\nWill not go faster than 60 seconds.");
```
gnachman: A migration path should be provided for users who had this set by using a computed default value.

sources/iTermPreferencesBaseViewController.m
```diff
@@ -243,6 +247,13 @@ - (void)updateValueForInfo:(PreferenceInfo *)info {
         case kPreferenceInfoTypeIntegerTextField: {
             assert([info.control isKindOfClass:[NSTextField class]]);
             NSTextField *field = (NSTextField *)info.control;
+            field.doubleValue = [self floatForKey:info.key];
```
gnachman: I think this change was made in the wrong place?

sources/iTermPreferencesBaseViewController.m
```diff
@@ -243,6 +247,13 @@ - (void)updateValueForInfo:(PreferenceInfo *)info {
...
+            break;
+        }
+
+        case kPreferenceInfoTypeFloatTextField: {
+            assert([info.control isKindOfClass:[NSTextField class]]);
+            NSTextField *field = (NSTextField *)info.control;
             field.intValue = [self intForKey:info.key];
```
gnachman: doubleForKey

sources/iTermPreferencesBaseViewController.m
```diff
@@ -344,6 +355,12 @@ - (int)intForString:(NSString *)s inRange:(NSRange)range
     return val;
 }
 
+- (double)FloatForString:(NSString *)s {
```
gnachman: Did you mean to leave this in?

sources/PTYSession.m
aljex: Ah yes. I think I missed that in the manual copy from my own hacked-up fork over to a clean reference branch of your master.

sources/PreferenceInfo.h
aljex: NSTimeInterval is double, so I started to use double everywhere, but there happened to already be a floatForKey or something else float, that happened to be actually cast as double, but just named float___, so I decided to use them rather than add essentially duplicates. But I'm all for improving consistency as long as no one looks at me funny for making a new (double)doublefoo right next to a preexisting (double)floatfoo. ;)

sources/ProfilesSessionPreferencesViewController.m
aljex: Got it. Sorry.

sources/iTermPreferencesBaseViewController.m
aljex: I'm 4 years old and you are more patient than I probably would be.

sources/PreferenceInfo.h
gnachman: Gah, what a mess. Please call the new one double, and some day I'll need to find all the instances of Float and rename them to Double.

sources/iTermPreferencesBaseViewController.m
gnachman: Not bad for a 4 year old! :P

aljex:
Cleaned up the float vs double stuff.

xib file conflicts like crazy now of course from the enq/answerback addition, I'll resolve that later by updating my branch from master.

My main problem right now is I don't know why "if(now >= _lastInput + _antiIdlePeriod)" isn't working in doAntiIdle().

...ah so it's not just me,
Even if I hard code a value like 10 in place of _antiIdlePeriod, (like it was originally except it was 60),
It will not output any more frequent than the hard coded value. So if I have `if(now >= _lastInput + 10)`
 and set the pref to 1, it only outputs the code every 10 seconds, with or without keyboard activity.

Set the pref to 20 and it will increase to 20 seconds between codes, But still not respecting keyboard activity.

I may still be handling _antiIdlePeriod wrong, but even with a hard coded value in it's place, it's like _lastInput is not being updated by keyboard activity.

I'm testing by running hexdump in the terminal btw since it shows the ^@'s , in case that matters somehow when it comes to detecting keyboard activity.

I do agree that it's probably not acceptable to just send the nulls regardless of activity, since that must surely corrupt something somewhere sometime. Though I haven't checked yet if other terminal emulators bother with that. I do know though that at least one commercial terminal emulator has a "tty_activity" keepalive script in their docs which you run on the host side from your .profile, which just goes into the background and sends nulls to your tty for the entire duration of your login session. And that script is not checking for activity, it's just sending a null every few seconds blindly forever.

gnachman:
The problem is _lastInput gets updated, so only every other call writes. Plus there's a race condition because the timer runs exactly as often as the antiIdlePeriod. You should write if it's close to time, or else add 100ms to the timer period. I tried logging the values of now and _lastInput, and it produced this:

````
2016-01-06 11:32:16.467 iTerm2[60203:20457741] now=473801536.467583 lastInput=473801531.467735 antiIdlePeriod=5
2016-01-06 11:32:21.467 iTerm2[60203:20457741] now=473801541.467829 lastInput=473801531.467735 antiIdlePeriod=5
2016-01-06 11:32:26.467 iTerm2[60203:20457741] now=473801546.467131 lastInput=473801541.467829 antiIdlePeriod=5
2016-01-06 11:32:31.467 iTerm2[60203:20457741] now=473801551.467387 lastInput=473801541.467829 antiIdlePeriod=5
2016-01-06 11:32:36.467 iTerm2[60203:20457741] now=473801556.467139 lastInput=473801551.467387 antiIdlePeriod=5
2016-01-06 11:32:41.466 iTerm2[60203:20457741] now=473801561.466509 lastInput=473801551.467387 antiIdlePeriod=5
2016-01-06 11:32:46.466 iTerm2[60203:20457741] now=473801566.466777 lastInput=473801561.466509 antiIdlePeriod=5
2016-01-06 11:32:51.466 iTerm2[60203:20457741] now=473801571.466456 lastInput=473801566.466777 antiIdlePeriod=5
2016-01-06 11:32:56.465 iTerm2[60203:20457741] now=473801576.46592 lastInput=473801566.466777 antiIdlePeriod=5
2016-01-06 11:33:01.466 iTerm2[60203:20457741] now=473801581.465966 lastInput=473801576.46592 antiIdlePeriod=5
````

aljex:
This pr has been obsoleted by a later one.

PR #257, opened 2016-01-05
Improve Anti-Idle take:2

aljex:
This doesn't work yet, but I'm trying to redo this more the way gnachman wants, without taking out so much of the existing code and behavior.

Basically now the ui is good and the prefs are being updated and saved to disk immediately on edit, and are being read back after restart all as expected.

But the anti-idle code is never being sent no matter what settings get changed.

In my other branch, aljex/iTerm2 idle-code, everything DOES work, with the exception that it does not detect idleness vs activity. But in that branch the anti-idle code IS sent, and the timing behavior of the code DOES reflect the user setting, and that behavior even changes immediately upon editing the pref (no new window needed).

But that branch works differently from this branch. That branch gets rid of the antiidle on/off bool and pretty much everything related to it, and it passes the KEY_IDLE_PEIOD value directly to the setAntiIdle() function as a parameter. This branch passes only the on/off bool as the parameter and I need to get the period value some other way.

I thought I was setting _antiIdlePeriod exactly the same way that _antiIdleCode is being set, but it doesn't seem to be working.

If I can get this version of the code to work, it will supercede the other pull request. But for now I'm keeping both branches until this branch works, because the other branch actually works.


sources/PTYSession.m
```diff
@@ -3072,29 +3077,25 @@ - (void)setBlend:(float)blendVal {
...
-    }
-
-    if (set) {
-        NSTimeInterval period = MIN(60, [iTermAdvancedSettingsModel antiIdleTimerPeriod]);
+- (void)setAntiIdle:(BOOL)set {
+    [_antiIdleTimer invalidate];
+    [_antiIdleTimer release];
+    _antiIdleTimer = nil;
+    
+    if ([self antiIdle]) {
```
gnachman: This should be `if (set) {` because `[self antiIdle]` indicates if a timer has already been started.

sources/PTYSession.m
```diff
@@ -3504,7 +3505,7 @@ - (void)scheduleUpdateIn:(NSTimeInterval)timeout {
 
 - (void)doAntiIdle {
     NSTimeInterval now = [NSDate timeIntervalSinceReferenceDate];
-    if (now >= _lastInput + 60) {
+    if (now >= _lastInput + _antiIdlePeriod) {   //FIXME
```
gnachman: I recommend:

````
static const NSTimeInterval kGracePeriod = 0.1;  // The timer runs at the same frequency as antiIdlePeriod, so if it's close we should write.
if (now >= _lastInput + _antiIdlePeriod - kGracePeriod) {
````

File: sources/ProfilesSessionPreferencesViewController.m [hunk 3]
```diff
@@ -110,6 +111,10 @@ - (void)awakeFromNib {
                            key:KEY_IDLE_CODE
                           type:kPreferenceInfoTypeIntegerTextField];
     info.range = NSMakeRange(0, 256);
+    
+    [self defineControl:_idlePeriod
+                           key:KEY_IDLE_PERIOD
+                          type:kPreferenceInfoTypeDoubleTextField];
```
gnachman: This should be disabled just like `_idleCode`. See how it uses the `observer` block on line 106.

sources/PTYSession.m
aljex: Ahh these are more interesting than I imagined. Thank you for the explanations.

aljex:
OK thank you very much for helping me figure this out. Now with your suggestions incorporated:

The UI is working. Both idle code and idle period fields are enabling/disabling based on the send idle code checkbox. The values are getting written to / read from the users local prefs immediately. The effects of all 3 options take effect immediately in the associated session(s).

The code is getting sent or not-sent based on the send idle code checkbox.
The code is getting sent on the schedule specified by idle period.

There is still one problem though which is that it still does not detect idleness. Set the period to 2 or 5 etc, run hexdump, and press spacebar continuously a few times per second, and the idle code will still appear every 2 or 5 seconds.

I'm looking now to try to see what all updates _lastInput.

aljex:
gah don't tell me I have to resolve collisions in the &^%&% xib file again ;)

aljex:
obsoleted, rebased.

PR #258, opened 2016-01-05
Cache filtered advanced settings

roandelyf:
This property is accessed multiple times per table view reload. We were unnecessarily
recomputing the same value hundreds of times.

This patch also caches a global sorted settings, as this value is constant and is accessed multiple times.

Advanced tab load time decreased dramatically.


gnachman:
Thanks!

PR #259, opened 2016-01-07
Improve navigation of the vertical/horizontal split modes in the right click menu

martinlindhe:


martinlindhe:
Perhaps this is opinionated, but here goes;

While I can perfectly figure out which is which, I still find myself thinking twice when seeing the words "vertical" and "horizontal". A little visual aid is helping me in this case.

gnachman:
This is a clever idea but I don't think it's quite right. Seems a little too unpolished (not to mention there are other menu items that would need to change).

martinlindhe:
Thanks, okay any pointers so i could try bring this patch into an acceptable state?

I saw in the OSX dev docs, one could embed icon images in the menu and was considering this. But none of the other menus have icons and i've no idea if it would be preferred :)

Also, which other menu items are you thinking of? The Shell -> Split horizontally ... options?

gnachman:
I don't see anything in Unicode that makes sense, so I think it would need to be an icon. I imagine a square that has been divided down the middle either vertically or horizontally—something really minimal. It looks like calling `-setImage:` on an NSMenuItem will add a graphic to a menu item in its left.

As for where it should go:
  * The context menu, which you already found
  * Four items in the *Shell* menu in `MainMenu.xib`.
  * Two items in `iTermEditKeyActionWindowController.xib` in the *Action* popup menu
  * Two buttons in `ProfileWindow.xib`

martinlindhe:
Thanks for the feedback. I'll see what I can come up with.

On a side-note, the screenshot below shows how Terminator (a linux terminal emulator app) does the same thing.
Do these icons look ok, then maybe i can start by trying to re-use them in iTerm?

terminator-menu

gnachman:
Seems a little heavy, but it's the right idea. I was thinking of the inverse—a thin box outline with a horizontal or vertical stroke through its center.

joaquinrulin:
Best PR ever! I keep confusing with this, and missing terminator little image but this is simpler and more effective.

martinlindhe:
Unfortunately, I won't have time to finish this off any time soon
Also, stuck in Windows atm so can't even enjoy iTerm :-(

Should we maybe turn this into a feature request and hope someone picks it up instead? I still think the idea is sound.

gnachman:
Yes, please open a feature request and make a reference to this PR.

martinlindhe:
Hm, the issue tracker is disabled on this repo. What do?

martinlindhe:
Thanks, feature request is at https://gitlab.com/gnachman/iterm2/issues/4875

PR #260, opened 2016-01-08
Anti idle 2

aljex:
Superseding previous pr's for the same feature. Rebased on current master branch.

* Fix: anti-idle period was ignoring user setting.<br>
  (a couple of different things combined to make neither less than nor greater than 60 possible)
* Fix: anti-idle was always sending, not actually detetcing activity and waiting for idleness. 
* Allow anti-idle more frequent than 60 seconds.
* Make settings changes take effect immediately instead of requiring a new session.
* Move anti-idle interval setting from app-wide advanced pref to per-profile pref.
  - Most sessions do not need or want any anti-idle nulls.
  - Some sessions need very frequent nulls.
  - Others need only infrequent nulls.
  - User can still set their own desired app-wide default behavior for new sessions by just setting the value in Default profile.
* Put all 3 anti-idle ui parts together in same panel, in Preferences>Profiles>Session


aljex:
I tried to do something to help out with the xib file.

It still has 500 changes in it, but if you look at the individual commits, there is one commit that has nothing but the 500 junk changes that xcode creates by itself.

I just opened the file, made a trivial change that was easy to find and discard/revert at commit time, and commit the rest.

Then the NEXT commit only has about 8 hunks, and 2 of those are in the header easy to ignore, leaving only about 6 hunks to look at and you can see every change that was actually part of this PR.

Other than that, for the PR as a whole, there is still the problem that it does not detect activity or idleness. Everything works as advertised except for that.

This is based on the the current master, and leaves in all the stuff you wanted left in.

Now I just have to find out why _lastInput isn't being updated, or figure out if that's even the problem.

aljex:
PTYSession.m keyDown()
certainly looks like it should be updating _lastInput on any non-empty keypress.

What we really care about is any tty activity at all though, so what do you think about something like this?:
PTYTask.m:
```obj-c
- (void)readTask:(char *)buffer length:(int)length {
    _lastTTYIO = [NSDate timeIntervalSinceReferenceDate];  // new
    [self logData:buffer length:length];
...
- (void)writeTask:(NSData*)data {
    _lastTTYIO = [NSDate timeIntervalSinceReferenceDate];  // new
    if (self.isCoprocessOnly) {
```
Also how do I get _lastTTY back in PTYSession.m where doAntiIdle() is?

aljex:
Simply changing doAntiIdle to use _lastOutput instead of _lastInput seems to have made it work.

Everything behaves as expected now.

Now all that's left (other than whatever you say ;) is:
* I want to address the one FIXME I still have in there just because.
* I don't know if updating _lastOutput has any undesirable effects yet. Like what else is watching _lastOutput and will this break it?

aljex:
As for the upgrade path issue, to carry the old deleted setting forward to the new setting:

I can confirm that the old setting didn't actually work anyways, in 2 different ways.

It was 60 seconds usually, with large and inconsistent variations sometimes. I had set the interval to 150 (chosen to be more than 2x 60, yet not a multiple of 60 either, to try to ensure when a code comes, I can tell for sure if it's because of the setting and not something to do with the hard coded 60 in there, or the logic of how/when the routine starts counting)  and then started a shell loop to time a "read" command reading 1 character and repeat. The first loop was 1:22, (ie: 82, not 60 or 120 or 150) the rest were mostly around 1:00 but with deviations. But never was it the specified 2:30.

That was without any keyboard or other input activity, so 100% idle.

It also didn't detect activity.

Starting hexdump and pressing the spacebar every 30 seconds or so, sometimes a null would get sent only 5 or 10 seconds after pressing enter or space several times. I don't have exact timings for keypresses vs anti-idle codes for the non-idle test, so I can't describe a real pattern there. Just that it is definitely not detecting activity.

Here's the screen capture of the fully idle test at least. The only input to the session was the anti-idle codes themselves. Any other keypress would have tripped the read and you'd see the character and the time here in the session.
This is after closing and re-opening iTerm to definitely reload the non-profile prefs.
```
Brians-Air:~ bkw$ defaults read com.googlecode.iterm2 |egrep '(Idle|Version)'
    AntiIdleTimerPeriod = 150;
[...]
    "iTerm Version" = "2.9.20160102";

Brians-Air:~ bkw$Brians-Air:~ bkw$Brians-Air:~ bkw$ while : ; do time read -n 1 ;done
^@
real	1m22.284s
user	0m0.000s
sys	0m0.000s
^@
real	1m0.001s
user	0m0.000s
sys	0m0.000s
^@
real	1m0.002s
user	0m0.000s
sys	0m0.000s
^@
real	1m10.001s
user	0m0.000s
sys	0m0.000s
^@
real	2m0.003s
user	0m0.000s
sys	0m0.000s
^@
real	2m0.003s
user	0m0.000s
sys	0m0.000s
^@
real	1m51.750s
user	0m0.000s
sys	0m0.000s
```

Given those 2 facts, does migrating the old setting really matter?

aljex:
It looks like updating _lastOutput in the anti-idle will prevent setLabelAttributesForIdleTabAtTime from ever happening, or might flip/flop it on/off/on/off forever depending on the combination of the users settings for both idlePeriod and idleTimeSeconds. Untested, just what it looks like.

So do you think I should maybe:
* copy _lastOutput to _lastOutputIncludingAntiIdle in PTYSession.m:init(), so init() updates them both the same from all normal traffic.
* isIdle & isProcessing continue to look at _lastOutput as before.
* doAntiIdle() only looks at, and only updates, _lastOutputIncludingAntiIdle ?

Or is it perfectly correct to let the anti-idle traffic end up suppressing the idle tab title and other idleness alerts, because in fact the session is not idle? In which case it's fine like it is?

I guess this depends on, what is the main purpose of setLabelAttributesForIdleTabAtTime ?
The session is being kept active by no-op traffic, but in every other way it is still idle in normal human terms.

(I'm going to go ahead with the separate variable idea while waiting)

aljex:
OK I made the 2nd variable and it works, but at first glance, updating _lastOutput doesn't seem to be hurting anything either.

I set idleTimerSeconds to 2, and idlePeriod to 4, and antiIdle to yes.
Started 2 tabs and switched to tab 2 and left tab 1 idle for more than 10 seconds.
In all that time the busy indicator icon never appeared on tab 1.
I started the ```while : ;do time read -n 1 ;done``` loop in tab 1 and switch back to tab 2 and waited.
This time every 4 seconds the activity spinner icon appears on tab 1.

This tells me that's at least one thing that that looks at _lastOutput that isn't adversely affected when anti-idle updates it.

So I'm not committing this change to add _lastOutputIncludingAntiIdle for now but I can if you want.

aljex:
Reverted that last. I guess the way it is now is the way it should be after all.

sources/PTYSession.m
```diff
@@ -223,6 +223,9 @@ @implementation PTYSession {
     // The code to send in the anti idle timer.
     char _antiIdleCode;
 
+    // The interval between sending anti-idle codes.
+    NSTimeInterval _antiIdlePeriod;
```
gnachman: It's not necessary to declare an ivar named `_antiIdlePeriod` when there is a property named `antiIdlePeriod`. It will be created automatically (this is called *automatic property synthesis*). Looks like you followed the model of `antiIdleCode`, which actually predates this language feature. Please remove both this and _antiIdleCode.

sources/PTYSession.m
```diff
@@ -3061,29 +3066,26 @@ - (void)setBlend:(float)blendVal {
...
-- (void)setAntiIdle:(BOOL)set
-{
-    if (set == [self antiIdle]) {
-        return;
+- (void)setAntiIdle:(BOOL)set {
+    [_antiIdleTimer invalidate];
+    [_antiIdleTimer release];
+    _antiIdleTimer = nil;
+    
+    // FIXME: corner case: part 1of2 of a kludgey way to prevent a flood when period is set to 1
```
gnachman: Let's remove this if statement…

sources/PTYSession.m
```diff
@@ -3061,29 +3066,26 @@ - (void)setBlend:(float)blendVal {
...
+        _antiIdlePeriod = 1.1;
     }
-
-    if (set) {
-        NSTimeInterval period = MIN(60, [iTermAdvancedSettingsModel antiIdleTimerPeriod]);
-
-        _antiIdleTimer = [[NSTimer scheduledTimerWithTimeInterval:period
+    
+    if (set && _antiIdlePeriod > 0) {
+        _antiIdleTimer = [[NSTimer scheduledTimerWithTimeInterval:_antiIdlePeriod
```
gnachman: And use MAX(_antiIdlePeriod, 1.1) here…

sources/PTYSession.m
```diff
@@ -3493,9 +3495,11 @@ - (void)scheduleUpdateIn:(NSTimeInterval)timeout {
 
 - (void)doAntiIdle {
     NSTimeInterval now = [NSDate timeIntervalSinceReferenceDate];
-    if (now >= _lastInput + 60) {
+    // _antiIdlePeriod-0.1 is to prevent missing every other cycle due to the timer calling us at the exact same period as we would check here.
+    // FIXME: corner case: _antiIdlePeriod>=1.1 is part 2of2 of a kludgey way to prevent a flood when period is set to 1.
+    if (now >= _lastOutput + _antiIdlePeriod - 0.1 && _antiIdlePeriod >= 1.1) {
```
gnachman: …and rewrite this as:

````
if (now >= _lastOutput + MAX(_antiIdlePeriod, kMinimumAntiIdlePeriod)) {
````

File: sources/PTYSession.m [hunk 5]
```diff
@@ -3493,9 +3495,11 @@ - (void)scheduleUpdateIn:(NSTimeInterval)timeout {
...
         [_shell writeTask:[NSData dataWithBytes:&_antiIdleCode length:1]];
-        _lastInput = now;
+        _lastOutput = now;
```
gnachman: I think `_lastInput` was correct. `_lastInput` tracks the last keystroke and `_lastOutput` tracks the last time stuff was received to display on the screen, escape sequences, etc. Sending the anti-idle code acts like a keystroke for the purposes of rate limiting the anti-idle code.

sources/iTermProfilePreferences.m
```diff
@@ -68,6 +68,17 @@ + (void)setFloat:(double)value
...
 
++ (double)doubleForKey:(NSString *)key inProfile:(Profile *)profile {
+        return [[self objectForKey:key inProfile:profile] doubleValue];
+}
+
++ (void)setDouble:(double)value
+           forKey:(NSString *)key
+        inProfile:(Profile *)profile
+            model:(ProfileModel *)model {
+        [self setObject:@(value) forKey:key inProfile:profile model:model];
```
gnachman: This line should be indented 4 spaces, not 8.

gnachman:
Sorry for the delay in getting back to you, was having various crises.

> I tried to do something to help out with the xib file.

Looks good to me. I'm sorry xib files are so bad.

> What we really care about is any tty activity at all though…

No, we only care about keystrokes. If the remote host is sending stuff, that doesn't count as idle. At least, that's how I understand people use it to prevent crappy networking equipment from dropping the connection on them.

> does migrating the old setting really matter
Yipes, I didn't realize how broken it was. But if someone has set it to be significantly faster/slower than 60 seconds, they'll be angry (and file bugs) when it's suddenly much different. Here's a diff that does it:

https://gist.github.com/gnachman/d16ad74dfec628a9817b

Regarding _lastOutput vs _lastInput, I believe the definition of idle is that you're not typing at the keyboard. The remote host could be chatting away but you're still idle. For example, you ssh somewhere else, and the ssh session is chatty, but the host you sshed from would still consider you idle and hang up on you. (Not that anti-idle works well for SSH—I'm not quite sure what it works well for—but it doesn't seem like a good idea to reduce its scope at this point).

sources/PTYSession.m
aljex: I'll do it of course, but for the record my reasoning was:

I am trying to avoid any arbitrary limits like that. It's not that I can think of a valid use-case for sending a continuous stream of nulls every 0.1 second, it's that it's none of my business to say 2 seconds is the shortest anyone else needs any more than 60 seconds was.

And the min would have to be 2 seconds to avoid that flood/race condition, which as a user just seems like an inexplicably arbitrary number. Taking 1 and internally nudging it up to 1.1 hides the problem and still works the same as 1.

Really I wanted to allow less than 1 second but I just couldn't figure out how to get it to accept floating point values, and whole seconds and a minimum of 1 second are both actually perfectly reasonable so I didn't pursue it.

I can see dispensing with the 0 option, because it makes more than one way to disable the feature. You have clicked the checkbox to enable sending, and yet nothing is sending. ok I get that.

2 seconds is actually well below my own lowest real value which is 10 seconds. Some connections are as aggressive as less than 1 minute, and you have to account for the fact that on some connections, 1 or 2 or even more packets might be lost and not make it to the aggressive component, and so in the worst cases you might need to send maybe every 10 seconds to ensure that you get 3 chances to get at least 1 packet all the way through within 30 seconds. So this still works for me, it's just slightly hokier to me from the users perspective. Why does it bottom out at 2?

sources/PTYSession.m
aljex: Why does this value need to be labeled and configurable?

sources/PTYSession.m
aljex: But lastInput doesn't work, and lastOutput does work. I understand the intended theory, but it doesn't actually work. "work" defined as it detects idleness and does not send the code when there is any other activity on the tty, which is what you want in a keepalive.

aljex:
Don't worry about how long it took you to look at this! I don't remember paying extra for rush service. ;)

You are describing a host process which is specifically caring about if the user is actively participating. That situation could exist, and so it should be allowed for, but I think that is by far the less usual case. It would definitely be not only not optimal but actively wrong in some of my own real use cases. Mostly my use cases don't care when the nulls are sent so they don't even have to detect idleness at all. But in those cases where properly detecting idleness does matter, it needs to be pacified by any traffic at all, and definitely not assume no-user-keypress=idle. Those come it 2 styles, each with their own different reason for needing it to work that way.

Remembering that there are tcp sessions, local serial sessions, and long chains of combinations of both, involving any unpredictable kind of software and embedded firmware of all different levels of sophistication and configurability:

For tcp sessions, all I've ever seen, the networking equipment does not care which direction traffic is flowing in a tcp session or what protocol or port number is being used. It just drops tcp sessions that haven't seen traffic. A network device can't say that a long download or stream that just doesn't happen to have back-traffic is somehow invalid. It's perfectly valid to use plain netcat to transfer a 9gig dvd iso and have that take hours. It's perfectly valid to have a data logger that just reads something like a sensor or the console output from a device forever.

Some network equipment basically violates some mechanisms/protocols that were meant to address exactly this need. They actually filter out TCP_KEEPALIVE, making that feature useless for it's explicit purpose, and so you have to generate in-band traffic, even though that is messy and might cause problems in some cases.

There IS a different thing that almost qualifies as "host looking for user activity" where a host daemon can detect idleness on the session and send an out-of-band "are you there?" packet to the client, and if the client doesn't respond, X times in a row, then the daemon concludes that the client's connection has broken or the client has crashed or basically the client has gone away ungracefully for whatever reason, and this is now a dead session that will never work again, and so it should be abandoned.

Those "are-you-there's" would also work to keep a tcp session alive in networking equipment, but you can't count on the host being configured to send them at all, let alone on the schedule you might need, nor can a mere terminal count on the user even using any particular kind of protocol or software to connect from one machine to another over a network, or even that the issue even involves any network.

Which leads to the other style, a serial connection:
The connection could be a local serial port, or raw netcat, or any number of any kinds of different things in a chain. I have had to ssh to one machine, to "telnet" from there to a serial port server appliance with a very crude telnet server embedded firmware that was really hardly more than netcat, to a serial port on that appliance, to the serial console of something with an even tinier/cruder embedded firmware like a remote control power strip, and had to do xmodem or even uucp file transfers over that.

A uucp download is a case where you do not want to send anything for a long time, while you are receiving-only for a long time, because, depending on the particular behavior of the "cu" program or the firmware of some device, the nulls might corrupt the download. And a uucp transfer is just simple one-way dump of traffic. No acks and checksums going back and forth as part of a chunked/windowed protocol like xmodem. It will be pure one direction traffic for possibly hours.

And yet you might still want the feature enabled in general, detecting idleness and sending nulls if the session does go idle, because either the firmware you are talking to or any part of the network between you and that console might be looking to kill the session, and that might be a problem even though the session would have stayed alive all during the actual transfer. For instance maybe it takes a long time for the host to start sending anything at all after you say "send me the stuff", or maybe having the connection drop after the transfer is done but before you perform some other step that comes next is a problem.

So for serial port sessions, it's also true that most of the time what you want is to be pacified by any traffic at all.

aljex:
As for xib files, reading up on it, I see apparently I have it good, and that before it was an xml file the problem was even worse. Lucky me. :)

Apparently 99% of the problem right now is just a disparity between different committers versions of xcode, where originally the <animations> tag didn't exist, and no problem. Then the <animations> tag was added and xcode automatically added it everywhere, creating diffs with 500 changes when one user has older xcode and another user has newer xcode. Then even newer xcode no longer adds the tag automatically, but does remove empty ones automatically, and so again it's only a problem while different committers are using different versions of xcode.

For my part I'm ok with continuing to do what I did in this PR. I don't know if on your end if you have a way of discarding one commit while accepting the rest. But even if not, at least this lets you and all other observers and future people looking at the history see the meaningful diff by itself (or copy or extract it for use outside of git), even if the merge still incorporates both diffs.

Although, this guy's post sure does sound good to me:
http://blog.teamtreehouse.com/why-i-dont-use-interface-builder

(...not having ever actually tried to do what he's talking about. Like how hard would it be to do all of iTerm2's ui in code? I don't know and I'm not volunteering to find out empirically.)

sources/PTYSession.m
aljex: I started to do this:
```obj-c
    static const int kMinimumAntiIdlePeriod = 2;
    
    _antiIdlePeriod = MAX(_antiIdlePeriod, kMinimumAntiIdlePeriod);
    
    if (set) {
        _antiIdleTimer = [[NSTimer scheduledTimerWithTimeInterval:_antiIdlePeriod
```

But shouldn't this really be in the ui? So they can't even enter 0 or 1, and it will never say 0 or 1 in the ui yet actually do 2. Not sure if you want that const defined somewhere else in a header or at the top of the file instead of right there in the function either.

aljex:
Ok most of that is adjusted as requested except lastOutput until we finish arguing that and/or fix lastInput so that it would actually function.

And I want to do something not spoken yet, which is make the UI not allow entering a 0 nor display a 0, since the code is no longer going to do a 0.

aljex:
Tried the patch you supplied but it didn't work. See the not on gistfile1.txt

gnachman:
Regarding the lower bound of 1 second: it's helpful to keep users from shooting themselves in the foot. Just because it's possible doesn't make it a good idea :)

Regarding using named constants for the grace period and minimum time: it's more for documentary purposes than to easy configuration.

Regarding lastInput not working (while lastOutput does): why not? It really ought to! Is it a problem with the grace period?

Regarding the definition of idleness: you make good points! I'm nervous about changing this because I don't know all the ways people use it. To redefine idleness more narrowly will break some users’ use cases, but I don’t see an offsetting benefit.

Regarding xib files sucking: I earnestly hope not to make any changes to the xib file in the near future, so hopefully no more rebasing will be necessary. I'm sorry it's so awful. I wish I had split the prefs panel into a dozen or more separate files, but it's super hard to do now. It's a terrible fit for manual layout because I often fiddle with positioning to get things looking nice.

Regarding limiting the minimum idle period in the UI: Yes, it should be limited there. You should be able to attach an NSNumberFormatter to the text field and specify a minimum value there. http://stackoverflow.com/questions/7479882/nstextfield-with-automatic-nsnumberformatter-in-interface-builder

Regarding my patch not working: let me look into that, will get back to you.

aljex:
Most of my original comments I subsequently deleted. For some bits I was flat wrong about what your suggestions would end up doing at run-time. After I did it and it all still worked exactly as before, I realized all that bloviating was idiotic. There's no 2 second minimum. The user can enter 1 just like before, and it will silently treat it as 1.1 inside just like before, and it all works fine. The only thing that's different is that now 0 doesn't disable it, 0 get's treated like 1.1, which is "good enough" to run with actually, but I'd like to get it to not allow entering or displaying a 0 setting now that it's not actually doing 0 behavior. But it's a small point now. The overall behavior is both better than before, and correct in my opinion.

And for other bits I just decided it made sense after 5 more minutes of thought, and besides, what do I care if a constant is named or just embedded? So I'm sorry you still ended up getting emails about those points. Maybe I should force myself to wait minimum 1 hour and/or until after I tried it, before responding, just as a blanket rule. ;)

The way lastInput does not work is, it does not detect keypresses and suppress keepalives. I don't know why not, because it looks like it should at least do what you are saying you want it to do. This is separate from me saying I don't think watching only the outgoing traffic is good enough. I'm saying it's like it's not seeing the outgoing traffic.

I'll look at it some more and see what I can see. The mystery might not be that deep, I just didn't even try to figure it out because I didn't think it was the right thing to do anyway.

aljex:
I found the numberformatter finally, and seem to have attached it to the field and set the min value. And it builds without error, and iterm runs, but when I try to open preferences, it just does the rainbow pinwheel spinner forever and prefs window never appears. But still better than an hour ago! So, getting there...

aljex:
There are a few different ways the idleperiod value can end up being 0 or empty or nil, despite the number formatter constraints in the ui. When that happens, we don't want it to use the lower bound of the allowable range, we want the default value (60).

Consider the user backspaces over the old value and then X's out of the dialog or force quits the entire app without ever writing in a new value. In that case the most sane thing is not to send codes every 1.1 seconds but to either disable the feature entirely, or use the default of 60 seconds. And since there is already a dedicated bool to enable/disable the feature, and this question only even arises when the user has explicitly enabled the feature, that leaves the most sane behavior is to send every 60 seconds not every 1.1 seconds.

gnachman:
Sorry for sending you down the path of NSNumberFormatter. I forgot how broken it is on editable text fields. I've cleaned up the rough edges and merged this as commit 17319d9a442b5f3286b286f79ab4ef4df8fbe610. Thanks for all the work on this one!

aljex:
Terrific! Thank you!

PR #261, opened 2016-01-12
Fix stack overflow in indicatorRectForTabCell: on startup

jgavris:
- On startup, a tab won't have a count, but it may have an indicator


gnachman:
Thanks for letting me know about this.
Fixed in a0c05fa.

jgavris:
Thanks, first time looking under the hood, didn't think I had that right.

PR #262, opened 2016-01-13
Fixes for Xcode 7.3 / clang 702.0.25

jgavris:
- Can no longer use __weak references in MRC files


gnachman:
I don’t think there’s any reason to nil out weak references in dealloc, so I've applied this change without that part. But please let me know if there’s something I’m not thinking of!

PR #264, opened 2016-01-19
Re-enable disabled tests (iTermShellHistoryController)

jgavris:


gnachman:
Thanks! That was much easier than I expected :)

PR #269, opened 2016-02-18
Fix a bunch of leaks

ronyfadel:


sources/SessionView.m
```diff
@@ -50,7 +50,9 @@ + (double)titleHeight {
 }
 
 + (void)initialize {
-    lastResizeDate_ = [[NSDate date] retain];
+    if (self == [SessionView self]) {
```
gnachman: I gather this is to prevent potential subclassing-related problems?

I've also never seen the `+self` class method. Is it the same as `+class`?

gnachman:
Great catches, thanks. Did you find these with static analysis, Leaks Instrument, or something else?

sources/SessionView.m
ronyfadel: Yeah, KVO autosubclasses our class, and gets +initialize called again.

The self surprised me too, but I found it neat. I got it from the docs:

https://developer.apple.com/library/mac/documentation/Cocoa/Reference/Foundation/Classes/NSObject_Class/#//apple_ref/occ/clm/NSObject/initialize

IMO they're strictly equivalent, but for the nitpicker, self is more correct since it returns *the* object being sent the message, whereas class returns *a* class object (that just happens to be that same object)

ronyfadel:
Leaks Instrument :)

PR #270, opened 2016-02-18
Call malloc once instead of 3 times in CRunStorage everytime we alloc…

ronyfadel:
…ate elements

malloc is potentially called hundreds of times when clearing the display
This has the added advantage of reduced memory fragmentation

### DIFF
--- a/sources/CharacterRun.m
+++ b/sources/CharacterRun.m
@@ -9,7 +9,14 @@
 #import "CharacterRun.h"
 #import "ScreenChar.h"
 
-@implementation CRunStorage : NSObject
+#define SIZEOF_CODES    sizeof(*codes_)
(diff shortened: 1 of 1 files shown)


gnachman:
Have you tried profiling to see if this makes a difference? I like avoiding calls to malloc but am concerned about adding extra complexity.

ronyfadel:
I'm investigating iTerm's memory usage.. Found some low hanging fruits.
I should profile this, and report back.

gnachman:
I ran some performance tests and didn't see a measurable difference. This was a good idea, but I think malloc is actually pretty fast compared to everything else that's going on during drawing so any benefit is dwarfed. This code is all going to go away in 3.1, anyway, so I'm going to close this out to avoid adding extra complexity.

ronyfadel:
sure!

PR #271, opened 2016-02-19
imgls usage - fix to match the current script name

tskirvin:
imgls, not image.sh

### DIFF
--- a/tests/imgls
+++ b/tests/imgls
@@ -20,7 +20,7 @@ function print_st() {
 }
 
 if [ $# -eq 0 ]; then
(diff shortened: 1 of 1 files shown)


gnachman:
I actually just changed it to list the files in the current directory in this case.

tskirvin:
Your patch doesn't work nicely with files with spaces in them (like, say, the upload directory from an iPhone into a Dropbox directory).  I don't know the right workaround off the top of my head, sorry...

gnachman:
Good catch. Fixed in 508fc22.

tskirvin:
Yep, that does the job.  Thanks!

PR #272, opened 2016-02-22
Print the image names (ie when using imgcat *.jpg)

christophe-f:


gnachman:
The `imgls` tool is for this use case, while `imgcat` is simply to display an image.

christophe-f:
thanks, didn't know about this script.

PR #275, opened 2016-03-13
Rony/memory usage

ronyfadel:
One thing I'm not too happy about is the following:

-    [self performSelector:@selector(callObserver) withObject:nil afterDelay:0];

Can we find a cleaner pattern for this, instead of performing after delay?

Thanks!

### DIFF
--- a/sources/PTYSession.m
+++ b/sources/PTYSession.m
@@ -3735,8 +3735,8 @@ static const NSTimeInterval kAntiIdleGracePeriod = 0.1;
         [[ProfileModel sessionsInstance] setBookmark:[self profile] withGuid:guid];
 
         // Update an existing one-bookmark prefs dialog, if open.
-        if (PreferencePanel.sessionsInstance.windowIfLoaded.isVisible) {
-            [PreferencePanel.sessionsInstance underlyingBookmarkDidChange];
+        if ([PreferencePanel sessionsInstance].windowIfLoaded.isVisible) {
+            [[PreferencePanel sessionsInstance] underlyingBookmarkDidChange];
         }
     }
 }
--- a/sources/PTYSession.m
+++ b/sources/PTYSession.m
@@ -3735,8 +3735,8 @@ static const NSTimeInterval kAntiIdleGracePeriod = 0.1;
         [[ProfileModel sessionsInstance] setBookmark:[self profile] withGuid:guid];
 
         // Update an existing one-bookmark prefs dialog, if open.
-        if ([[[PreferencePanel sessionsInstance] window] isVisible]) {
-            [[PreferencePanel sessionsInstance] underlyingBookmarkDidChange];
+        if (PreferencePanel.sessionsInstance.windowIfLoaded.isVisible) {
+            [PreferencePanel.sessionsInstance underlyingBookmarkDidChange];
         }
     }
 }
--- a/sources/PTYTextView.m
+++ b/sources/PTYTextView.m
@@ -710,10 +710,10 @@ static const int kDragThreshold = 3;
 
 - (void)changeFont:(id)fontManager
 {
-    if (PreferencePanel.sharedInstance.windowIfLoaded.isVisible) {
-        [PreferencePanel.sharedInstance changeFont:fontManager];
-    } else if (PreferencePanel.sessionsInstance.windowIfLoaded.isVisible) {
-        [PreferencePanel.sessionsInstance changeFont:fontManager];
+    if ([PreferencePanel sharedInstance].windowIfLoaded.isVisible) {
+        [[PreferencePanel sharedInstance] changeFont:fontManager];
+    } else if ([PreferencePanel sessionsInstance].windowIfLoaded.isVisible) {
+        [[PreferencePanel sessionsInstance] changeFont:fontManager];
     }
 }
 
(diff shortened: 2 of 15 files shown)


sources/PTYSession.m
```diff
@@ -3735,8 +3735,8 @@ - (void)changeFontSizeDirection:(int)dir
         [[ProfileModel sessionsInstance] setBookmark:[self profile] withGuid:guid];
 
         // Update an existing one-bookmark prefs dialog, if open.
-        if ([[[PreferencePanel sessionsInstance] window] isVisible]) {
-            [[PreferencePanel sessionsInstance] underlyingBookmarkDidChange];
+        if (PreferencePanel.sessionsInstance.windowIfLoaded.isVisible) {
```
gnachman: I'd prefer not to introduce stylistic inconsistency here; class methods are invoked with square brackets in the rest of the code.

sources/PTYTextView.m
```diff
@@ -710,10 +710,10 @@ - (void)setFont:(NSFont*)aFont
 
 - (void)changeFont:(id)fontManager
 {
-    if ([[[PreferencePanel sharedInstance] window] isVisible]) {
-        [[PreferencePanel sharedInstance] changeFont:fontManager];
-    } else if ([[[PreferencePanel sessionsInstance] window] isVisible]) {
-        [[PreferencePanel sessionsInstance] changeFont:fontManager];
+    if (PreferencePanel.sharedInstance.windowIfLoaded.isVisible) {
```
gnachman: Here too.

sources/PreferenceInfo.m
```diff
@@ -34,6 +34,8 @@ - (void)dealloc {
     [_shouldBeEnabled release];
     [_onChange release];
     [_customSettingChangedHandler release];
+    [_onUpdate release];
```
gnachman: Good catch! :+1:

sources/PreferenceInfo.m
```diff
@@ -42,10 +44,10 @@ - (void)setObserver:(void (^)())observer {
     _observer = [observer copy];
     // Call the observer after a delayed perform so that the current profile can be set and then the
     // control's value gets initialized.
-    [self performSelector:@selector(callObserver) withObject:nil afterDelay:0];
+    [self _callObserver];
```
gnachman: I'm not convinced we can get rid of the delayed perform. Every observer would need to be analyzed to ensure no assumptions were broken. I'm not crazy about using a delayed perform, either, but it is nice and simple. One alternative would be to run all the observers after the prefs window finishes loading, perhaps by notification.

sources/PreferencePanel.m
```diff
@@ -139,24 +140,33 @@ @implementation PreferencePanel {
...
-    });
-    return instance;
+    if (!_sessionsInstance) {
+        _sessionsInstance = [[self alloc] initWithProfileModel:[ProfileModel sessionsInstance]
+                                        editCurrentSessionMode:YES];
+    }
+    return _sessionsInstance;
+}
+
++ (void)_releaseAndNilifySingletonInstancesIfNeeded:(PreferencePanel *)instance {
```
gnachman: It's pretty weird to release a singleton. This only gets you back about 30-50 megs of ram, so I'm not sure it's worth the added complexity.

gnachman:
I'm not quite sure why, but the prefs panel sharedInstance now looks just like the sessionsInstance (i.e., *iTerm2>Preferences* looks like *Edit>Edit Session*.

sources/PTYTextView.m
ronyfadel: Fixed in subsequent commits

sources/PreferencePanel.m
ronyfadel: Moved to properties on app delegate.
For both panels combined, that's ~50-80 megs of ram saved.
That's both faster startup times, and shrinking app memory usage by ~50% with 1-2 tabs open.

ronyfadel: Moved to properties on app delegate.
For both panels combined, that's ~50-80 megs of ram saved.
That's both faster startup times, and shrinking app memory usage by ~50% with 1-2 tabs open.

sources/PreferencePanel.m
```diff
@@ -220,14 +212,34 @@ - (BOOL)importColorPresetFromFile:(NSString*)filename {
...
+    }
+    return nil;
+}
+
 - (WindowArrangements *)arrangements {
     return arrangements_;
 }
 
 - (void)run {
+    (void)self.window;
```
gnachman: Does this need to be done earlier than it used to be?

gnachman:
I made some mostly style changes in my reduceMemoryUsageForPreviewPanelsAndQuicklook branch. Want to take a look and see if it makes sense? I was able to make a few things a little simpler.

ronyfadel:
I'm sorry for my absence, I was on a github hiatus.
I've commented you commit. Thanks!

gnachman:
I'm closing this PR as it's moved to the reduceMemoryUsageForPreviewPanelsAndQuiclook branch.

gnachman:
@ronyfadel I'm looking into integrating this branch now and I've run into a problem where the prefs window doesn't save & restore its frame properly. Because the window gets dealloced each time it's closed, it goes through the new window creation code path and gets a different frame given to it by the system each time.

gnachman:
nm, I worked around it. Merged in commit 7543187.

ronyfadel:
lgtm, thanks!

PR #276, opened 2016-03-13
overflowImage.tiff and overflowImagePressed.tif aren't retina-ready

ronyfadel:
overflowImage.tiff and overflowImagePressed.tif weren't retina-ready
took a stab at a lighter design


gnachman:
Thanks!

PR #277, opened 2016-04-11
Feature - Toggle: Only apply transparency to default bg

polarathene:
This fixes a recent issue I raised [here](https://gitlab.com/gnachman/iterm2/issues/4523) and it's related [June 2013 issue](https://gitlab.com/gnachman/iterm2/issues/2478). The problem is also brought up in other repo's and communities. The main advantage being Powerline themes that don't break from having transparent backgrounds.

I've added a UI element to opt-in to the feature, it should not cause any  problems for those not interested(Although it's great for other cases such as messages like session restore text being more readable on a solid grey instead of transparent ;) ). See each commit message for more information :)

Here's a comparison with and without:
defaultAlphaBGOnly feature

### DIFF
--- a/Interfaces/PreferencePanel.xib
+++ b/Interfaces/PreferencePanel.xib
@@ -2615,11 +2615,11 @@ DQ
             <point key="canvasLocation" x="-62.5" y="288.5"/>
         </customView>
         <customView id="77f-eM-Dma" userLabel="Prefs - Profiles - Window" customClass="iTermSizeRememberingView">
-            <rect key="frame" x="0.0" y="0.0" width="559" height="424"/>
+            <rect key="frame" x="0.0" y="0.0" width="559" height="437"/>
             <autoresizingMask key="autoresizingMask" widthSizable="YES" heightSizable="YES"/>
             <subviews>
                 <textField verticalHuggingPriority="750" id="4911">
-                    <rect key="frame" x="94" y="174" width="130" height="22"/>
+                    <rect key="frame" x="94" y="187" width="130" height="22"/>
                     <autoresizingMask key="autoresizingMask" flexibleMaxX="YES" flexibleMinY="YES"/>
                     <textFieldCell key="cell" scrollable="YES" lineBreakMode="clipping" selectable="YES" editable="YES" continuous="YES" sendsActionOnEndEditing="YES" state="on" borderStyle="bezel" alignment="left" drawsBackground="YES" usesSingleLineMode="YES" id="4951">
                         <font key="font" metaFont="system"/>
@@ -2632,7 +2632,7 @@ DQ
                     </connections>
                 </textField>
                 <textField verticalHuggingPriority="750" id="4912">
-                    <rect key="frame" x="28" y="174" width="67" height="19"/>
+                    <rect key="frame" x="28" y="187" width="67" height="19"/>
                     <autoresizingMask key="autoresizingMask" flexibleMaxX="YES" flexibleMinY="YES"/>
                     <textFieldCell key="cell" sendsActionOnEndEditing="YES" alignment="left" title="Columns:" id="4950">
                         <font key="font" metaFont="system"/>
@@ -2641,7 +2641,7 @@ DQ
                     </textFieldCell>
                 </textField>
                 <textField verticalHuggingPriority="750" id="4913">
-                    <rect key="frame" x="51" y="144" width="44" height="19"/>
+                    <rect key="frame" x="51" y="157" width="44" height="19"/>
                     <autoresizingMask key="autoresizingMask" flexibleMaxX="YES" flexibleMinY="YES"/>
                     <textFieldCell key="cell" sendsActionOnEndEditing="YES" alignment="left" title="Rows:" id="4949">
                         <font key="font" metaFont="system"/>
@@ -2650,7 +2650,7 @@ DQ
                     </textFieldCell>
                 </textField>
                 <textField verticalHuggingPriority="750" id="4914">
-                    <rect key="frame" x="94" y="144" width="130" height="22"/>
+                    <rect key="frame" x="94" y="157" width="130" height="22"/>
                     <autoresizingMask key="autoresizingMask" flexibleMaxX="YES" flexibleMinY="YES"/>
                     <textFieldCell key="cell" scrollable="YES" lineBreakMode="clipping" selectable="YES" editable="YES" sendsActionOnEndEditing="YES" state="on" borderStyle="bezel" alignment="left" drawsBackground="YES" id="4947">
                         <font key="font" metaFont="system"/>
@@ -2663,7 +2663,7 @@ DQ
                     </connections>
                 </textField>
                 <textField verticalHuggingPriority="750" id="4992">
-                    <rect key="frame" x="15" y="204" width="190" height="20"/>
+                    <rect key="frame" x="15" y="217" width="190" height="20"/>
                     <autoresizingMask key="autoresizingMask" flexibleMaxX="YES" flexibleMinY="YES"/>
                     <textFieldCell key="cell" sendsActionOnEndEditing="YES" alignment="left" title="Settings for New Windows" id="4993">
                         <font key="font" metaFont="systemBold"/>
@@ -2672,7 +2672,7 @@ DQ
                     </textFieldCell>
                 </textField>
                 <popUpButton verticalHuggingPriority="750" id="4916">
-                    <rect key="frame" x="324" y="172" width="221" height="26"/>
+                    <rect key="frame" x="324" y="185" width="221" height="26"/>
                     <autoresizingMask key="autoresizingMask" flexibleMaxX="YES" flexibleMinY="YES"/>
                     <popUpButtonCell key="cell" type="push" title="Fullscreen" bezelStyle="rounded" alignment="left" lineBreakMode="truncatingTail" state="on" borderStyle="borderAndBezel" tag="1" imageScaling="proportionallyDown" inset="2" selectedItem="4944" id="4941">
                         <behavior key="behavior" lightByBackground="YES" lightByGray="YES"/>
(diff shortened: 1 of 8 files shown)


unphased:
Really cool :smile:

Interfaces/PreferencePanel.xib
```diff
@@ -2902,30 +2903,43 @@ DQ
...
                         <behavior key="behavior" changeContents="YES" doesNotDimImage="YES" lightByContents="YES"/>
                         <font key="font" metaFont="system"/>
                     </buttonCell>
                     <connections>
                         <action selector="settingChanged:" target="BMA-OA-Kzc" id="VYD-F1-fDI"/>
-                        <outlet property="nextKeyView" destination="4978" id="6418"/>
+                        <outlet property="nextKeyView" destination="fwj-4Y-nja" id="u8P-d1-x0a"/>
+                    </connections>
+                </button>
+                <button toolTip="Enable this to prevent background colors for text from being affected by the transparency setting. Useful for having a transparent terminal background but opaque Powerline segments(background) to match opaque Powerline separators(foreground)." id="fwj-4Y-nja">
```
gnachman: This copy is a little stilted. Try:

````
Enable this to make non-default background colors always opaque. This is useful when using Powerline in a translucent window.
````

File: sources/ITAddressBookMgr.h [hunk 2]
```diff
@@ -139,6 +139,7 @@
 // Terminal
 #define KEY_DISABLE_WINDOW_RESIZING           @"Disable Window Resizing"
 #define KEY_PREVENT_TAB                       @"Prevent Opening in a Tab"
+#define KEY_DEFAULT_BG_ALPHA_ONLY             @"Only The Main BG Color Uses Transparency"
```
gnachman: Change "Main" to "Default"

sources/PTYSession.m
```diff
@@ -3098,6 +3100,11 @@ - (void)setBlend:(float)blendVal {
     [_textview setBlend:blendVal];
 }
 
+- (void)setDefaultBGAlphaOnly:(BOOL)value
```
gnachman: I don't generally use abbreviations (although you'll find some in very old code). Please name this `setTransparencyAffectsOnlyDefaultBackgroundColor:` (and make the corresponding change in various other places)

sources/ProfilesWindowPreferencesViewController.m
```diff
@@ -119,6 +120,10 @@ - (void)awakeFromNib {
                     key:KEY_PREVENT_TAB
                    type:kPreferenceInfoTypeCheckbox];
 
+   [self defineControl:_defaultBGAlphaOnly
+                   key:KEY_DEFAULT_BG_ALPHA_ONLY
+                  type:kPreferenceInfoTypeCheckbox];
```
gnachman: This should be disabled when transparency is off. Assign the result of this method call to `info` and then do
````
    info.observer = ^() { _defaultBGAlphaOnly.enabled = (_transparency.doubleValue > 0); };
````

(look at how the blur slider gets disabled for what an example of what I mean)

**Comment by polarathene** [comment]
@gnachman Updated with your change requests.

Did you want to update the key `KEY_DEFAULT_BG_ALPHA_ONLY` as well? It would become a very long key which seems unnecessary since following it's few occurrences will connect it to the expanded name `transparencyAffectsOnlyDefaultBackgroundColor`.

File: Interfaces/PreferencePanel.xib [hunk 5]
```diff
@@ -2902,30 +2903,43 @@ DQ
...
                     <connections>
                         <action selector="settingChanged:" target="BMA-OA-Kzc" id="VYD-F1-fDI"/>
-                        <outlet property="nextKeyView" destination="4978" id="6418"/>
+                        <outlet property="nextKeyView" destination="fwj-4Y-nja" id="u8P-d1-x0a"/>
+                    </connections>
+                </button>
+                <button toolTip="Enable this to make non-default background colors always opaque. This is useful when using Powerline in a translucent window." id="fwj-4Y-nja">
+                    <rect key="frame" x="15" y="9" width="389" height="18"/>
+                    <autoresizingMask key="autoresizingMask" flexibleMaxX="YES" flexibleMinY="YES"/>
+                    <buttonCell key="cell" type="check" title="Only apply transparency to the main background color." bezelStyle="regularSquare" imagePosition="left" alignment="left" inset="2" id="sBm-XA-tIi">
```
gnachman: Change "main" to "default" here

sources/ITAddressBookMgr.h
```diff
@@ -139,6 +139,7 @@
 // Terminal
 #define KEY_DISABLE_WINDOW_RESIZING           @"Disable Window Resizing"
 #define KEY_PREVENT_TAB                       @"Prevent Opening in a Tab"
+#define KEY_DEFAULT_BG_ALPHA_ONLY             @"Only The Default BG Color Uses Transparency"
```
gnachman: Yes, let's change the constant too. It's confusing when the names don't match up, and long identifiers are a fact of life in objective c.

gnachman:
Please rebase so I can merge this after the next batch of changes.

polarathene:
This was my first rebase with a pull request. Hopefully I did it right with the force push.

I missed a file with the last commit, I assume it's ok to ammend and force push in this case.

```
/Users/travis/build/gnachman/iTerm2/sources/PTYSession.m:3059:16: 'PTYTextView' may not respond to 'setTransparencyAffectsOnlyDefaultBackgroundColor:'
```

I'm not sure why Travis/XCode is warning about this, is it due to length?

gnachman:
Merged, thanks!

Tenchi2xh:
Awesome feature, was waiting for it! Thanks a lot :)

unphased:
I'm building in xcode but I can't seem to find the option under the Profiles/Window tab. Help!

polarathene:
@unphased My commit is in the master branch history, it's a single merge commit, you can see it [here]( https://github.com/gnachman/iTerm2/commit/ffe69593b8c42e8f2cc72900c9d26d0c559d6bd6)

[Here](https://github.com/gnachman/iTerm2/compare/ffe69593b8c42e8f2cc72900c9d26d0c559d6bd6...233df34d3bb098d4d0e11e43e41e390f411e9fbd?diff=split&name=233df34d3bb098d4d0e11e43e41e390f411e9fbd#diff-335cffbb56d78c6407e3ff306c98233eL61) is a compare view between that commit and the latest commit on master branch currently. I've had it select/view a line that has introduced a small refactor to my work since it was committed. There doesn't appear to be anything else deleted/modified since then besides moving a `{` to the same line as it's method declaration.

Looking at the linked files [history](https://github.com/gnachman/iTerm2/commits/233df34d3bb098d4d0e11e43e41e390f411e9fbd/sources/ProfilesWindowPreferencesViewController.m), right now there is just a single commit after this PR's merge commit. Build from master branch at the [merge commit](https://github.com/gnachman/iTerm2/commit/ffe69593b8c42e8f2cc72900c9d26d0c559d6bd6#diff-335cffbb56d78c6407e3ff306c98233e) and see if that works for you, if it does, then build the [commit that refactored that code](https://github.com/gnachman/iTerm2/commit/3d8f56356ce8a062f33699474201b85fef8b189e#diff-335cffbb56d78c6407e3ff306c98233e). If it no longer works, then a bug was introduced.

unphased:
@polarathene I see it now. All that was needed to confuse me was that the new setting shows up in a new spot near the top under the Transparency slider (which is not in the Misc section as you show in your original branch). 

Everything's working great.

polarathene:
@unphased I've yet to update the official iTerm2, I wasn't aware it had also been relocated, sorry about that.

gnachman:
I moved some things around and made other improvements to that settings pane, sorry for not mentioning it in here.

sassanh:
@polarathene @gnachman Thanks for this, it works great. But I guess we need something like this for dimming panes too, look at this:
https://www.dropbox.com/s/fq8h0a80t13gxno/Screenshot%202016-07-29%2023.07.17.png?dl=0

gnachman:
Would be garish for the red to not get dimmed, don't you think?

polarathene:
@gnachman I don't think @sassanh meant that. The dimming is inconsistent between the background and foreground colours, just like with the transparency patch I made in this PR. 

I have an animated example at the top that shows the segments/separators inconsistency when applying transparency. While that favours visibility by not making them transparent, @sassanh is probably wanting the colours to remain consistent for the segment/separators, so both should dim?

I no longer have access to a Mac or use iterm2 at present, so I'm unable to look into this.

sassanh:
@polarathene yeah, that's what I meant.

sassanh:
What's wrong with powerline fonts? I mean why they don't behave like normal fonts?

gnachman:
Oh, I see, the issue is that the left arrow isn't dimmed the same as the red to the right of it.

gnachman:
Hm, dims OK for me, but I'm sure our settings are different. @sassanh please file an issue at https://iterm2.com/bugs.

screen shot 2016-08-06 at 9 48 36 am

sassanh:
Sure!

nneekro:
@polarathene i can't find it. Is it removed?

nneekro:
@polarathene I can't find the toggle. Is it removed?

polarathene:
@nneekro I no longer have access to a system with macOS and haven' t used iTerm2 for several years as a result, so no clue :confused:

gnachman:
Moved here: 
image

nneekro:
@gnachman thanks

EduardoOliveira:
Use transparency has no effect, now some glyphs are transparent and other solid...
image

gnachman:
@EduardoOliveira Are you referring to the triangles at the bottom? They should be opaque because they're foregorund, not background. You might want to turn on "Keep background colors opaque" under "transparency"

EduardoOliveira:
That's it thanks

PR #278, opened 2016-04-14
Increases active window refresh rate to 60Hz

csmulhern:
This leads to a much nicer user experience. I type reasonably quickly and have a high key repeat rate and I'm frequently bothered by the perceived input latency.

### DIFF
--- a/sources/PTYSession.m
+++ b/sources/PTYSession.m
@@ -166,7 +166,7 @@ static const NSTimeInterval kAntiIdleGracePeriod = 0.1;
 
 // Timer period between updates when active (not idle, tab is visible or title bar is changing,
 // etc.)
-static const NSTimeInterval kActiveUpdateCadence = 1.0 / 30.0;
+static const NSTimeInterval kActiveUpdateCadence = 1.0 / 60.0;
 
 // Timer period for background sessions. This changes the tab item's color
 // so it must run often enough for that to be useful.


johannbotha:
+1 

This change makes tmux integration with many windows much smoother.

Baekalfen:
+1 Also noticed the input lag the other day. I wouldn't have thought somebody had chosen a refresh rate lower than 60hz.

gnachman:
The reason I chose a lower frame rate is to improve performance when printing a big file. I just benchmarked again and it doesn't seem to make much difference, though, so let's give 60fps a try.

felixbuenemann:
Maybe iTerm could lower the framerate when scrolling large amounts of text, but increase it while typing? 

Another alternative would be to have a lower refresh rate for windows/panes/tabs that are not in focus.

Currently I often redirect text output to a file and tail it with `less` just to keep iTerm cpu usage low (less updates the screen at a much lower interval).

gnachman:
@felixbuenemann I've tried that in the past and it caused all kinds of hard-to-fix bugs. Does the CPU usage differ much for you between these builds when printing a big file? The first is 30fps and the second is 60fps:

http://iterm2.com/downloads/nightly/iTerm2-2_9_20160416-nightly.zip
http://iterm2.com/downloads/nightly/iTerm2-2_9_20160417-nightly.zip

Drawing is only a small component of what happens; token parsing and updating the scrollback buffer use a lot of CPU too.

felixbuenemann:
I was using the `less` technique before the switch to 60hz, so I wasn't referring to the new build.

**Nevertheless I just did some fairly unscientific benchmarking:**

I compared build 20160416 to 201417 and 20160419 and the results between test runs were so wildly different that it's hard to say wether the switch to 60 fps really made a performance difference although at smaller window sizes the 30 fps build was consistently faster.

It's notable that iTerm performance depends very heavily on how big the terminal window is. I also checked Terminal.app, which is much faster and also suffers less of a performance hit for increasing window sizes.

@gnachman I have posted the [benchmark results as a gist](https://gist.github.com/30df9c14df46b88ae52bf4e8cc0f1404), because the post became much too long and potentially off-topic for this issue. It also contains some interesting comparisons between Terminal's and iTerm2's behavior in different situations.

Btw. I measured the actual screen refresh rate using the FrameMeter from Quartz Debug.

gnachman:
@felixbuenemann Thanks for the analysis. This part caught my eye: "It's notable that at the less taxing 80x25 window size, the 60 fps build is consistently much slower than the 30 fps build, even though it manages to update the screen at ~50 fps.". What was the wall clock difference that you saw between the two versions?

gnachman:
This issue suggests that at least for some people 60fps affects performance significantly:

https://gitlab.com/gnachman/iterm2/issues/4559

felixbuenemann:
30 fps: find /Volumes/FELIX4TBA/Restore  0,09s user 0,36s system 24% cpu 1,821 total
60 fps: find /Volumes/FELIX4TBA/Restore  0,09s user 0,35s system 3% cpu 11,672 total

total is the walltime.

gnachman:
Oh, the comma threw me off. Yeah, that's pretty bad. I'm going to revert this for now, but maybe it can come back in 3.1 if it improves drawing performance enough.

felixbuenemann:
Sorry, should've used `LOCALE=C`.

gnachman:
I should be less anglcentric :). Reverted to 30 fps in commit 0211a90.

csmulhern:
Would you prefer if I made this an advanced setting?

felixbuenemann:
@csmulhern That sound like a good idea, if it's possible. It would be great if it allowed a number instead of a toggle between 30/60 fps.

gnachman:
@csmulhern Yeah, advanced setting would be fine.

gnachman:
Commit 233df34 adds adaptive frame rates back. It is off by default. Enable it with **Prefs > Advanced > Use adaptive framerate**.

felixbuenemann:
@gnachman Cool, seems to work as far as I can tell. Could you make the lower framerate configurable? That would allow to choose between smooth scrolling and low cpu usage.

gnachman:
This is still kind of a hack. A better version of adaptive framerate will come that shouldn't require configuration.

PR #280, opened 2016-05-01
Setup for custom title delay settings.

clementpadovani:
Added the possibility to change the delay for the app to set a tab’s title.

This can be achieved by running the following command:

``` bash
defaults write com.googlecode.iterm2 DelayedTitleCustomDelayTimeInterval -float 0.5
```

If no value is set in the defaults then the default value of `0.1` is used.

### DIFF
--- a/sources/iTermDelayedTitleSetter.m
+++ b/sources/iTermDelayedTitleSetter.m
@@ -7,11 +7,18 @@
 //
 
 #import "iTermDelayedTitleSetter.h"
+#import "DebugLogging.h"
 
 NSString *const kDelayedTitleSetterSetTitle = @"kDelayedTitleSetterSetTitle";
 NSString *const kDelayedTitleSetterTitleKey = @"title";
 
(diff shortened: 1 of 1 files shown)


gnachman:
Out of curiosity why would you want to do this?

ClementPadovani:
Hey @gnacgman,

I missed the instant process name change after migrating from Terminal to iTerm.

gnachman:
I'd be surprised if a .1 second delay were perceptible. The lion's share of the delay is that Terminal does some undocumented magic to get the process name instantly, whereas we have to poll periodically every 1/2 second. If you really want to speed it up increase the polling interval in ProcessCache.m line 353, but it'll cost you some CPU.

ClementPadovani:
Alright, thanks @gnachman for your help. I’ll try messing around with `ProcessCache` when I have more time.

TBH it’s really a detail and it’s some of my classmates noticed the difference when switching from the stock term to iTerm that’s all.

PR #281, opened 2016-05-02
download.sh fix and small improvement

aka-rider:
Hello.

Thanks for the great app.

### DIFF
--- a/tests/download.sh
+++ b/tests/download.sh
@@ -1,11 +1,12 @@
 #!/bin/bash
-if [ $# -ne 1 ]; then
+if [ $# -lt 1 ]; then
   echo "Usage: download.sh file ..."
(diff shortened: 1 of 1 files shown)


tests/download.sh
```diff
@@ -1,11 +1,12 @@
 #!/bin/bash
-if [ $# -ne 1 ]; then
+if [ $# -le 1 ]; then
```
gnachman: In the old code, 1 argument was ok (while more than 1 should have been but wasn't). With this test 1 argument would fail. This should be `-lt` not `-le`.

aka-rider: Yep, my bad.

aka-rider:
I updated pull request

PR #282, opened 2016-05-08
scroll wheel sends escape sequences in alternate screen mode

mmaxim:
The point of this patch is to add an option to make the mouse scroll wheel configurable to send escape sequences in alternate screen mode. This adds onto the similar arrow key option that is already present by creating two more configuration options in the Advanced pane. 

The motivation for this change is to be able to implement the technique described here:
https://www.staldal.nu/tech/2009/01/11/how-to-use-mousewheel-in-gnu-screen/

Note: I'm not an Objective-C expert by any means, so I won't be surprised if I did something here that isn't best practice.

### DIFF
--- a/sources/PTYTextView.m
+++ b/sources/PTYTextView.m
@@ -1546,7 +1546,7 @@ static const int kDragThreshold = 3;
     [super rightMouseDragged:event];
 }
 
-- (BOOL)scrollWheelShouldSendArrowForEvent:(NSEvent *)event at:(NSPoint)point {
+- (BOOL)scrollWheelShouldSendAuxForEvent:(NSEvent *)event at:(NSPoint)point upData:(NSData**)upData downData:(NSData**)downData  {
     NSRect liveRect = [self liveRect];
     if (!NSPointInRect(point, liveRect)) {
         return NO;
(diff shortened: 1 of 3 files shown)


sources/PTYTextView.m
```diff
@@ -1546,7 +1546,7 @@ - (void)rightMouseDragged:(NSEvent *)event
     [super rightMouseDragged:event];
 }
 
-- (BOOL)scrollWheelShouldSendArrowForEvent:(NSEvent *)event at:(NSPoint)point {
+- (BOOL)scrollWheelShouldSendAuxForEvent:(NSEvent *)event at:(NSPoint)point upData:(NSData**)upData downData:(NSData**)downData  {
```
gnachman: Let's separate computing the data to send from determining if data should be sent. That means two methods:

````
- (BOOL)scrollWheelShouldSendDataForEvent:(NSEvent *)event
````
(by not taking a `point` we're doing a touch of cleanup here since the point can be computed from the `event` and the caller doesn't need it for anything but passing to this method)

and
````
- (NSData *)dataToSendForScrollEvent:(NSEvent *)event
````
which calls `enclosingScrollView` and `accumulateVerticalScrollFromEvent`, and then returns the appropriate data.

File: sources/PTYTextView.m [hunk 2]
```diff
@@ -1560,40 +1560,52 @@ - (BOOL)scrollWheelShouldSendArrowForEvent:(NSEvent *)event at:(NSPoint)point {
...
+        *upData = [_dataSource.terminal.output keyArrowUp:event.modifierFlags];
+        *downData = [_dataSource.terminal.output keyArrowDown:event.modifierFlags];
+        return YES;
+    } else if (alternateMouseScrollEscapeUp.length && alternateMouseScrollEscapeDown.length) {
+        
+        NSString* e_ups = [NSString stringWithFormat:@"\e%@", alternateMouseScrollEscapeUp ];
+        NSString* e_dps = [NSString stringWithFormat:@"\e%@", alternateMouseScrollEscapeDown ];
+        
+        *upData = [ e_ups dataUsingEncoding:_dataSource.terminal.encoding ];
+        *downData = [ e_dps dataUsingEncoding:_dataSource.terminal.encoding ];
```
gnachman: Use `_delegate.textViewEncoding` instead of reaching through to the terminal.

sources/PTYTextView.m
```diff
@@ -1560,40 +1560,52 @@ - (BOOL)scrollWheelShouldSendArrowForEvent:(NSEvent *)event at:(NSPoint)point {
...
     }
-    if (!showingAlternateScreen) {
+    
+    if (alternateMouseScroll) {
+        *upData = [_dataSource.terminal.output keyArrowUp:event.modifierFlags];
+        *downData = [_dataSource.terminal.output keyArrowDown:event.modifierFlags];
+        return YES;
+    } else if (alternateMouseScrollEscapeUp.length && alternateMouseScrollEscapeDown.length) {
+        
+        NSString* e_ups = [NSString stringWithFormat:@"\e%@", alternateMouseScrollEscapeUp ];
```
gnachman: This should probably send multiple repeats of the code when you scroll fast. Also, don't prepend `\e`. Instead, use `-[NSString stringByExpandingVimSpecialCharacters]`, which converts a `\e` into a <kbd>esc</kbd>.

sources/PTYTextView.m
```diff
@@ -1560,40 +1560,52 @@ - (BOOL)scrollWheelShouldSendArrowForEvent:(NSEvent *)event at:(NSPoint)point {
...
+    if (!showingAlternateScreen) {
         return NO;
     }
-    if (!showingAlternateScreen) {
+    
+    if (alternateMouseScroll) {
+        *upData = [_dataSource.terminal.output keyArrowUp:event.modifierFlags];
+        *downData = [_dataSource.terminal.output keyArrowDown:event.modifierFlags];
+        return YES;
+    } else if (alternateMouseScrollEscapeUp.length && alternateMouseScrollEscapeDown.length) {
```
gnachman: Seems like this could cause confusion if a user intentionally sets only one to experiment with it.

sources/PTYTextView.m
```diff
@@ -1560,40 +1560,52 @@ - (BOOL)scrollWheelShouldSendArrowForEvent:(NSEvent *)event at:(NSPoint)point {
         return NO;
     }
     BOOL alternateMouseScroll = [iTermAdvancedSettingsModel alternateMouseScroll];
+    NSString* alternateMouseScrollEscapeUp = [iTermAdvancedSettingsModel alternateMouseScrollEscapeUp];
```
gnachman: (style nitpick) Put the space before the `*`

sources/iTermAdvancedSettingsModel.m
```diff
@@ -64,6 +64,8 @@ + (NSString *)name { \
 DEFINE_BOOL(allowDragOfTabIntoNewWindow, YES, @"Tabs: Allow a tab to be dragged and dropped outside any existing tab bar to create a new window.");
 
 #pragma mark Mouse
+DEFINE_STRING(alternateMouseScrollEscapeUp, @"", @"Mouse: Scroll wheel sends the specified escape sequence in alternate screen mode (up)");
```
gnachman: Add a newline and explain that the string should use vim syntax, such as `\e` for escape.

mmaxim:
@gnachman Thanks for the response! I pushed some changes which I think addresses all of your feedback. Let me know, and thanks!

gnachman:
Merged as commit 9c63f6f06b88afe8d139a0a371fb22aeaa8ac45f. Thanks!

mmaxim:
Great, thanks!

PR #284, opened 2016-05-25
set the min & max font size for badge

namtzigla:
This patch allow user to set a min & a max font size for badge font. I manly done this for users with retina display & external monitor where the current badge options are making the feature almost unusable.


gnachman:
Thanks for the PR!

If we enforce the minimum font size then we also need to truncate the text properly by putting an ellipsis at the end. I believe that passing `(NSStringDrawingTruncatesLastVisibleLine | NSStringDrawingUsesLineFragmentOrigin)` as the `options:` argument of `drawWithRect:options:attributes:` will do this. Also, this will need to be pulled into the v3.1 branch as I'm no longer adding features to 3.0 (which is on the master branch for a few more days).

PR #286, opened 2016-05-29
Adding an advanced terminal setting that lets underlining always use a static color

rsmmr:
Adding an advanced terminal setting that lets underlining always use a static color, independent of the color that the character itself is using. The underline color is currently hardcoded to red.

### DIFF
--- a/sources/ITAddressBookMgr.h
+++ b/sources/ITAddressBookMgr.h
@@ -105,6 +105,7 @@
 #define KEY_USE_CURSOR_GUIDE       @"Use Cursor Guide"
 #define KEY_CURSOR_GUIDE_COLOR     @"Cursor Guide Color"
 #define KEY_BADGE_COLOR            @"Badge Color"
+#define KEY_UNDERLINE_COLOR        @"Underline Color"
 
 // Display
 #define KEY_ROWS                   @"Rows"
--- a/sources/iTermAdvancedSettingsModel.h
+++ b/sources/iTermAdvancedSettingsModel.h
@@ -123,6 +123,7 @@
 + (BOOL)allowDragOfTabIntoNewWindow;
 + (BOOL)typingClearsSelection;
 + (BOOL)focusReportingEnabled;
++ (BOOL)staticUnderlineColor;
 
 + (BOOL)hideFromDockAndAppSwitcher;
 + (BOOL)hotkeyWindowIgnoresSpotlight;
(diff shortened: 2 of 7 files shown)


gnachman:
What's the use case for this? I need to understand more about the problem you're trying to solve

rsmmr:
I want to underline text in my editor in a consistent color,
independent of what color the characters themselves have. My immediate
use case is spell-checking: underline all misspelled words in red.
Without the patch, if I set my editor to underline the words, the line
will appear in the same color as the text itself, which is much less
noticeable. With the patch, I get a consistent red underlining.

gnachman:
This doesn’t feel quite right to me. I guess I could see having a setting in **Prefs > Colors** to set the underline color (with an off-by-default checkbox to enable/disable the feature), but an advanced pref that turns it red has very narrow applications.

rsmmr:
Sure, that would be better, I just didn't know if it warrants a place outside the advanced settings, and I didn't see way a to make the color configurable inside there. So would you consider including this if it were in Prefs/Color?

gnachman:
Yeah, I'd be ok with putting it in **Prefs>Color**. This is a more involved change, and it should be done in the `v3.1` branch. Here's what you'd need to do, and sorry about how involved it is (iTerm2 is complex!):

* Modify PreferencePanel.xib to add a checkbox and a color well in the **Basic Colors** section (like *Tab color*)
* Modify ProfileColorsPreferencesViewController.m to add IBOutlets and define the control (use tab color as a model here). Modify *updateColorControlsEnabled* to disable the color well when the checkbox is unchecked. Modify *colorWellDictionary* to include the underline color.
* Add `KEY_UNDERLINE_COLOR` and `KEY_USE_UNDERLINE_COLOR` to ITAddressBookMgr.h and add default values to iTermProfilePreferences.m
* Modify `-[iTermColorPresets colorKeys]` to add the underline color
* Modify PTYSession+Scripting.m to add a `setUnderlineColor:` and `setUseUnderlineColor:` methods (use setForegroundColor: and 
* Modify iTerm2.sdef to add "use underline color" and "underline color" properties (use foreground color as a model for underline color, and use type="boolean" for *use underline color*.
* Modify iTermColorMap.h and .m to add a new constant `kColorMapUnderlineColor` with value 9. Increment `kColorMap8bitBase` to 10.
* Modify `-[iTermColorMap profileKeyForColorMapKey:]` for the new property.
* Modify `-[iTermColorMap setColor:forKey:]` to remove the object from `_map` when the color is `nil`.
* Modify `-[PTYSession setPreferencesFromAddressBookEntry:]` to add the underline property to the keyMap variable. If underlines aren't being used the value should be `[NSNull null]`.
* Modify `iTermTextDrawingHelper` to use `[self.colorMap colorForKey:kColorMapUnderline]` if it is nonnil.
* Update the comment in `-[iTermColorPresets addColorPresetNamed:toProfile:]` to explain that the color could be nil if a custom underline color is not in use.
* Create two new tests in PTYTextViewTest.m: `-testCustomUnderline` (which ensures the underline color equals the custom color regardless of the foreground text color) and `-testUnderline` (which ensures the underline color equals the foreground text color). Use `testDimmingText` as your model, since it modifies the color map. The escape sequence to turn on underlining is `\e[4m` and you can turn it off with `\e[24m`. 
* After the tests are written, three sets of "golden" images must be created so tests will pass. Retina, non-retina, and Travis (which runs in a VM and renders anti-aliased text differently). Set the `createGolden:` argument to `YES` and run the tests. They will pass and log a message telling you where the golden files were placed. Open them in an image viewer and ensure they are correct. Then copy them into iTerm2/tests/Goldens. Then set iTerm2 to run in a non-retina configuration by right-clicking on its icon in finder, select Get Info, and turn on *Open in Low Resolution*. Re-run the tests. Copy them into the Goldens directory, but modify the filename to include `-nonretina-` (look at how the other files are named for a model). Finally, set `createGolden:` to NO and run the test suite in Travis. The tests will fail and a message will be logged in travis with a link to a zip file that has the golden images. Copy them into the Goldens directory with `-travis-` in the name (look at how other files are named for a model).
* Send a separate PR for the iterm-website github repo to update the docs. Update `source/_includes/documentation-preferences.md`, and `source/_includes/documentation-applescript.md`. There isn't a v3.1 branch yet, but you can just send it on master and I'll create a 3.1 branch when I merge this.

rsmmr:
Perfect, thanks for the detailed instructions, very helpful! I'll work through as soon as I find time.

PR #288, opened 2016-05-29
we could be leaking PTYTasks when calling "continue;" if we have dupl…

ronyfadel:
…icate fds,

in which case "[task release]" would not be called

### DIFF
--- a/sources/TaskNotifier.m
+++ b/sources/TaskNotifier.m
@@ -385,7 +385,7 @@ NSString *const kTaskNotifierDidSpin = @"kTaskNotifierDidSpin";
                     PtyTaskDebugLog(@"Duplicate fd %d", fd);
                     continue;
                 }
-                [task retain];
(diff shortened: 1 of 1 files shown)


gnachman:
Good catch. Yikes. This code needs to be cleaned up in a big way.

ronyfadel:
Yes, it was part of my broken pipe investigation.
Hopefully I'll get to help cleaning up this code one day 😄

gnachman:
If you do get around to it, there are two things to keep in mind:
1. Consider using kevent instead of poll
2. There's a bug in select. If a file descriptor is closed at the same time select is called, it never becomes readable. See the comment here (which unfortunately seems to refer to the wrong bug number 😢 ) https://github.com/gnachman/iTerm2/blob/master/sources/PTYTask.m#L445

ronyfadel:
👍

PR #289, opened 2016-06-06
add font constrains on font size for badges

namtzigla:
@gnachman, (this is the same PR as GH-284) 
I'm trying to avoid this 
screen shot 2016-06-05 at 17 21 22
and limit the font size to obtain this:
screen shot 2016-06-05 at 17 24 30

the patch does not affect the the current behavior unless the user is not going to use the


sources/iTermBadgeLabel.m
```diff
@@ -99,9 +99,20 @@ - (NSImage *)freshlyComputedImage {
          _fillColor,
          NSStringFromSize(_viewSize),
          [NSThread callStackSymbols]);
+    float maxFont = [iTermAdvancedSettingsModel badgeMaxFontSize];
```
gnachman: Why not change the implementation of `-idealFontSize` since it already has a `min` and `max`? Then this becomes a much simpler change.

namtzigla:
Hi @gnachman , I did your proposed change (thx, much better) 
but the tests are filing b/c I've merge the main repo v3.1 branch.
I think this PR is ready as it is and you can merge it.

gnachman:
Looks good. Tests are pretty much hosed on 3.1 at the moment.

PR #292, opened 2016-06-13
Adding a new color preference setting that defines a custom color for underlining.

rsmmr:
This is on top of the v3.1 branch. I believe I have followed most of the instructions in #286, except for two things:

I had to modify [self.colorMap colorForKey:kColorMapUnderline] to account for the fact that we now expect it to return nil in this case. It would normally have returned a hardcoded red. Not so nice, maybe I'm missing something?

I only added 1 one of the 3 test images, the non-retina version, as that's what I'm working on here. Not sure if/how I can create a retina version on a non-retina system? Also, I wasn't sure what exactly "run the test suite in Travis" means, so I left that part out. I would appreciate if you could add the missing images for testing when merging.

Please have a look. Feel free to send back to me if there's more to do.


sources/PTYSession.m
```diff
@@ -2635,14 +2635,19 @@ - (void)setPreferencesFromAddressBookEntry:(NSDictionary *)aePrefs
         _tmuxTitleOutOfSync = YES;
     }
 
+    BOOL use_underline = [iTermProfilePreferences boolForKey:KEY_USE_UNDERLINE_COLOR inProfile:aDict];
```
gnachman: Minor nitpick: `useUnderline` instead of `use_underline`.

sources/PTYSession.m
```diff
@@ -2635,14 +2635,19 @@ - (void)setPreferencesFromAddressBookEntry:(NSDictionary *)aePrefs
...
                               @(kColorMapLink): KEY_LINK_COLOR,
                               @(kColorMapCursor): KEY_CURSOR_COLOR,
-                              @(kColorMapCursorText): KEY_CURSOR_TEXT_COLOR };
+                              @(kColorMapCursorText): KEY_CURSOR_TEXT_COLOR,
+                              @(kColorMapUnderline): (use_underline ? KEY_UNDERLINE_COLOR : [NSNull null])
+                              };
+
     for (NSNumber *colorKey in keyMap) {
         NSString *profileKey = keyMap[colorKey];
         NSColor *theColor = [[iTermProfilePreferences objectForKey:profileKey
```
gnachman: `objectForKey:inProfile:` expects an NSString* for the first argument, and passing an `NSNull` works only by accident. Better to say

````
NSColor *theColor = nil;
if ([profileKey isKindOfClass:[NSString class]]) {
  theColor = ...
}
````

File: sources/iTermColorMap.m [hunk 3]
```diff
@@ -86,19 +87,30 @@ - (void)setMutingAmount:(double)mutingAmount {
 }
 
 - (void)setColor:(NSColor *)theColor forKey:(iTermColorMapKey)theKey {
-    if (!theColor || theColor == _map[@(theKey)] || theKey >= kColorMap24bitBase) {
+    if (theKey >= kColorMap24bitBase)
+        return;
+
+    if ( !theColor ) {
```
gnachman: Style nitpick: No spaces inside the parens. Should be `if (!theColor) {`

sources/iTermColorMap.m
```diff
@@ -86,19 +87,30 @@ - (void)setMutingAmount:(double)mutingAmount {
...
+        [_map removeObjectForKey:@(theKey)];
```
gnachman: The change to remove the object from the map when the color is nil will change behavior when called from `-[PTYSession screenSetColor:forKey:]`. This would happen when we get an escape sequence to change the color map with an improperly specified argument. It used to have no effect, but now it would leave the colormap with a missing entry. Guard against it by changing `-[PTYSession screenSetColor:forKey:]` to do nothing if its color argument is `nil`.

sources/iTermColorMap.m
```diff
@@ -86,19 +87,30 @@ - (void)setMutingAmount:(double)mutingAmount {
...
         _backgroundBlue = [theColor blueComponent];
     }
-    theColor = [theColor colorUsingColorSpaceName:NSCalibratedRGBColorSpace];
-    _map[@(theKey)] = theColor;
+
     if (theKey == kColorMapBackground) {
         _backgroundBrightness = [theColor perceivedBrightness];
     }
+
+    theColor = [theColor colorUsingColorSpaceName:NSCalibratedRGBColorSpace];
```
gnachman: Did you move this intentionally? I think backgroundBrightness should be calculated from the RGB colorspace since that is what gets displayed.

sources/iTermColorMap.m
```diff
@@ -113,7 +125,7 @@ - (NSColor *)colorForKey:(iTermColorMapKey)theKey {
         return [NSColor colorWith8BitRed:red green:green blue:blue];
     } else {
         NSColor *result = _map[@(theKey)];
-        if (!result) {
+        if (!result && theKey != kColorMapUnderline) {
```
gnachman: It's ok to just return result here. The code to return red was to catch bugs, but it hasn't been needed any more.

gnachman:
Good work! Very close to submittable; I just wrote a few small comments. I'll take care of adding the golden image for Travis.

sources/iTermColorMap.m
rsmmr: I did, but I didn't think right. Will revert.

sources/iTermColorMap.m
rsmmr: If I return result directly, I get a crash right after startup that seems unrelated to the changes in this branch:

```
        0   CoreFoundation                      0x00007fff971374f2 __exceptionPreprocess + 178
        1   libobjc.A.dylib                     0x00007fff9666773c objc_exception_throw + 48
        2   CoreFoundation                      0x00007fff9719e4bd +[NSException raise:format:] + 205
        3   AppKit                              0x00007fff8e7603e0 -[NSColorSpaceColor initWithColorSpace:components:count:] + 120
        4   AppKit                              0x00007fff8e760295 +[NSColor colorWithColorSpace:components:count:] + 29
        5   iTerm2                              0x000000010ecfcb69 -[iTermColorMap processedBackgroundColorForBackgroundColor:] + 953
        6   iTerm2                              0x000000010eba766b -[PTYTextView defaultBackgroundColor] + 203
        7   iTerm2                              0x000000010eba78e2 -[PTYTextView updateMarkedTextAttributes] + 66
        8   iTerm2                              0x000000010eba636b -[PTYTextView initWithFrame:colorMap:] + 395
        9   iTerm2                              0x000000010eb6d6c7 -[PTYSession setScreenSize:parent:] + 1143
        10  iTerm2                              0x000000010ece5068 -[PseudoTerminal setupSession:title:withSize:] + 2216
        11  iTerm2                              0x000000010eceb9cf -[PseudoTerminal addSessionInNewTab:] + 191
        12  iTerm2                              0x000000010ecea964 -[PseudoTerminal createTabWithProfile:withCommand:] + 1236
        13  iTerm2                              0x000000010ec4bedd -[iTermController launchBookmark:inTerminal:withURL:isHotkey:makeKey:canActivate:command:block:] + 1229
        14  iTerm2                              0x000000010ec4b4a7 -[iTermController launchBookmark:inTerminal:] + 119
        15  iTerm2                              0x000000010ec45f66 -[iTermController newWindow:possiblyTmux:] + 182
        16  iTerm2                              0x000000010ed05e08 -[iTermApplicationDelegate newWindow:] + 104
        17  iTerm2                              0x000000010ed01ec9 -[iTermApplicationDelegate performStartupActivities] + 873
```

Any idea?

rsmmr:
Updated with your feedback, except for the change to *result*, see above.

sources/iTermColorMap.m
gnachman: Commit f0c064d should fix the crash you saw.

gnachman:
Merged. Thank you!

rsmmr:
Thanks for merging and the feedback, and more generally for all your great work on iTerm2!

josepheiba:
Hi, is there anyway I can change iTerm2 Colorspace from sRGB to P3 or to my current display's colorspace?

gnachman:
@josepheiba The beta uses p3

PR #297, opened 2016-07-27
Tweaks to imgcat script: Add flag to print image name

coljac:
I use this script constantly. Since it updates itself with Iterm2, and since these tiny tweaks are useful, perhaps this is worth considering.

I invoke imgcat a lot on big groups of images, often via xargs. The following tweaks are handy:

Added a "-p" flag which prints the image filename before displaying the image. Useful when invoked on a large number of images at once and you want to see which is which.

Added a tweak where print_image returns if no filename supplied; this was a problem when invoked via xargs, it was displaying the broken image icon to the screen between each image in a list of images using zsh on Linux and OS X.


gnachman:
Please update the show_help function to mention the `-p` flag.

I think the filenames should go beneath the images, like a caption.

coljac:
Good suggestions, have changed accordingly.

tests/imgcat
```diff
@@ -25,6 +25,9 @@ function print_st() {
 #   inline: 0 or 1
 #   base64contents: Base64-encoded contents
 function print_image() {
+    if [[ -z "$1" ]]; then
+      exit 0
```
gnachman: I think this is wrong. The reason you got a broken image is the call on line 105. Adding this would break `cat image.png | imgcat`. A better fix would be to not accept input from stdin if there is a filename command-line argument.

coljac:
Yes. Now it:

- Prints the filename on -p below the image
- Skips reading from stdin if a filename arg is passed

This is more useful, for instance, in the case of `ls *png | head -10 | xargs imgcat -p`.

tests/imgcat
```diff
@@ -74,14 +77,18 @@ while [ $# -gt 0 ]; do
...
     -*)
         error "Unknown option flag: $1"
         show_help
         exit 1
       ;;
     *)
         if [ -r "$1" ] ; then
-            print_image "$1" 1 "$(base64 < "$1")"
+            has_stdin=f
+            print_image "$1" 1 "$(base64 < "$1")" "$print_filename"
```
gnachman: Please update the comments before print_image to describe its new argument.

gnachman: Also, on line 103, pass `""` to `print_image` since it now takes four arguments.

coljac:
Done, thank you.

gnachman:
Thank you!

PR #299, opened 2016-08-07
Fixed the yen button error that happened if the current keyboard layout doesn't produce "¥".

golddranks:
iTerm2 mimics MacOS's Terminal and produces "\" instead "¥", when the yen button of a Japanese keyboard is pressed. This is because of a legacy quirk of Japanese keyboard layouts, that use ¥ instead of \\.

However, the current implementation naively checks only if the pressed key is the yen key (the key code). However, as the user can change and edit keyboard layouts, the yen keys doesn't necessarily produce "¥".

With this commit, it checks whether the character production of the key is "¥". It still needs to check the keycode in order to prevent any other mappings that produce "¥" from turning "\". (Another solution would be to check if the `charactersIgnoringModifiers` equals to "¥" and leave out the key code check... I'm not aware which one has less corner cases.)

### DIFF
--- a/sources/NSEvent+iTerm.m
+++ b/sources/NSEvent+iTerm.m
@@ -58,26 +58,20 @@
 - (NSEvent *)eventByChangingYenToBackslash {
     // NSEvent: type=KeyDown loc=(0,477) time=103943.2 flags=0x80120 win=0x7fd5786432b0 winNum=3667 ctxt=0x0 chars="\" unmodchars="¥" repeat=0 keyCode=93
     
-    if (self.keyCode == kVK_JIS_Yen) {
-        NSString *characters = @"\\";
-        if (self.modifierFlags & NSShiftKeyMask) {
-            characters = @"|";
-        } else {
-            characters = @"\\";
-        }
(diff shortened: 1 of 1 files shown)


golddranks:
Btw. I tested this both with a "normal" Japanese keyboard layout, and my custom layout. Now the Japanese layout produces "\" on the yen key, but the custom layout are unaffected, which is the same behaviour that MacOS's Terminal does.

golddranks:
Sorry for minor history editing. I ended up using `charactersIgnoringModifiers` and squashed the changes.

The semantics of `charactersIgnoringModifiers` seems to be spot-on for this case:

1) If the user uses an original Japanese layout, the yen button produces "\" as it should. Shift is the only modifier that `charactersIgnoringModifiers` doesn't ignore, so when typing "|" it ignores the if branch and produces "|".

2) If the user has a custom layout, pressing a button that produces "¥" unmodified, it produces now "\", like the button would be the yen button. However, if a combination like alt+y produces "¥", it's likely that the user really meant to produce "¥" so it produces "¥".

gnachman:
Thank you!

PR #300, opened 2016-08-08
Fixed yet another cornercase with the "¥".

golddranks:
I came up with another corner case. Thanks, #showerthoughts.

It now checks that the production that would be emitted originally by the keypress would be "¥" also – this for the (rare) case that alt + ¥ would produce something special that the user intended to produce.

In the normal case (the standard Japanese keyboard on Mac) alt + ¥ is mapped to emit "\", so in the majority of cases "\" will be emitted either way.


gnachman:
I believe this would only affect shift+¥, right?

PR #301, opened 2016-08-24
Match OS X tab bar styling more closely

csmulhern:
This PR makes the tab bar styling used in iTerm more closely match the tab bar styling used across OS X. To more closely match Safari et al., the following changes have been made:

- Show a divider line visible between a selected tab and the status bar.
- Increase the tab bar control height to 26 points.
- Center the content within the tab cells.
- Use a font size of 11 points.

### DIFF
--- a/ThirdParty/PSMTabBarControl/source/PSMDarkHighContrastTabStyle.m
+++ b/ThirdParty/PSMTabBarControl/source/PSMDarkHighContrastTabStyle.m
@@ -31,8 +31,4 @@
   return [NSColor colorWithCalibratedWhite:value alpha:1.00];
 }
 
-- (CGFloat)fontSize {
-  return 12.0;
-}
-
 @end
--- a/ThirdParty/PSMTabBarControl/source/PSMDarkHighContrastTabStyle.m
+++ b/ThirdParty/PSMTabBarControl/source/PSMDarkHighContrastTabStyle.m
@@ -31,4 +31,8 @@
   return [NSColor colorWithCalibratedWhite:value alpha:1.00];
 }
 
+- (CGFloat)fontSize {
+    return 12.0;
+}
+
 @end
--- a/ThirdParty/PSMTabBarControl/source/PSMYosemiteTabStyle.m
+++ b/ThirdParty/PSMTabBarControl/source/PSMYosemiteTabStyle.m
@@ -647,12 +647,12 @@
             labelRect.size.width -= ([self objectCounterRectForTabCell:cell].size.width + kPSMTabBarCellPadding);
         }
 
-        NSSize boundingRect = [attributedString boundingRectWithSize:labelRect.size options:0].size;
-        labelRect.origin.y = floor((cellFrame.size.height - boundingRect.height) / 2.0);
-        labelRect.size.height = boundingRect.height;
+        NSSize boundingSize = [attributedString boundingRectWithSize:labelRect.size options:0].size;
+        labelRect.origin.y = cellFrame.origin.y + floor((cellFrame.size.height - boundingSize.height) / 2.0);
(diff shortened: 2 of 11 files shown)


ThirdParty/PSMTabBarControl/source/PSMDarkHighContrastTabStyle.m
```diff
@@ -31,8 +31,4 @@ - (NSColor *)backgroundColorSelected:(BOOL)selected highlightAmount:(CGFloat)hig
   return [NSColor colorWithCalibratedWhite:value alpha:1.00];
 }
 
-- (CGFloat)fontSize {
-  return 12.0;
```
gnachman: The text is intentionally oversize in the high-contrast themes to aid legibility at the expense of beauty.

gnachman:
The tab label's baseline should equal the shortcut digit's baseline. It's off by half a point with this change.

baseline

ThirdParty/PSMTabBarControl/source/PSMYosemiteTabStyle.m
```diff
@@ -410,11 +409,7 @@ - (CGFloat)fontSize {
 #pragma mark - Drawing
 
 - (NSColor *)topLineColorSelected:(BOOL)selected {
-    if (selected) {
-        return [_tabBar.window backgroundColor];
-    } else {
-        return [NSColor colorWithSRGBRed:182/255.0 green:179/255.0 blue:182/255.0 alpha:1];
-    }
+    return [NSColor colorWithSRGBRed:182/255.0 green:179/255.0 blue:182/255.0 alpha:1];
```
gnachman: The OS's tab bar uses a lighter color for the top line of the selected cell. Looks off to me using the same color all the way across.

screen shot 2016-08-24 at 10 35 27 pm

sources/iTermRootTerminalView.m
```diff
@@ -16,7 +16,7 @@
 #import "iTermToolbeltView.h"
 #import "PTYTabView.h"
 
-const CGFloat kHorizontalTabBarHeight = 22;
+const CGFloat kHorizontalTabBarHeight = 24;
```
gnachman: This is a somewhat disruptive change because it will affect peoples' window arrangements, by making windows taller or causing them to lose a row in some cases. It is nice to get closer to the system tab bar's look, but it's going to cause some pain. I just want to think about this some more.

csmulhern:
Thanks for the feedback.

1. The removal of 12pt font for the high contrast styling has been reverted.
2. I've fixed the baseline alignment issue; there were problems with both the centering of the tab name, and tab number.
3. I've also fixed the coloration of the cell borders not matching the native styling.
4. I've added a configuration option (off by default) that allows the tab bar cells to stretch to fill the available area.

In regards to the tab bar height change being disruptive, that's understandable. I'd be a bit hesitant to suggest making it an advanced option (I imagine it just going there to never get exposure), but maybe that's worth consideration. Although switching this from being a compile time constant could be a little hairy.

I'm also interested in adding a + tab button to the tab bar in the native style. Before I put any more effort in, Is this something I should continue pursuing?

gnachman:
Sorry for the delay in getting to this.

Thank you for fixing the dumpster fire that was PSMTabBarControl's update method.

gnachman:
Merged with some additional style and refactoring. Thanks for taking this on. The code is quite a bit nicer now and I think it looks good. FYI, this is in master which is currently only in the nightly build as it's on track to be 3.1 which is not yet feature complete.

csmulhern:
Cheers.

How do you feel about me adding a "+" button to the tab bar?

gnachman:
I think it's a good idea. If the overflow menu is visible, it should go away to avoid too much clutter.

PR #302, opened 2016-09-23
Fix curly quotes in tip text.

marnen:
Some apostrophes were curled the wrong way (see http://practicaltypography.com/straight-and-curly-quotes.html). This pull request fixes that.


gnachman:
How embarrassing. Thank you. (The font I use in XCode shows both opening and closing single quotes identically, so I have an excuse!)

PR #303, opened 2016-09-26
Add "Raw coprocess command" to semantic history.

grundprinzip:
screen shot 2016-09-26 at 1 13 06 pm
This patch adds the possibility to run a script / program
as a deferred script. This is necessary to have any interaction
back with iTerm.

One use case is for example a small script that will write text
back to iTerm to interact with the remote system while being
logged in via SSH.

This is related to: https://gitlab.com/gnachman/iterm2/issues/5197


grundprinzip:
I'm absolutely unsure how to deal with the XIB file. I manually edited only the one line needed to add the menu item, but open saving the project the other modifications were added. What's the SOP here?

sources/iTermSemanticHistoryController.m
```diff
@@ -287,9 +290,11 @@ - (BOOL)openPath:(NSString *)path
...
-        DLog(@"No file exists at %@, not running semantic history", path);
-        return NO;
+    if (isRawCoProcess) {
+        DLog(@"Launch coproress with script %@", script);
+        assert(delegate_);
+        [delegate_ semanticHistoryLaunchCoprocessWithCommand:script];
+        return YES;
     }
 
     if ([prefs_[kSemanticHistoryActionKey] isEqualToString:kSemanticHistoryCommandAction]) {
```
gnachman: Moving the test for `fileExistsAtPath:isDirectory:` below this changes the **Command** to run even when the file doesn't exist, which seems like an unnecessary change.

sources/iTermSemanticHistoryController.m
```diff
@@ -311,6 +314,13 @@ - (BOOL)openPath:(NSString *)path
         return YES;
     }
 
+    if ([prefs_[kSemanticHistoryActionKey] isEqualToString:kSemanticHistoryCoprocessAction]) {
```
gnachman: Why should coprocess actions not run on directories?

sources/iTermSemanticHistoryPrefsController.m
```diff
@@ -192,7 +193,8 @@ + (NSDictionary *)tagToActionMap {
               @3: kSemanticHistoryEditorAction,
               @4: kSemanticHistoryCommandAction,
               @5: kSemanticHistoryRawCommandAction,
-              @6: kSemanticHistoryCoprocessAction };
+              @7: kSemanticHistoryRawCoprocessAction,
```
gnachman: Please put the new item last so the numbers are in order.

sources/iTermSemanticHistoryPrefsController.m
```diff
@@ -262,6 +264,12 @@ - (IBAction)actionChanged:(id)sender
             hideCaveat = NO;
             hideText = NO;
             break;
+        case 7:
+            [[text_ cell] setPlaceholderString:@"Enter command"];
+            [caveat_ setStringValue:@"Coprocess runs when you activate Semantic History on any filename (even if it's not a avlid filename). Use \\1 for filename, \\2 for line number, \\3 for text before click, \\4 for text after click, \\5 for pwd."];
```
gnachman: Spelling of "valid"

gnachman:
Don't worry about the XIB file. They're horrible this way. It looks OK.

PR #304, opened 2016-09-27
Fix imgcat writing -n in output

remram44:
It seems that 'echo -n' is unreliable. printf never outputs a newline and is safer.

Fixes GL-5201


gnachman:
Handled in commit db4ed5d

PR #305, opened 2016-10-03
Pre-Sierra Compatibility

bergman:
I needed to make these changes to compile on OS X 10.10. Not sure if it's worth dropping support for earlier OS X versions to make that type change to `NSWindowStyleMask`.

### DIFF
--- a/sources/PopupWindow.m
+++ b/sources/PopupWindow.m
@@ -15,7 +15,7 @@
 }
 
 - (instancetype)initWithContentRect:(NSRect)contentRect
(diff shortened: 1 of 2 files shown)


gnachman:
Development is only supported on the latest XCode version. Apple makes it very challenging not to use the latest XCode, unfortunately (e.g., interface builder crapping on your files for no reason, XCode crashing when it sees something it doesn't like in a project file).

bergman:
Fair enough... I thought it would break in runtime as well but if that's not the case I am fine with closing.

PR #307, opened 2016-10-13
Windows >> Bring All To Front 

pauloba:
Option "Bring All To Front" doesn't bring the focus to iTerm. 
Not sure what this option was built for.
It would be wonderfull to have a linux-like option 'always on top', for the iTerm in order to not loose the focus of the terminal while you navigate through other documents; and always have the iTerm on top of the other programs.


gnachman:
This pull request has more commits than I think were intended.

Bring all to front is a standard item in the Window menu, and it's similarly useless in all apps as far as I can tell. If some of the windows are obscured by other apps, it raises them above.

I'm going to close this PR because the feature is working as intended.

PR #308, opened 2016-10-31
Fix incorrect text selection

patkujawa-wf:
I didn't see any Issues page, just PRs (derp, because the readme was at the bottom of the file listing; I'll use https://iterm2.com/bugs/), so here's one that *might* fix the issue I just encountered today after updating to the nightly and suddenly copying and text selection aren't working (grabbing characters not under cursor). Seems exacerbated by zooming in and out, possibly.

Haven't tested it, since I've never built locally, but just here for discussion and FYI.

This reverts commit de3f4276cb5c316dc140c3880dbc1d53d91696ed.


njarvis:
I opened https://gitlab.com/gnachman/iterm2/issues/5295

aes512:
+1 - text selection is all f**ked up now. Please regress to the old behavior!

gnachman:
@aes512 This problem was only around for a day or so back in November. Are you on that version?

PR #309, opened 2016-11-08
add keybindings for swap pane with left/right/above/below

jcs:
This adds 4 new menu options to the Key Mappings action list to be able to define keys to swap a split pane with the pane to the left/right/above/below, similar to the existing ones for just selecting that pane.

This makes it possible to swap panes with "show per-pane title bar with split panes" disabled, and makes it much faster to swap panes without the mouse.


gnachman:
Thank you!

PR #310, opened 2016-11-10
Added setting to disable mouse wheel in mouse reporting mode

theodm:
Hello,

this pull-request adds a setting, which allows to disable/enable sending mouse wheel events to the running application in "mouse reporting mode". 

My simple use case is avoiding application lag when accidentally swiping on the touchpad (MacBook). It happens in applications, which do a lot of processing while scrolling ([radare2](https://github.com/radare/radare2) in my case). 

I'm beginner and this is my first pull request, please don't be too hard with me :)


gnachman:
Looks good, thank you! I made a few minor changes but this was a good PR.

PR #312, opened 2016-11-22
Update bundle identifier for TextMate

jacob-carlborg:
Since TextMate has been released as a release candidate it has changed
the bundle identifier and removed the ".preview" suffix [1].

Not sure if we should keep TextMate 1 or not, but I kept it for now.

[1] https://github.com/textmate/textmate/commit/204c735079d01a5bcc43923756827a26a3fbf107


jacob-carlborg:
@gnachman ping.

gnachman:
Sorry for the delay and thanks for the PR!

jacob-carlborg:
Thanks, no worries 😃.

gnachman:
Ugh, launch services treats bundle identifiers case-insentively, which means there is no meaningful difference between TextMate and TextMate 2. I had to revert this and I'm renaming "TextMate 2" to "TextMate Preview". Probably nobody still uses the old TextMate, anyway.

jacob-carlborg:
Ah, ok. As long as the identifier works with the latest version of TextMate I'm happy.

PR #314, opened 2016-12-03
Move travis to `osx_image: xcode8.1`.

phs:
This gives us macOS 10.12.1, which contains `NSTouchBar`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 before_install: ./ci/before_install.sh
 bundler_args: --without documentation --without development --deployment --jobs=3 --retry=3
-osx_image: xcode8
+osx_image: xcode8.1
 language: objective-c
 script: ./ci/script.sh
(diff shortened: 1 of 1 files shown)


phs:
Worth a shot.

gnachman:
Seems like a good idea, was there a problem with moving to 8.1?

phs:
Yes.  I explored the problem over in #315.  In short the large majority of `PTYTextViewTest` tests fail due to a difference in font rendering, at least within the test suite.  If those could be fixed, then travis could move to 8.1 and be able to compile the touch bar support.

Notably when I just use the compiled binary directly, the fonts look the same as compared to a binary pulled down from homebrew.

phs:
Note you can see travis failing those tests directly [in this PR's build](https://travis-ci.org/gnachman/iTerm2/builds/181028020)

PR #315, opened 2016-12-04
Log the `backingScaleFactor`, show log in travis.

phs:
This PR is not meant to merge, only to show how the xcode 8.0 environment (which I otherwise don't have access to) behaves, in comparison to the 8.1 environment I have.

The reason we can't move to 8.1 (which contains the touch bar API that iTerm2 now requires to compile) is large number of "golden" image test failures in `PTYTextViewTest`.

Despite having separate fixtures for retina displays (like me), non-retina displays and a separate one for travis, both travis and local environments fail the same way on xcode 8.1.  Pulling up a few failing images from `/tmp/failed-*.png` shows that the "actual" images seem to match the non-retina images (with the thicker letters) instead of the expected retina ones.

`PTYTextView` decides that its window's screen is "retina" or not based on the `backingScaleFactor` attribute, updated in `viewDidChangeBackingProperties`.  If this value is greater than one, it should be a retina device.

After adding some logging, in my local 8.1 dev environment I found that the reported `backingScaleFactor` is actually zero.  Since it is supposed to be a ratio of scales, that doesn't make much sense.  Let's see what the 8.0 environment in travis says it is.

In order to run the offending tests, this branch begins at the last commit prior to the touch bar support, before it adds the logging.

### DIFF
--- a/ci/script.sh
+++ b/ci/script.sh
@@ -2,10 +2,18 @@
 
 set -euo pipefail
 
-xcrun xcodebuild build test \
+xcrun xcodebuild build \
   NSUnbufferedIO=YES \
   -workspace iTerm2.xcworkspace \
   -scheme iTerm2 \
   -sdk macosx \
   CODE_SIGNING_REQUIRED=NO CODE_SIGN_IDENTITY= \
     | xcpretty -c -f `xcpretty-travis-formatter`
+
+xcrun xcodebuild test \
+  -only-testing:iTerm2XCTests/PTYTextViewTest/testBasicDraw \
+  NSUnbufferedIO=YES \
+  -workspace iTerm2.xcworkspace \
+  -scheme iTerm2 \
+  -sdk macosx \
+  CODE_SIGNING_REQUIRED=NO CODE_SIGN_IDENTITY=
--- a/sources/PTYSession.m
+++ b/sources/PTYSession.m
@@ -6074,7 +6074,6 @@ ITERM_WEAKLY_REFERENCEABLE
     }
     iTermTextExtractor *textExtractor = [[[iTermTextExtractor alloc] initWithDataSource:_screen] autorelease];
     NSString *word = [textExtractor fastWordAt:VT100GridCoordMake(_screen.cursorX - 1, _screen.cursorY + _screen.numberOfScrollbackLines - 1)];
-    [[_delegate realParentWindow] currentSessionWordAtCursorDidBecome:word];
 }
 
 - (void)textViewBackgroundColorDidChange {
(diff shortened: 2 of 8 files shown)


phs:
Looks like the travis 8.0 environment consistently reports a `backingScaleFactor` of zero in the tests as well.  Travis is not expected to be a retina environment; it is possible this behavior has been around for some time.  Perhaps it is expected?  A retina 8.0 environment would be an interesting acid test.

phs:
It seem [we're not alone][0].

[0]: http://stackoverflow.com/q/36399309/580412

phs:
From https://mail.gnome.org/archives/commits-list/2014-December/msg05296.html

> Sending backingScaleFactor to a NULL NSWindow will silently give the
    value 0 for the scale factor, causing insidious divide-by-zero bugs down
    the line.

phs:
Not sure why this closed itself, reopening soon as github stops fighting me.

phs:
Since reverting the commits since the compile error became more painful this round, here's what I did to automate.  First tag `last-good` to something safe:

```
$ git log -1 --pretty=oneline last-good
3f885c5dcc391ca302d043dd51d530c1c22713b9 Do not raise hotkey windows when switching spaces.
```

Then we can make a loop:

```
$ for sha in $(git log --format=%h last-good..master); do git revert --no-edit $sha; done
```

After that `cherry-pick` the diagnostic commit I need to log state.

phs:
`[self window]` is indeed "false-ish" most of the time in both travis and my local 8.1.  It appears we don't have a window at least some of the time `viewDidChangeBackingProperties` is called.

phs:
`backingScaleFactor` being zero may be a red herring.  When I hard-wire it to `2.0` locally (in `viewDidChangeBackingProperties`), the test still fails, producing the thick letters.

phs:
Ok, I can't tell what's going on.  Cleaning this up for now.

gnachman:
Thanks for looking into this. I've been meaning to deal with this since forever, and this is a useful start.

phs:
The only real concrete take-away I found is the golden image failures are mostly due to the font stroke being too thick with the 8.1 sdk.  There doesn't appear to be a change in logic or control flow that would reflect that though.  It simply appears to be a difference in either the underlying fonts or API.

gnachman:
Unfortunately the font stroke seems to change with every OS and Xcode version, which makes the golden tests less useful than I'd like. Not sure what to do about it besides building some more infastructure to make updating them easier.

phs:
Fair enough; that's exactly the kind of contextual knowledge I don't have in this area.  Manual verification of the new expectations seems painful, but after that's done I don't see anything stopping you from moving travis to 8.1.

PR #283, opened 2016-05-18
Increase MARGIN and VMARGIN by 10

jpruetting:
This change increases the margins around the text to remove the visual tension that is currently present.

Adequate whitespace is a common design best practice:
- https://medium.com/@erikdkennedy/7-rules-for-creating-gorgeous-ui-part-1-559d4e805cda#b7c4
- https://designschool.canva.com/blog/white-space-design/
- http://blog.teamtreehouse.com/white-space-in-web-design-what-it-is-and-why-you-should-use-it

This is a common change request that is often completed by compiling your own version of iTerm2:
- http://superuser.com/questions/810582/iterm-osx-terminal-change-text-margin-from-edge-of-window
- https://www.reddit.com/r/unixporn/comments/43e9ch/iterm2_with_adjustable_margins/
- https://www.reddit.com/r/unixporn/comments/37csyp/osx_heres_my_custom_iterm2_without_borders_and/
- https://github.com/jaredculp/iterm2-borderless-padding
- https://gist.github.com/unrolled/c5e04dd7138377fed18e
- https://gitlab.com/gnachman/iterm2/issues/950

Having to modify the source and build a custom version of iTerm2 is a pain. It also makes taking new updates difficult.

PR #316, opened 2016-12-04
Move MARGIN/VMARGIN to iTermAdvancedSettingsModel

jcs:
A mostly mechanical change of `MARGIN` -> `[iTermAdvancedSettingsModel terminalMargin]` and `VMARGIN` -> `[iTermAdvancedSettingsModel terminalVMargin]` to allow these to be adjusted by the user (or to easily change the defaults from 5 and 2, as per #283).

It would probably be nice to move these to per-profile settings so different profiles can have different margins, but these values are used in so many places that make it hard to reach the current profile settings.


sources/iTermAdvancedSettingsModel.m
```diff
@@ -134,6 +134,8 @@ + (NSString *)name { \
 DEFINE_BOOL(indicateBellsInDockBadgeLabel, YES, @"General: Indicate the number of bells rung while the app is inactive in the dock icon’s badge label");
 DEFINE_STRING(downloadsDirectory, @"", @"General: Downloads folder.\nIf set, downloaded files go to this location instead of the user’s $HOME/Downloads folder.");
 DEFINE_FLOAT(pointSizeOfTimeStamp, 10, @"General: Point size for timestamps");
+DEFINE_INT(terminalMargin, 5, @"General: Terminal Margin\nHow much space to leave between the left and right edges of the terminal.");
```
gnachman: Add this text to both terminalMargin and terminalVMargin at the end of the description, please:

You must restart iTerm2 after modifying this property.

gnachman:
This is scary to change because it's completely metastasized. The effort to move it to a profile preference would be immense, as you note, and I think making it configurable this way is a pretty big win for a lot of people. Thanks for taking this on.

jcs:
I've made those requested changes.

ysupr:
Hi i really interested with this https://github.com/jaredculp/iterm2-borderless-padding/issues/19#issuecomment-273011908 

does it mean this PR we will have margin in top and left like that screenshot?

gnachman:
@ysupr Yes, try the nightly build and see how you like it.

JesseLeite:
@gnachman Is this still only available in the nightly build?

luco:
This is very, very nice.

yifanchen:
I have nightly build 3.1 - 20170303 now, in advanced setting, I changed the margin to 25 and 25, restarted the iterm, nothing changed, any ideas?

gnachman:
@JesseLeite You can get 3.1.beta.1 from the downloads page now. The march to 3.1 is underway!

@yifanchen If you save prefs to a folder (in prefs > general), maybe it's not set to save changes back to the folder?

JesseLeite:
@gnachman Just downloaded the beta and found the experimental setting. Love it! ❤️

yifanchen:
@gnachman Hmm, it didn't work either. Isn't the padding/margin setting not in the place where the following image showing?

image

JesseLeite:
@yifanchen I found it here:

yifanchen:
@JesseLeite That works, thanks.

jpruetting:
This is soooooo nice. Super happy this finally got added!

aquartier:
It's look like this functionality is missing in 3.0.15 beta.

gnachman:
It's not missing so much as it is a new feature of 3.1.

ErraticFox:
I cannot find it anywhere and @JesseLeite photos broken...

JesseLeite:
@ErraticFox fixed photo :)

piotryordanov:
Is it possible to have the margins be different for every profile? or is this a global setting for item?

gnachman:
It is a global setting. I would like to make it per-profile one day.

KLVTZ:
For those wondering how to get rid of the title bar:

Open Preferences, click the Profiles tab, select your profile, click Window and select No Title Bar from the Style dropdown menu.

kizerkizer:
How do you change this setting?

gnachman:
It's easier to change it in **Prefs > Advanced > Height of top and bottom margins in terminal panes** and **Width of left and right margins in terminal panes**.

cbartondock:
I still have a right margin in vim after setting all the margins to zero in iTerm2 preferences:
screen shot 2018-06-26 at 1 00 43 pm

Any thoughts?

KLVTZ:
@cbartondock Maybe your .vimrc has some column configuration set?

gnachman:
@cbartondock Your window's width is not an exact multiple of the width of a cell. If it's a fullscreen window, that is unavoidable since text size is limited to integer points size. If it's not a fullscreen window resize its width a little and it should snap to the grid and fix it.

whistlegraph:
Was this setting removed?

gnachman:
It is now in regular prefs, not advanced prefs.

Screen Shot 2021-04-05 at 10 08 14 PM

mikepietruszka:
Are there any recent plans to make this per profile?

gnachman:
> Are there any recent plans to make this per profile?

I hadn't planned to do that. What's the use case?

mblarsen:
> > Are there any recent plans to make this per profile?
> 
> I hadn't planned to do that. What's the use case?

E.g. vim that handles its own margins. Switch to vim profile (borderless) but default profile has some margins.

gnachman:
I'd rather not make it per-profile because I don't think it's useful enough to justify the added complexity.

aasutossh:
image
not too sure about this but the margin still wont disappear completely, is this because of the font's dimension?

gnachman:
There can always be a margin on the right or bottom if the window's size isn't an exact multiple of the cell size

aasutossh:
I understand. Thank you.

naavid-morshed:
I still have border at the right and bottom when I open emacs in iterm2. the left and bottom boder combines at bottom right and creates a small box shape. Setting both margins in appearances did not totally remove the border.

gnachman:
@naavid-morshed box shape? Share a screen shot

PR #318, opened 2016-12-15
[Experimental, WIP] Merge title-bar and tabs bar

prashnts:
**Experimental and explosive!**

This is with reference to [issue #5404](https://gitlab.com/gnachman/iterm2/issues/5404). I've implemented a prototype with minimal changes required to existing tab code.

OSX 10.11 onwards [1], it's supposedly easy to implement this with minimal changes, as this fab gif here demonstrates:

iterm mov

### Caveats:
- Window transparency/blur isn't visible -- it's a solid window. I couldn't find a fix for this as my Objective C/Cocoa is so, so.
- If you drag the tab bar (which detaches it from main window normally), it crashes iTerm2, possibly because the way it is handled?

My feeling is that the above happens because the `contentView` controls the blur/transparency. 

I'm willing to fix the above if this looks like a good approach, though could use some help.

-Prashant

[1] http://stackoverflow.com/a/34562491/2648954

--
Also works well with light background without any changes whatsoever.
snip 2016-12-15 at 16 31 52


gnachman:
That is surprisingly little code. Congrats on finding a clever solution :)

Unfortunately the issue with `NSFullSizeContentViewWindowMask` breaking transparency is hard to fix. It opts you in to layer backing, and layer-backed windows cannot be transparent (https://openradar.appspot.com/radar?id=4968871706492928).

I have a feeling that in order to support transparency you'll need to change the window's style mask to remove the title bar and close, minimize, and maximize buttons. Then we'll have to re-implement the buttons like Chrome does.

prashnts:
Thanks for having a look!

I suspected that too. Agree that implementing the buttons is the best option here -- I shall work on it later today and tomorrow and update this PR.

tal:
I think it's also feasible to get around the limitations of the various tab positioning by making this style work only when the configuration is in the proper setup. I don't think it even makes sense to implement for the others.

gnachman:
If tabs are on the bottom or side then there should be a regular title bar, but maybe we should force it to be black by default.

prashnts:
> If tabs are on the bottom or side then there should be a regular title bar

Yup, I think that's the best solution.

@gnachman merged with your latest commit ed6754b191301f317ea9a1f661f21415c7966d58 and the transparency works (blur doesn't, though), but it still looks rad!

snip 2016-12-27 at 14 51 50


I noticed that the background color of "content-area" could be moved to the window's background color instead. You can see the thin background color from root layer below tab bar:

snip_2016-12-27_at_14_54_18

Also, the window corners are much much cleaner with your recent commit. (Bottom one is without `NSFullSizeContentViewWindowMask`, and current alpha channel build; top one is `HEAD` I built locally).

snip 2016-12-27 at 15 04 01


I think it might be worth exploring a fix for blur (since you already fixed the transparency issue!).

Thanks!

gnachman:
The blur thing is going to be a problem. It looks like the private API we use for blur just doesn't work when there is a layer. I spent a little time trying to figure out how NSVisualEffectView works but all I learned is that it's very complex :).

My only idea—and it's a bad one—is to create a clear, content-less borderless window with blur and keep it behind the terminal window.

gnachman:
After thinking it over, this window style should either use NSVisualEffectView to blur this window style. The UI in preferences will be hard to get right, though.

prashnts:
@gnachman Implemented the `NSVisualEffectView` and sure enough, the blur works! (It still can't be merged and please treat it as a proof-of-concept :)

Problem with this is, however, that the blur radius isn't controllable explicitly -- the public API enumerates several blur "stages"[1]. Though I feel maybe changing alpha level of `visualEffectView` might work?

snip 2016-12-30 at 15 16 51

[1] https://developer.apple.com/reference/appkit/nsvisualeffectmaterial?language=objc

gnachman:
Great job! Did you see the second answer here: http://stackoverflow.com/questions/25529500/how-to-set-the-blurradius-of-uiblureffectstyle-light

I bet this will work on Mac OS by changing UI to NS.

gnachman:
My previous suggestion was way off but this looks promising:
http://stackoverflow.com/questions/37231536/adjust-nsvisualeffectview-blur-radius-and-transparency

gnachman:
I dug some more and it looks like adjusting the alpha value might be our only option (and even that is discouraged 😢 )

sources/iTermRootTerminalView.m
```diff
@@ -46,6 +47,14 @@ - (instancetype)initWithFrame:(NSRect)frameRect
     if (self) {
         _delegate = delegate;
 
+        self.visualEffectView = [[NSVisualEffectView alloc] initWithFrame:frameRect];
```
gnachman: Add an `autorelease` here. Yay for manual reference counting :)

prashnts: I'd added manual release in `dealloc` method, because I assumed the `subview` will be a reference?

```objectivec
(void)dealloc {
   [_visualEffectView release];
```

prashnts:
Yeah, I looked into this too, and I feel we probably should not use `NSFullSizeContentViewWindowMask` after all, as it is breaking more things than what I'd consider acceptable!

I found that the window behaved weirdly as soon as I enable layers using `wantsLayer`.

I propose that I could instead implement your initial recommendation of drawing the traffic light buttons in a `BorderlessWindowMask`, in another pull request? This would obviate the need for all this new hack!

PS: Can we have rounded corners on `BorderlessWindow` without manually drawing a Bezier curve and not enabling layers? With layers it's simply done by setting `layer.cornerRadius = 4` and clipping mask.

gnachman:
What weird behavior did you see when layers were enabled? I did try enabling layers on the scrollview at one time and that completely broke it, but I haven't tried putting a layer on the higher-up views.

If we draw the traffic light buttons ourselves we'll need to add a view on top of all the other views that draws the corners. Either as a bezier path or as images. The nice thing about not using layers is that it's easy to punch a hole through the window like this. I'm not aware of anything like cornerRadius in a world without layers.

The downsides of using a borderless window are:
1. It's deprecated. But it'll keep working for a long time, so I guess I don't care that much.
2. No shadow. But we've lived with this for a long time.

prashnts:
I uploaded a screencast with the glitches I saw with layers enabled. In addition to that, I did notice drawing performance was noticeably lagging. 

https://www.youtube.com/watch?v=Pf1gMmvj4II

Thanks for the explanation about border situation, btw! I will update the pr as I figure that out.

gnachman:
Somewhere I have a branch that tries to make things work with layers, but I gave up on it because as you noted it's slower on some machines (and faster on others). Some day I will get rid of NSScrollView (as Terminal.app did a few versions back) and then it'll be worth taking another shot at layers. AFAICT it's just fundamentally incompatible with layers.

danqing:
Just wanted to say that this is really dope..

jgavris:
Wanted to 👍  and say I'd love to see this land! That is one mighty fine terminal.

FledgeXu:
It‘s so beautiful.Just wanna to it.

AtomicCoding:
Just wanted to ask, is development still going on this?

gnachman:
It's in the queue, but it's a long queue. I'll revisit it after I'm done with performance improvements for 3.2.

amirmasoudabdol:
I'm not sure if this is possible or not, or what stage development is at now, but I I was wondering if it's possible to make Tab's color adaptive to the selected color scheme. Sublime Text does something similar to this. It's in fact very subtle but it blends tabs to the editing environment nicely.

gnachman:
Once macOS 10.14 is out this will be practical. They've fixed a bunch of issues with layers that were blocking this UI.

prashnts:
That's a brilliant news, I'm glad it's still open, Thanks @gnachman !

August is going to be free from school for me -- I'll get my hands on 10.14 beta soon, and see if the old patch works.

gnachman:
This is in the nightly build now, but requires macOS 10.14, Set your window style in **Prefs>Profiles>Window** to **Compact** and select the minimal theme in **Prefs > Appearance**.

PR #319, opened 2016-12-23
Borderless rounded

DrabWeb:
screen shot 2016-12-23 at 5 17 00 am

Changes borderless terminal windows to be rounded instead of square, 10.10+ only

### DIFF
--- a/sources/PseudoTerminal.m
+++ b/sources/PseudoTerminal.m
@@ -394,7 +394,13 @@ static NSRect iTermRectCenteredVerticallyWithinRect(NSRect frameToCenter, NSRect
         case WINDOW_TYPE_LEFT_PARTIAL:
         case WINDOW_TYPE_RIGHT_PARTIAL:
         case WINDOW_TYPE_NO_TITLE_BAR:
-            return mask | NSBorderlessWindowMask | NSResizableWindowMask;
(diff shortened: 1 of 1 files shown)


gnachman:
I believe this would break transparency (**Prefs > Profiles > Window > Transparency**) because it causes the window's content view to be layer-backed.

DrabWeb:
Well both borderless rounded implementations require layer-backing, so I guess there's not much I can do

gnachman:
Electron does something interesting we might be able to copy:
https://github.com/electron/electron/blob/026711e7502a90670b628d82f1e632c98b375ced/atom/browser/native_window_mac.mm#L1404

gnachman:
Hey, I made some progress! The `noTitleBar` branch has a proof of concept showing rounded corners and shadows on a no-title-bar window. There are a bunch of bugs to work out where various things are or are not transparent. The good news is that layers aren't the kiss of death as I had thought—it just requires making sure that every view (and the window) under PTYTextView is transparent.

NickClark:
Just compiled the `noTitleBar` branch, nice to see it working. So far the most irritating bug is the window "stoplight" buttons showing over the top. Must be one of those bugs you mentioned 😉

PR #320, opened 2016-12-23
Fix opening files with line numbers using TextMate

jacob-carlborg:
Since TextMate changed the bundle identifier and removed the
".preview" suffix iTerm2 has been updated [1]. But the original bundle
identifier was never updated. This mostly worked fine because in the
operating system bundle identifiers are case insensitive.

What did not work is opening a files using TextMate with line numbers.
The problem is that iTerm2 checks if an application supports opening a
file with line numbers, this check looks at the bundle identifiers,
but it uses a case sensitive comparison. Therefore TextMate was
treaded as an editor that did not support opening files with line
numbers.

I would guess that in general all comparisons of bundle identifiers within iTerm2 should be case insensitive to better match the operating system.

[1] f31454263c02b1612573ca6525c032e544a82a2a


gnachman:
Thanks for the cogent description.

PR #322, opened 2017-02-16
Tabs vs. accessibility, as discussed via email

petermaurer:
The tab bar and its cells now behave more standard-like in terms of accessibility. As a reminder, this uses macOS 10.10+ APIs.


gnachman:
Looks good. Thanks!

PR #324, opened 2017-03-21
Support cmd-click file:line:col and VSCode.

marcelocantos:
This PR adds support for column-number components in cmd-clicked paths (e.g.: `a/b/file.txt:25:11`) and for VSCode.

I was a bit lazy and didn't add any tests to confirm :col works, since the existing tests don't seem to test :line anyway (only that it doesn't fail). I did nonetheless add :col to the new VSCode tests.


gnachman:
This looks good. Would you mind renaming `colNumber` to `columnNumber`, please? Abbreviations are not generally used in this code.

marcelocantos:
No problems. I've pushed a version with `columnNumber` everywhere.

gnachman:
Thanks!

PR #326, opened 2017-04-03
Reduce the amount of amount of python2

grimreaper:
Also stop recommending that people run pip as root and instead just use
virtualenv.

### DIFF
--- a/api/examples/python/README.md
+++ b/api/examples/python/README.md
@@ -1,3 +1,5 @@
+# iTerm2 API
+
 This is an example of how to use the iTerm2 API in Python. It sets up a
 websocket connection, sends an RPC to subscribe to notifications about changes
 to the current username, hostname, or working directory. It prints incoming
@@ -7,16 +9,24 @@ notifications.
 
 First, install the dependencies.
 
+### Using virtualenvwrapper
+
+```
+mkvirtualenv iterm2api
+pip install websocket-client
+pip install protobuf
+```
+
+### Using system python
+
+It is also possible to install the dependencies as system libraries.
+
 ```
 sudo /usr/bin/python -m pip install websocket-client
 sudo /usr/bin/python -m pip install protobuf
-
 ```
 
 Then `cd api/examples/python` and then `./iterm2.py` to run the program.
-
-This silly way of running pip is needed if you have more than one version of
-python installed.
 
 Now you can run `iTerm2/api/examples/python`.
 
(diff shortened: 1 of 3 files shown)


api/examples/python/README.md
```diff
@@ -8,16 +8,13 @@ notifications.
 First, install the dependencies.
 
 ```
-sudo /usr/bin/python -m pip install websocket-client
-sudo /usr/bin/python -m pip install protobuf
-
+mkvirtualenv iterm2api
```
gnachman: What's mkvirtualenv? I don't have it by default on my system. I agree sudo is awful, but I don't want to add even more dependencies on third party packages.

grimreaper: mkvirtualenv is a fairly common wrapper around python's virtualenv tools: 
https://virtualenvwrapper.readthedocs.io/en/latest/command_ref.html

You can get the same effect with
```
mkdir -p ~/.virtualenv
cd ~/.virtualenv
virtualenv xxx
source ./xxx/bin/activate
```

gnachman: OK to add a section describing how to do it using virtualenv as an alternative to `sudo python` but I don't want to insist on more deps than are necessary. Dependencies are the root of all frustration.

grimreaper: IMHO modifying system python using sudo is more likely to lead to eventual frustration than using isolated environments, but I'll rewrite this section to be an alternative

Cabalist:
Hey there,
I support this pull request but it would be worthwhile to remove the space after `print` to clarify that it is a function (PEP8 \*cough\*). 

Also as the is example code I don't think creating a `main()` function really adds anything.   It is totally fine to put it under `if __name__...`

My 2¢ as a random python dev...

grimreaper:
> it would be worthwhile to remove the space after print to clarify that it is a function (PEP8 *cough*).

done.

> It is totally fine to put it under if __name__...
Personally I always use a main() even if writing simple scripts.

mennis:
virtenv is an additional dependency both in terms of code and conceptually.  The --user option to the pip install command will install a package for the current user, rather than for all users.

gnachman:
@mennis Does that mean you wouldn't need to use sudo? How should the doc be changed to use `--user`?

mennis:
Yes.   `pip install --user  websocket-client protobuf ` will work without sudo. In my case it will install the packages in ${HOME}/Library/Python/2.7/lib/python/site-packages/ .

grimreaper:
Its generally best practice to not install project dependencies either as sudo or as user. They should be in a virtualenv. The user of --user meant for binaries that you expect to be on the PATH.

PR #328, opened 2017-05-24
fix activity indicator in the tabbar

343max:
The activity spinner in iTerm would look pretty different then in most other apps: a segment would start very light and become darker and darker instead of starting dark and fading out. This diff reverses the fading effect.

Before:
spinnerbefore

After:
spinnerafter


gnachman:
How come I never noticed that? Thanks for the fix.

PR #331, opened 2017-09-19
Resolved issue with Cursor Type overlapping Blinking Cursor

taylorzane:
This PR resolves issue [#6046](https://gitlab.com/gnachman/iterm2/issues/6046)

I've manually tested, and confirmed, that this issue has been resolve with this fix.


gnachman:
Thanks! I decided to go ahead and move it right instead of putting it on top, and bumping a few other checkboxes around to make things line up nice.

PR #332, opened 2017-09-19
Turn off DNS Lookups by default

JosephRedfern:
Many people are requesting that this feature be removed entirely, but disabling by default is a good start (and possibly a reasonable compromise?) (see https://gitlab.com/gnachman/iterm2/issues/6050).


tuxxy:
+1 for this.

If a user can unwittingly leak their passwords through this feature, it should be disabled by default and possibly have a warning before allowing it to be enabled.

gnachman:
I fixed this before seeing the PR, but thank you.

PR #334, opened 2017-10-01
Add "check for updates" to menubar icon

andystanton:
If you don't restart iTerm for a long time you can miss updates. This is typically solved by using the option in the "iTerm" menu. However if you have the "Exclude from Dock and ⌘-Tab Application Switcher" option enabled, there is no easy way to check for updates without disabling this option or restarting iTerm.

This pull request adds an option to the menu bar icon to trigger the "Check for Updates" functionality.

screen shot 2017-10-01 at 23 31 26

### DIFF
--- a/sources/iTermApplicationDelegate.m
+++ b/sources/iTermApplicationDelegate.m
@@ -166,6 +166,7 @@ static const NSTimeInterval kOneMonth = 30 * 24 * 60 * 60;
     IBOutlet NSMenuItem *showFullScreenTabs;
     IBOutlet NSMenuItem *useTransparency;
     IBOutlet NSMenuItem *maximizePane;
+    IBOutlet SUUpdater * suUpdater;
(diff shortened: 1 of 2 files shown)


sources/iTermApplicationDelegate.m
```diff
@@ -1478,6 +1484,10 @@ - (IBAction)checkForIncompatibleSoftware:(id)sender {
     }
 }
 
+- (IBAction)checkForUpdatesFromMenu:(id)sender {
+    [[[SUUpdater alloc] init] checkForUpdates:(sender)];
```
gnachman: Is this kosher? We will have two SUUpdater objects. I think it's better to use the one that already exists, which is an object in MainMenu.xib. Add an IBOutlet to iTermApplicationDelegate so you can call it from here.

andystanton:
@gnachman Thanks for the kind suggestion. I've made the change and squashed the commits.

gnachman:
Thanks!

PR #335, opened 2017-10-04
Fix position of 'Paste' button

KarboniteKream:
The `Paste` button in the special paste window was not properly aligned.

Before:
Before

After:
After


bunam:
hi
it would be nice if this panel would be configurable to show up every time you attempt to paste a multiline item

gnachman:
Thanks, @KarboniteKream!

@bunam Please file a feature request at https://iterm2.com/bugs

PR #336, opened 2017-10-04
Fix warnings introduced in Xcode 9

KarboniteKream:


gnachman:
Thanks for the fix. Would you mind making the changes in `iTerm2/ColorPicker/ColorPicker` instead? The files you changed are in the build output of that project, which is only in the repo to keep build time fast.

KarboniteKream:
Done!

gnachman:
Thank you!

gnachman:
I've built the framework and committed it in ccddc395f26494efb995cd87469a8736ebff0c69

PR #337, opened 2017-10-18
Allow ~ in HTTP URL paths

brenns10:
This allows URLs such as http://sharedserver.com/~homedir/index.html to be parsed and opened correctly.

I've updated my own profile preferences, but I feel like people would benefit from having this as the default.


gnachman:
This seems like a worthwhile change but please be aware that it won't affect existing profiles, which already have a copy of the regex.

brenns10:
Makes sense - I already adjusted my regex anyway. Thanks for merging :)

PR #338, opened 2017-10-25
Auto select the account in the password manager after auth.

yilei:
Every time I re-open my compiled iTerm2, it requires re-authentication (Touch ID). Is that because I disabled code-signing? So I wasn't able to test the case when iTerm2 is first opened, but doesn't require authentication.


gnachman:
Thanks for the PR, Yilei! Hope all is well. Merged :)

yilei:
You too, George! I do miss Objective C sometime 🤣

gnachman:
>  I do miss Objective C sometime 🤣

What are you doing now? Say it's not C++

yilei:
https://github.com/abseil/abseil-py :)

gnachman:
Cool project! I look forward to using it :)

PR #341, opened 2017-11-05
Fix regression focus after switching between apps

trufae:


trufae:
Removing this code fixes the problem of rotating terminals when switching between apps. it's pretty anoying. I dont know why is this code here, so I'm just sending a PR for discussion. Feel free to fix that properly in your repo

gnachman:
The goal of this code was to keep the window under the cursor active, since you might position the mouse and then cmd-tab to iTerm2. I guess when you cmd-tab out, move the mouse, and cmd-tab back you want to return to the window that used to be active, even if the mouse is no longer over it?

trufae:
No need to move the cursor to reproduce this issue. Just open multiple iterm windows and ciclce between apps, everytime that iterm gets focus a different window is raised

trufae:
Yep sounds reasonable but i should test the behaviour once fixed

Thanks

gnachman:
Please test tomorrow's nightly build. The change is in commit 2b4e2f59

PR #344, opened 2017-11-25
Support OSC 10 and OSC 11 (dynamic colors)

stevecheckoway:
The foreground color can be set or reported using OSC 10. The background
color can be set or reported using OSC 11. There are 10 dynamic colors
defined, this only implements the first two and does not implement the
resetting (OSC 110 and OSC 111).

This PR has the side effect of changing how colors are parsed in OSC 4 in two ways.

1. Like XTerm, it supports `rgb:R/G/B`, `rgb:RR/GG/BB`, `rgb:RRR/GGG/BBB`, and `rgb:RRRR/GGGG/BBBB` (or any mixing of component lengths). This should be a strict superset of the current functionality.
2. Colors are not translated into the calibrated color space. Doing so meant that setting a color to `rgb:ff/ff/ff` did not get reported as ff/ff/ff:

   ``` bash
   $ echo -ne '\033]4;105;rgb:ff/ff/ff\033\\'
   $ echo -ne '\033]4;105;?\033\\'
   ^[]4;105;rgb:fe/fe/fe^G
   ```

   With this patch, the colors round-trip correctly:

   ``` bash
   $ echo -ne '\033]4;105;rgb:ff/ff/ff\033\\'
   $ echo -ne '\033]4;105;?\033\\'
   ^[]4;105;rgb:ff/ff/ff^G
   ```

I suspect that the second change is likely to make this PR unpalatable given that the rest of the codebase uses the calibrated color space. Perhaps the report functions should convert back to sRGB.

### DIFF
--- a/sources/VT100Terminal.m
+++ b/sources/VT100Terminal.m
@@ -1916,6 +1916,11 @@ static const int kMaxScreenRows = 4096;
             [self executeXtermSetRgb:token];
             break;
 
+        case XTERMCC_SET_FG:
+        case XTERMCC_SET_BG:
+            [self executeXtermSetDynamicColor:token];
+            break;
+
         case DCS_TMUX_CODE_WRAP:
             // This is a no-op and it shouldn't happen.
             break;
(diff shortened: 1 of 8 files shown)


sources/VT100Terminal.m
```diff
@@ -1972,46 +2011,79 @@ - (void)executeXtermSetRgb:(VT100Token *)token {
...
+    MOUSE_FG,
+    MOUSE_BG,
+    TEKTRONIX_FG,
+    TEKTRONIX_BG,
+    HIGHLIGHT_BG,
+    TEKTRONGIX_CURSOR,
+    HIGHLIGHT_FG,
+};
+
+static int XtermDynamicColorForToken(VT100Token *token) {
```
gnachman: Move this up with the other static functions

sources/VT100Terminal.m
```diff
@@ -1972,46 +2011,79 @@ - (void)executeXtermSetRgb:(VT100Token *)token {
...
+        if ([colorSpec isEqualToString:@"?"]) {
+            switch (dynamic_color) {
+                case TEXT_FG:
+                    theColor = [delegate_ terminalForegroundColor];
+                    break;
+                case TEXT_BG:
+                    theColor = [delegate_ terminalBackgroundColor];
+                    break;
+            }
+            [delegate_ terminalSendReport:[self.output reportColor:theColor forDynamicColor:dynamic_color]];
```
gnachman: Condition this on `if ([delegate_ terminalShouldSendReport])`

stevecheckoway:
It looks like some of this code got incorporated into iTerm in the (yikes) two years since I submitted this PR. In particular support for OSC 10 and 11.

Colors don't appear to round trip correctly just as they did not before. It's been long enough that I forget exactly what the "calibrated color space" is and what sRGB is. My interest was somewhat more narrow: I wanted to know if the current background is light or dark.

Therefore, I'll close this PR and leave some sort of generic color space to someone else who is more interested in this.

Sorry for the inordinately long delay in responding and thank you for implementing the OSC 10/11 in the meantime.

PR #345, opened 2017-11-26
Fix build failure on case-sensitive file systems

gentoo-root:
On a case-sensitive file system iTerm2 fails to build because of missing Profile.h. I believe this is a typo, and actually ProfileModel.h is meant (and iTerm2 builds successfully then).

The build succeeds on case-insensitive file systems, because ProfileModel.h is pulled via ITAddressBookMgr.h, and Profile.h resolves to the system header profile.h (note the small letter p).


gnachman:
Thanks for the fix.

PR #346, opened 2017-12-04
Make tmux detection in imgcat more reliable

timvisee:
Detecting whether `imgcat` is used inside `tmux` by checking the `$TERM` variable is not reliable.
The current version checks whether a `screen*` terminal is running.
I'm running `xterm-256color` however, therefore `tmux` is not detected by `imgcat`.

This PR checks for `tmux` using the `$TMUX` variable instead, a variable `tmux` creates when attaching to a session.

### DIFF
--- a/tests/imgcat
+++ b/tests/imgcat
@@ -4,7 +4,7 @@
 # <sequence> ST, and for all ESCs in <sequence> to be replaced with ESC ESC. It
 # only accepts ESC backslash for ST.
 function print_osc() {
-    if [[ $TERM == screen* ]] ; then
+    if [[ -n $TMUX ]] ; then
         printf "\033Ptmux;\033\033]"
     else
         printf "\033]"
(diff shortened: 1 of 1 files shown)


gnachman:
The problem with this is that `$TERM` gets forwarded through ssh but `$TMUX` does not. There is no perfect solution, and tmux seems popular enough that this is a reasonable tradeoff versus not detecting any mutiplexer over ssh.

timvisee:
Interesting, and now I understand this is used. I think it might be good to document this in code, or somewhere else.

Now, what about `xterm-...` detection though. Should this terminal type be added as additional hint for `tmux` terminals, as I'm using? Or would that possibly break `xterm` terminals outside `tmux`, as the current solution probably does too?

Thanks for your response.

gnachman:
I've updated the comment in imgcat to explain the reasoning. If you're using `xterm-256color` with tmux then you're going to have a bad time. You should use `screen-256color` instead.

PR #347, opened 2017-12-07
Alter auto comp select

wuyihao14:
Does anyone find `Auto Command Completion` annoying as me ?

I add/altered 3 features:

On `Auto Command Completion` displaying,

1. Press `Enter` to send '\n' to psudo-tty
2. Press `^Enter` to select row
3. Press `^C` to close `Auto Command Completion` window

### DIFF
--- a/sources/CommandHistoryPopup.m
+++ b/sources/CommandHistoryPopup.m
@@ -99,4 +99,15 @@
     }
 }
 
+- (void)insertNewline:(id)sender
+{
+    [self insertText:@"\n"];
+    [self closePopupWindow];
+}
+
+- (void)insertLineBreak:(id)sender
+{
+    [self rowSelected:self];
+}
+
 @end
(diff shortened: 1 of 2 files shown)


wuyihao14:
https://superuser.com/questions/1089955/disable-autocomplete-hints-on-iterm2

Comments below may suffer from similar problems. I do too.

Maybe configuration is an option, in case anyone in favor of existing behavior ?

wuyihao14:
I noticed, it might be better if if replace `^Enter` to `Tab`.

gnachman:
I think this is a good improvement, but the change from Enter to Control-Enter is going to be hard for users to discover. What about using right-arrow instead?

Also, please put this behind a configuration flag. I prefer to test out such changes on beta users to see how it's received. Add a setting to the "Experimental Features" section of `iTermAdvancedSettingsModel.[hm]`, enabled by default for beta users. Look for `supportREPCode`, which already does this. Make the new behavior conditional on its being enabled.

PR #349, opened 2018-02-07
Add advanced setting preventing directional focus from wrapping

brads55:
Currently the directional focus hotkeys will wrap, focusing a view on the opposite side of the window.

Personally I find this very confusing and often end up in situations whereby I have no idea where focus has gone, (mostly because I've come from a tiling window manager that behaved slightly differently).

This commit adds an advanced option (probably not important enough to be a real option) that prevents such wrapping.


gnachman:
Thanks!

PR #352, opened 2018-02-12
Add F21-F24 to Touch Bar

msvticket:
I need to use F21-F24 (for tn5250) so I added support for that. This also makes it possible to assign labels (using SetKeyLabel escape sequence) for these keys.


sources/PseudoTerminal+TouchBar.m
```diff
@@ -522,6 +522,10 @@ - (void)sendFunctionKeyToCurrentSession:(NSInteger)number {
         kVK_F18,
         kVK_F19,
         kVK_F20,
+        NSF21FunctionKey,
```
gnachman: These values aren't actually virtual key codes so they don't belong in this list. There is no virtual keycode for F21-F24 that Apple defines. I suggest putting a value like 0xfff for these with a comment saying "There is no virtual keycode for these function keys, but it doesn't matter for our purposes."

The only place where a bogus virtual keycode would cause problems is in the call to `routeEventToShortcutInputView:` on line 544. Modify that function to return YES but do nothing when the virtual keycode is 0xffff.

Finally, don't use 0xffff as a magic number—define a constant for it in `iTermApplication.h` like:

```
// Used for keys you can press on the touch bar that have no equivalent on physical keyboards that Apple recognizes
extern unsigned short iTermBogusVirtualKeyCode;
````

and in `iTermApplication.m` like

```
unsigned short iTermBogusVirtualKeyCode = 0xffff;
```

msvticket:
The branch has been updated according to the comment.

I guess it's pretty clear I don't have much clue  about this.

gnachman:
Merge din commit 8a42e50b907a9333f2421326853b5dc0a268c4ce. Thanks!

msvticket:
I apparently didn't test the pull request properly after doing the requested change. In turns out that after changing the code according to the review comment it stopped working. That is F21-F24 are visible in the Touch Bar, but they don't work. I just tried reverting that change locally and rebuild and then the keys indeed do work. So how do we proceed?

gnachman:
I fixed it. Try tomorrow's nightly build and let me know if there are any issues.

msvticket:
Yeah, now it works

PR #353, opened 2018-03-01
Fixes singleton, refactor power monitoring mode (#6550)

mspasov:
- Fixes singleton implementation of the power manager instance (bug #6550)
- Changes the way power status is monitored to a more optimal callback
when power source has been changed to a limited power source
(IOPSCreateLimitedPowerNotification)

ps. excuse my objc, it's a little bit rusty. :)

### DIFF
--- a/sources/iTermPowerManager.m
+++ b/sources/iTermPowerManager.m
@@ -11,29 +11,48 @@
 
 NSString *const iTermPowerManagerStateDidChange = @"iTermPowerManagerStateDidChange";
 
-@implementation iTermPowerManager
+@implementation iTermPowerManager {
+    CFRunLoopRef _runLoop;
+    CFRunLoopSourceRef _runLoopSource;
+}
+
+static BOOL iTermPowerManagerIsConnectedToPower(void) {
+    CFStringRef source = IOPSGetProvidingPowerSourceType(NULL);
+    return [@kIOPMACPowerKey isEqualToString:(__bridge NSString *)(source)];
+}
+
+static void ITPMpowerSourceChanged(void *context) {
+    iTermPowerManager* pm = (__bridge iTermPowerManager *)(context);
+    [pm setConnected:iTermPowerManagerIsConnectedToPower()];
+}
 
 + (instancetype)sharedInstance {
-    return [[self alloc] initPrivate];
+    static iTermPowerManager *sharedInstance = nil;
+    static dispatch_once_t onceToken;
+    dispatch_once(&onceToken, ^{
+        sharedInstance = [[self alloc] initPrivate];
+    });
+    return sharedInstance;
 }
 
 - (instancetype)initPrivate {
     self = [super init];
     if (self) {
-        [self updateCharging];
-        [NSTimer it_scheduledTimerWithTimeInterval:5 repeats:YES block:^(NSTimer * _Nonnull timer) {
-            BOOL before = _connectedToPower;
-            [self updateCharging];
-            if (_connectedToPower != before) {
-                [[NSNotificationCenter defaultCenter] postNotificationName:iTermPowerManagerStateDidChange object:nil];
-            }
-        }];
+        _runLoop = CFRunLoopGetMain();
+        _runLoopSource = IOPSCreateLimitedPowerNotification(ITPMpowerSourceChanged,(__bridge void *)(self));
+        _connectedToPower = iTermPowerManagerIsConnectedToPower();
+        if (_runLoop && _runLoopSource){
+            CFRunLoopAddSource(_runLoop,_runLoopSource,kCFRunLoopDefaultMode);
+        }
     }
     return self;
 }
 
-- (void)updateCharging {
-    _connectedToPower = (IOPSGetTimeRemainingEstimate() == kIOPSTimeRemainingUnlimited);
+- (void)setConnected:(BOOL)connected {
+    if (_connectedToPower != connected) {
+        _connectedToPower = connected;
+        [[NSNotificationCenter defaultCenter] postNotificationName:iTermPowerManagerStateDidChange object:nil];
+    }
 }
 
 @end


sources/iTermPowerManager.h
```diff
@@ -12,7 +12,8 @@ extern NSString *const iTermPowerManagerStateDidChange;
 
 @interface iTermPowerManager : NSObject
 
-@property (nonatomic, readonly) BOOL connectedToPower;
+@property (nonatomic, assign) BOOL connectedToPower;
```
gnachman: Keep this readonly. In the .m file add:

````
@interface iTermPowerManager()
@property (nonatomic, readwrite) BOOL connectedToPower;
@end
````

This prevents external callers from changing the value.

File: sources/iTermPowerManager.m [hunk 2]
```diff
@@ -11,29 +11,52 @@
...
+    
+    CFStringRef source = IOPSGetProvidingPowerSourceType(NULL);
+    if (source){
+        pm.connectedToPower = [@"AC Power" isEqualToString:(__bridge NSString *)(source)];
+    }
+}
 
 + (instancetype)sharedInstance {
-    return [[self alloc] initPrivate];
+    static iTermPowerManager *sharedInstance = nil;
```
gnachman: 😳Thank you for fixing this. I don't know what I was thinking.

sources/iTermPowerManager.m
```diff
@@ -11,29 +11,52 @@
...
+        }
     }
     return self;
 }
 
-- (void)updateCharging {
-    _connectedToPower = (IOPSGetTimeRemainingEstimate() == kIOPSTimeRemainingUnlimited);
+- (void)setConnectedToPower:(BOOL)connectedToPower {
+    if (_connectedToPower != connectedToPower) {
+        _connectedToPower = connectedToPower;
```
gnachman: Is the notification delivered not on the main queue, necessitating the dispatch_async? If so the assignment to `_connectedToPower` is unsafe and needs to be protected by a lock of some sort. I'd use `@synchronized(self)` to keep it simple as this will happen very rarely and performance isn't a concern.

sources/iTermPowerManager.m
```diff
@@ -11,29 +11,52 @@
...
-            BOOL before = _connectedToPower;
-            [self updateCharging];
-            if (_connectedToPower != before) {
-                [[NSNotificationCenter defaultCenter] postNotificationName:iTermPowerManagerStateDidChange object:nil];
-            }
-        }];
+        _runLoop = CFRunLoopGetMain();
+        _runLoopSource = IOPSCreateLimitedPowerNotification(ITPMpowerSourceChanged,(__bridge void *)(self));
+        if (_runLoop && _runLoopSource){
+            CFRunLoopAddSource(_runLoop,_runLoopSource,kCFRunLoopDefaultMode);
```
gnachman: 😃 So much better!

sources/iTermPowerManager.m
```diff
@@ -11,29 +11,52 @@
...
-        [self updateCharging];
-        [NSTimer it_scheduledTimerWithTimeInterval:5 repeats:YES block:^(NSTimer * _Nonnull timer) {
-            BOOL before = _connectedToPower;
-            [self updateCharging];
-            if (_connectedToPower != before) {
-                [[NSNotificationCenter defaultCenter] postNotificationName:iTermPowerManagerStateDidChange object:nil];
-            }
-        }];
+        _runLoop = CFRunLoopGetMain();
+        _runLoopSource = IOPSCreateLimitedPowerNotification(ITPMpowerSourceChanged,(__bridge void *)(self));
```
gnachman: I think this might  not call post a notification when you plug in. The description of the function says:

````
Returns a CFRunLoopSourceRef that notifies the caller when power source
changes from an unlimited power source (like attached to wall, car, or airplane power), to a limited
power source (like a battery or UPS).
````

Comment (mspasov) [reply sources/iTermPowerManager.m]: Yep, tested it. called every time back and forth.

File: sources/iTermPowerManager.m (hunk 3 above)
Comment (mspasov) [reply sources/iTermPowerManager.m]: The crash happens when the notification is sent from the dispatch_once block.
```    0x103456138 <+80>:  ja     0x103456107               ; <+31>
    0x10345613a <+82>:  leaq   0x1fdc7(%rip), %rcx       ; "BUG IN CLIENT OF LIBDISPATCH: trying to lock recursively"
    0x103456141 <+89>:  movq   %rcx, 0x4ae08(%rip)       ; gCRAnnotations + 8
```
The call is from the main thread, my fix is just a workaround. I'll try to find what goes wrong.

mspasov:
Should be better now. I don't have a workaround for the initial call of the block.

ps. I've forgotten how Xcode is supposed to run.. FAST. And I've forgotten how much i hate CoreFoundation :)

sources/iTermPowerManager.m
```diff
@@ -11,29 +11,51 @@
...
+        if (_runLoop && _runLoopSource){
+            CFRunLoopAddSource(_runLoop,_runLoopSource,kCFRunLoopDefaultMode);
+            dispatch_async(dispatch_get_main_queue(), ^{
```
gnachman: I think there need to be two changes here:

1. Eliminate dispatch_async so that the value of connectedToPower will be correct as soon as the object is initialized.
2. Don't call ITPMpowerSourceChanged because it causes a notification to be posted, and that doesn't make sense to do at this time. Callers aren't prepared to handle that notification at this time.

I think a better design would be to do:
```
static BOOL iTermPowerManagerIsConnectedToPower(void) {
    CFStringRef source = IOPSGetProvidingPowerSourceType(NULL);
    return [@"AC Power" isEqualToString:(__bridge NSString *)(source)]];
}

static void ITPMpowerSourceChanged(void *context) {
    iTermPowerManager* pm = (__bridge iTermPowerManager *)(context);
   [pm setConnected:iTermPowerManagerIsConnectedToPower()];
}
```

and in initPrivate, do:
```
_connectedToPower = iTermPowerManagerIsConnectedToPower();
```

mspasov: Thanks, looks perfect, I'll make the changes and squash everything in one commit.

PR #354, opened 2018-03-19
Native Notifications

tessarin:
After the introduction of the built-in notification system in Mountain Lion (10.8) Growl has been falling into disuse. The 2.0 framework supports the native `NSUserNotificationCenter` and acts as a bridge, which iTerm is still using to send notifications. Considering iTerm2 was made for 10.10+, Growl is in fact acting as a forward service only in most of the cases (except for users who still prefer to install and use Growl).

The main problem with using Growl as a forwarding service is the lack of customization. The framework hasn't been updated for a long time, meaning iTerm is stuck with the frozen notification format implemented by Growl. Forwarded notifications don't have a sound, for example. Even though `NSUserNotificationCenter` supports it, Growl does not use it. Nor implements support for action buttons which could be used in the future.

This PR might be the first step into deprecating Growl without actually deprecating it. A new advanced setting was added (enabled by default to beta users only) to skip Growl and send the notification directly. The class `iTermGrowlDelegate` was renamed to `iTermNotificationController` and so were its methods, to remove Growl references. Clicking a native notification will also bring the corresponding view to front.

### DIFF
--- a/iTerm2.xcodeproj/project.pbxproj
+++ b/iTerm2.xcodeproj/project.pbxproj
@@ -595,7 +595,7 @@
 		1D5FDD601208E8F000C46BA3 /* PSMTabStyle.h in Headers */ = {isa = PBXBuildFile; fileRef = DD02573909CB9398008F320C /* PSMTabStyle.h */; };
 		1D5FDD621208E8F000C46BA3 /* PSMTabDragAssistant.h in Headers */ = {isa = PBXBuildFile; fileRef = F6E708B70A9D0EA400D0C4EF /* PSMTabDragAssistant.h */; };
 		1D5FDD651208E8F000C46BA3 /* PSMTabDragWindow.h in Headers */ = {isa = PBXBuildFile; fileRef = F62D15F00AA64B2F0075A287 /* PSMTabDragWindow.h */; };
-		1D5FDD661208E8F000C46BA3 /* iTermGrowlDelegate.h in Headers */ = {isa = PBXBuildFile; fileRef = F69E78910AB7AC85001EC0FF /* iTermGrowlDelegate.h */; };
+		1D5FDD661208E8F000C46BA3 /* iTermNotificationController.h in Headers */ = {isa = PBXBuildFile; fileRef = F69E78910AB7AC85001EC0FF /* iTermNotificationController.h */; };
 		1D5FDD681208E8F000C46BA3 /* CarbonHelpers.h in Headers */ = {isa = PBXBuildFile; fileRef = F6441F620E748404000EC682 /* CarbonHelpers.h */; };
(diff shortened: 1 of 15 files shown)


asmeurer:
A regression with the native notifications: clicking the notification no longer opens the relevant tab (you have to click the "show" button).

tessarin:
I think you changed your system preferences to use _Alerts_ for iTerm, right?  
Sys prefs. > Notifications > iTerm

asmeurer:
By the way, it seems you have to do a full restart of iTerm2 when toggling the setting to fully reproduce this. If you disable native notifications then re-enable them, the ability to click the notification sticks until iTerm2 is restarted.

tessarin:
I tried in many ways but I just can't reproduce this. It always opens the view to me.

iterm

Also, a quick search on alert vs. banner style notifications didn't return any interesting results. I'm not sure if this is a macOS quirk or something relating to the implementation and Growl framework.

If you have the time, it would be interesting to compile iTerm and set a breakpoint to see if `didActivateNotification:` is firing at all.

https://github.com/gnachman/iTerm2/blob/71c164de5881ac7a7be6089dbda0750a2e8e6f22/sources/iTermNotificationController.m#L202-L205

asmeurer:
I can try, but be aware that I'm not actually an Objective-C/Swift developer. 

> Edit: also, because of the way it is implemented, a restart should be irrelevant to switching this on or off.

Yes, I noticed that restarting wasn't necessary to enable sound, for instance. So this leads me to believe that there is some bug in iTerm2's handling of this stuff.

tessarin:
If the delegate method is called but still no view comes to front then there is something I'm overlooking here. Calling for @gnachman expertise, I think he will be able to shed some light on this.

gnachman:
Does it play the system beep sound after you select the notification? That happens if the session that originated the notification isn't around any more.

asmeurer:
Nope, no beeps, even if I enable sounds for the notification itself.

gnachman:
Please make a debug log reproducing the issue using tomorrow's nightly build.

asmeurer:
I think I might have confused the issue here a bit (unless the behavior changed in a recent build, but I doubt that). 

I looks like clicking the notification does indeed open the correct tab. The issue I have here is that clicking the notification doesn't make it go away (you can even see this in @tessarin's GIF above). I think by it not going away I assumed that it was not working at all. I don't know if that is supposed to be the behavior of native notifications in "alert" mode, but I personally prefer the old way (this is also the behavior that "sticks" unless you do a full restart, by the way). 

Sorry for any wasted time on my incorrect report above.

gnachman:
Thanks for the analysis. That makes sense. Commit c8a3b66 removes notifications after you click on them.

PR #356, opened 2018-04-11
Add .sketch and .icns files for refreshed app icon

jasonlong:
@gnachman Here are the assets for the proposed refreshed app icon. I've included the `.sketch` file as well as a built `.icns`file. Note that there are several sizes of the icon, each with minor tweaks so things don't render on partial pixels, etc.

image


gnachman:
It still feels a little cramped at small sizes, like this:

screen shot 2018-04-15 at 1 36 28 pm

Could you make one where the $ and cursor are about 20% larger?

gnachman:
Can you scale it up a bit more? I'd do it myself but I don't know how the different sized images get generated. I played around with it in sketch and I think the $ looks good at this size:

icon_512x512 2x

gnachman:
Something still isn't right. If I use only the largest image from the sketch file it looks good, but when I use them all the $ gets too small again. The trouble seems to start when I include 256x256.

With only 512x512@2x:

biggest only good

With all sizes installed:

all images bad

This GIF demonstrates the difference:

jasonlong:
I'm not sure where that smaller version would be coming from. The `.icns` file can be opened in Preview.app and all of the thumbnails appear to be roughly the same. There's a slight variance at some sizes to avoid things lining up on subpixels. The `.sketch` file appears to have the correct sizes as well since this is where the source assets were exported from.

gnachman:
I replaced the 256* and 128* images with simple downscalings of the 512x512@2x image and the problem went away. I've merged the change. Thanks for the new icon!

jasonlong:
Awesome, glad you got it figured out.

Hum4n01d:
Why does the new icon have a title bar?

Hum4n01d:
It looks strange on both light and dark docks

screen shot 2018-08-12 at 12 51 33 pm

Almost blends in making the icon look short

screen shot 2018-08-12 at 12 52 19 pm

Contrast is good but again the icon doesn’t match now with the dark theme for mac

PR #357, opened 2018-05-09
Hyperlink Trigger

muhqu:
Trigger to turn text into a hyperlink.

https://gitlab.com/gnachman/iterm2/issues/4682

### DIFF
--- a/sources/VT100Grid.m
+++ b/sources/VT100Grid.m
@@ -623,6 +623,20 @@ static NSString *const kGridSizeKey = @"Size";
     }
 }
 
+- (void)setURLCode:(unsigned short)code
+        inRectFrom:(VT100GridCoord)from
+                to:(VT100GridCoord)to {
+    for (int y = from.y; y <= to.y; y++) {
+        screen_char_t *line = [self screenCharsAtLineNumber:y];
+        for (int x = from.x; x <= to.x; x++) {
+            line[x].urlCode = code;
+        }
+        [self markCharsDirty:YES
+                  inRectFrom:VT100GridCoordMake(from.x, y)
+                          to:VT100GridCoordMake(to.x, y)];
+    }
+}
+
 - (void)copyCharsFromGrid:(VT100Grid *)otherGrid {
     if (otherGrid == self) {
         return;
--- a/sources/VT100Screen.h
+++ b/sources/VT100Screen.h
@@ -123,6 +123,10 @@ extern int kVT100ScreenMinRows;
    basedAtAbsoluteLineNumber:(long long)absoluteLineNumber
                       colors:(NSDictionary *)colors;
 
+- (void)linkTextInRange:(NSRange)range
+   basedAtAbsoluteLineNumber:(long long)absoluteLineNumber
+                     URLCode:(unsigned short)code;
+
 // Load a frame from a dvr decoder.
 - (void)setFromFrame:(screen_char_t*)s len:(int)len info:(DVRFrameInfo)info;
 
--- a/sources/VT100Screen.m
+++ b/sources/VT100Screen.m
@@ -4289,18 +4289,13 @@ basedAtAbsoluteLineNumber:(long long)absoluteLineNumber
 }
 
 - (void)linkRun:(VT100GridRun)run
-       withURLCode:(unsigned short)code {
-    screen_char_t fg = { 0 };
-    screen_char_t bg = { 0 };
-    
-    fg.urlCode = code;
+    withURLCode:(unsigned short)code {
     
     for (NSValue *value in [currentGrid_ rectsForRun:run]) {
         VT100GridRect rect = [value gridRectValue];
-        [currentGrid_ setBackgroundColor:bg
-                         foregroundColor:fg
-                              inRectFrom:rect.origin
-                                      to:VT100GridRectMax(rect)];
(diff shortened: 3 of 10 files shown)


sources/HyperlinkTrigger.h
```diff
@@ -0,0 +1,12 @@
+//
+//  ReplaceTrigger.h
```
gnachman: The filename in the comment should match the actual filename

sources/HyperlinkTrigger.h
```diff
@@ -0,0 +1,12 @@
...
+//  iTerm2
+//
+//  Created by leppich on 09.05.18.
+//
+
+#import "Trigger.h"
+
+@interface HyperlinkTrigger : Trigger
```
gnachman: New classes use the iTerm prefix, so `iTermHyperlinkTrigger`.

sources/HyperlinkTrigger.m
```diff
@@ -0,0 +1,65 @@
...
+    return @"Hyperlink";
+}
+
+- (BOOL)takesParameter {
+    return YES;
+}
+
+- (NSString *)paramPlaceholder
+{
+    return @"http://\\0";
```
gnachman: Would be nice to suggest `https` instead.

sources/Trigger+ReplaceTrigger_m.h
```diff
@@ -0,0 +1,12 @@
+//
+//  Trigger+ReplaceTrigger_m.h
```
gnachman: Did you mean to include this file?

sources/VT100Screen.h
```diff
@@ -123,6 +123,10 @@ extern int kVT100ScreenMinRows;
    basedAtAbsoluteLineNumber:(long long)absoluteLineNumber
                       colors:(NSDictionary *)colors;
 
+- (void)linkTextInRange:(NSRange)range
+   basedAtAbsoluteLineNumber:(long long)absoluteLineNumber
+                        code:(unsigned short)code;
```
gnachman: `code` should be `URLCode` so that it's clear that his has to do with hyperlinks.

sources/VT100Screen.m
```diff
@@ -4277,6 +4288,22 @@ - (void)dumpScreen
     NSLog(@"%@", [self debugString]);
 }
 
+- (void)linkRun:(VT100GridRun)run
+       withCode:(unsigned short)code {
```
gnachman: Change `withCode` to `withURLCode`

sources/VT100Screen.m
```diff
@@ -4277,6 +4288,22 @@ - (void)dumpScreen
...
+    screen_char_t fg = { 0 };
+    screen_char_t bg = { 0 };
+    
+    fg.urlCode = code;
+    
+    for (NSValue *value in [currentGrid_ rectsForRun:run]) {
+        VT100GridRect rect = [value gridRectValue];
+        [currentGrid_ setBackgroundColor:bg
```
gnachman: This will have the undesirable side-effect of changing the background and foreground colors when all you want is to set the URL code. It would make sense to add a new method to `VT100Grid` like `setURLCode:inRectFrom:to:`.

sources/Trigger+ReplaceTrigger_m.h
muhqu: no… I guess this was a result of refactoring in XCode… 😕will remove it

sources/VT100Screen.m
muhqu: @gnachman you're absolutely right… I did not thought about this. I've followed your advice with 7279776.

sources/iTermHyperlinkTrigger.m
```diff
@@ -0,0 +1,69 @@
...
+    return @"Hyperlink";
+}
+
+- (BOOL)takesParameter {
+    return YES;
+}
+
+- (NSString *)paramPlaceholder
+{
+    return @"https://\\0";
```
muhqu: @gnachman for this use case, it would be nice if the `paramPlaceholder` would be used as a default value.

sources/VT100Grid.m
```diff
@@ -623,6 +623,20 @@ - (void)setBackgroundColor:(screen_char_t)bg
     }
 }
 
+- (void)setURLCode:(unsigned short)code
+        inRectFrom:(VT100GridCoord)from
+                to:(VT100GridCoord)to {
+    for (int y = from.y; y <= to.y; y++) {
+        screen_char_t *line = [self screenCharsAtLineNumber:y];
+        for (int x = from.x; x <= to.x; x++) {
+            line[x].urlCode = code;
```
muhqu: @gnachman I'm not sure, but maybe it would make sense to check there is no existing `urlCode` in a given range before changing the url… WDYT?

sources/iTermHyperlinkTrigger.m
muhqu: added a `paramDefault` method to `Trigger` in 95849ca

gnachman:
Merged as commit 275684dc1346a32d856b3b99f0510b4141da43ab with a few tweaks. Thank you!

muhqu:
@gnachman thanks for merging! And ofcourse for your continuous work on iTerm!

PR #360, opened 2018-05-17
Update the command range when the cursor moves.

yilei:
Previously when:

1. Move the cursor to the right
2. A command occupies more than one line, the cursor is at the second column and second line, and move the cursor 1 column left

The command range became incorrect, and "Select Current Commands" would select a wrong range.

Calling `-screenCommandDidChangeWithRange:` in `-terminalCursorRight:` fixes (1), `-terminalCarriageReturn:` fixes (2).

I'm not sure it's necessary in other methods, but they seem to be related.

### DIFF
--- a/sources/VT100Screen.m
+++ b/sources/VT100Screen.m
@@ -2728,7 +2728,7 @@ basedAtAbsoluteLineNumber:(long long)absoluteLineNumber
     }
     [delegate_ screenTriggerableChangeDidOccur];
     if (commandStartX_ != -1) {
-        [delegate_ screenCommandDidEndWithRange:[self commandRange]];
+        [delegate_ screenCommandDidChangeWithRange:[self commandRange]];
     }
 }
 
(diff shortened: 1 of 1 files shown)


sources/VT100Screen.m
```diff
@@ -2712,6 +2727,9 @@ - (void)terminalCarriageReturn {
         [currentGrid_ moveCursorToLeftMargin];
     }
     [delegate_ screenTriggerableChangeDidOccur];
+    if (commandStartX_ != -1) {
+        [delegate_ screenCommandDidEndWithRange:[self commandRange]];
```
gnachman: I don't think this is a good idea. `screenCommandDidEndWithRange:` should only be called when you're totally sure the command ended (you got an escape sequence saying so) or the situation is hopeless (the screen resized). A shell might use carriage return when redrawing a prompt. It would be safer to call `screenCommandDidChangeWithRange:` here.

PR #363, opened 2018-07-21
Add advanced option to customize path to ssh for ssh:// links

stefansundin:
Hi there. I am far from an expert, so please let me know if you want me to make any changes.

I've been using this modification for months, and I figured it is time that I tried to contribute it upstream.

I have a chrome extension that adds ssh:// links to the AWS console, but the only problem is that I have to add extra options, such as ProxyJump, because the servers are not accessible without tunneling through another host first.

So to accomplish this, I made iTerm run a shell script, which parses the arguments, and adds more arguments in turn. It then runs the standard ssh binary.

My shell script:
```shell
#!/bin/bash -e
extra_args=()
for arg in "$@"; do
  if [[ "$arg" == *".example.com" ]]; then
    extra_args+=("-J")
    extra_args+=("bastion.example.com")
  fi
done
echo ssh "${extra_args[@]}" "$@"
ssh "${extra_args[@]}" "$@"
```

Example ([jsfiddle](https://jsfiddle.net/ztp4mcu6/1/)):
- Clicking `ssh://server.example.com` will result in `ssh -J bastion.example.com server.example.com` being run.
- Clicking `ssh://username@server.example.com` will result in `ssh -J bastion.example.com -l username server.example.com` being run.

Let me know what your opinion is on this change. Thank you!

### DIFF
--- a/sources/iTermAdvancedSettingsModel.h
+++ b/sources/iTermAdvancedSettingsModel.h
@@ -215,6 +215,7 @@
 + (double)activeUpdateCadence;
 + (BOOL)openNewWindowAtStartup;
 + (BOOL)resetSGROnPrompt;
++ (NSString *)sshSchemePath;
 
 #if ENABLE_LOW_POWER_GPU_DETECTION
 + (BOOL)useLowPowerGPUWhenUnplugged;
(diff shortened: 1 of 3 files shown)


gnachman:
Is this really needed? You could modify ~/.ssh/config to accomplish the same, per https://wiki.gentoo.org/wiki/SSH_jump_host#Setup

stefansundin:
You are right that the example I provided above can be achieved with the ssh config file. The script that I am using is more complicated than the example I provided, and can look up IP addresses based on EC2 instance id, which is dynamic and hard to setup in the ssh config in advance (e.g. `ssh://i-01234fe342a908658-us-east-2`).

I agree that the use case is somewhat unique, so I'm ok with closing this PR and I will just continue to compile my own branch.

If you see any problems in my change however, please feel free to point them out anyway.

gnachman:
Thanks for clarifying. That sounds pretty reasonable. Let's change the description to be a little more general. How about "Command to run when handling an ssh:// URL."

stefansundin:
I updated the description.

I noticed that it is dangerous to set the setting to an empty string. Clicking on `ssh://df` is like running `df` on the users computer. How do you want me to handle this?

gnachman:
Good catch. Default it to "ssh" if it contains no alphanumeric characters.

PR #364, opened 2018-08-21
Various changes (mainly NSBundle) to make it easier to build an embeddable library/framework

lemonmojo:
- Replace `[NSImage imageNamed:]` with `[NSImage it_imageNamed:forClass:]`
- Replace `[NSBundle mainBundle]` with `[NSBundle bundleForClass:]`
- Pass `[NSBundle bundleForClass:]` to `initWithNibName:bundle:` calls
- Expose some `PTYTextView` properties and methods
  - `showTimestamps`
  - `anyAnnotationsAreVisible`
  - `selectCurrentCommand:`
  - `selectOutputOfLastCommand:`
- In `PTYSession`, handle `paste:` if not sent from an `NSMenuItem`

### DIFF
--- a/sources/PTYTextView.h
+++ b/sources/PTYTextView.h
@@ -331,6 +331,10 @@ typedef NS_ENUM(NSInteger, PTYTextViewSelectionExtensionUnit) {
 typedef void (^PTYTextViewDrawingHookBlock)(iTermTextDrawingHelper *);
 @property(nonatomic, copy) PTYTextViewDrawingHookBlock drawingHook;
 
+@property(nonatomic, assign) BOOL showTimestamps;
+
+@property(nonatomic, readonly) BOOL anyAnnotationsAreVisible;
+
 // For tests.
 @property(nonatomic, readonly) NSRect cursorFrame;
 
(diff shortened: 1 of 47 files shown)


sources/PTYTextView.h
```diff
@@ -331,6 +331,10 @@ typedef NS_ENUM(NSInteger, PTYTextViewSelectionExtensionUnit) {
 typedef void (^PTYTextViewDrawingHookBlock)(iTermTextDrawingHelper *);
 @property(nonatomic, copy) PTYTextViewDrawingHookBlock drawingHook;
 
+@property(nonatomic, assign) BOOL showTimestamps;
```
gnachman: Why was it necessary to expose these properties?

sources/PTYTextView.h
```diff
@@ -589,6 +593,9 @@ typedef void (^PTYTextViewDrawingHookBlock)(iTermTextDrawingHelper *);
 
 - (void)setTransparencyAffectsOnlyDefaultBackgroundColor:(BOOL)value;
 
+- (IBAction)selectCurrentCommand:(id)sender;
```
gnachman: Most IBActions aren't exposed in the header file. IB is able to find them in .m files. Are these needed for a particular purpose?

sources/PTYTextView.h
lemonmojo: I'm using these in my embeddable iTerm2 control to change the title/state of menu items.
For the `showTimestamps` case I'm showing `Hide Timestamps` if it's `YES` and `Show Timestamps` if it's `NO`.

sources/PTYTextView.h
lemonmojo: Of course, if you wire up those actions using IB, they can stay in the implementation file. If however you need to access those methods from another class, they need to be visible in the header. That's why I've exposed them.

lemonmojo:
@gnachman Just added exactly that.

gnachman:
Merged as ec5d3269c21a6327db7bed35e15db72c74150175. Thanks!

PR #365, opened 2018-08-27
Fix scroll-back in another way. Issue 6982

ghoostqq:
Hi, I propose to fix issue 6982 this way. Here's my thoughts:

First, it seems to me that line 107 is unnecessary and causing problems. I don't think you need to bump up remaining delta at end of scroll. If the remaining delta was opposite to the scroll direction, it can cause scroll back. So this seems like the original cause of the issue.

Second, rounding `_accumulatedDeltaY` and having a possibility that `_accumulatedDeltaY` can be opposite to the actual moving direction makes the code hard to understand intuitively, imo. 
If you think it is necessary to make the initial move easier to trigger, there should be other ways to do so.

How do you think?

### DIFF
--- a/sources/iTermScrollAccumulator.m
+++ b/sources/iTermScrollAccumulator.m
@@ -79,10 +79,13 @@
     
     // Deltas will be accumulated into _accumulatedDeltaY.
     // If it is large (>1), return its integer part. This enables quick scroll, which feels like natural trackpad.
+    // If `delta * _accumulatedDeltaY < 0`, it means turnaround. Round delta to turn around quickly (fabs 0.5 is enough to move).
     // If it is not large enough, return 0 and keep accumulating.
     if (absAccumulatedDelta > 1) {
         roundDelta = _accumulatedDeltaY > 0 ? floor(_accumulatedDeltaY) : ceil(_accumulatedDeltaY);
         _accumulatedDeltaY -= roundDelta;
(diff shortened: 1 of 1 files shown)


sources/iTermScrollAccumulator.m
```diff
@@ -92,33 +76,18 @@ - (CGFloat)accumulatedDeltaYForTrackpadEvent:(NSEvent *)event {
...
-    if (absAccumulatedDelta > 0.5) {
-        // The rounded accumulated value will not be zero so consume it.
-        roundDelta = round(_accumulatedDeltaY);
+    
+    // Deltas will be accumulated into _accumulatedDeltaY.
+    // If it is large (>1), return its integer part. This enables quick scroll, which feels like natural trackpad.
+    // If `delta * _accumulatedDeltaY < 0`, it means turnaround. Round delta to turn around quickly (fabs 0.5 is enough to move).
+    // If it is not large enough, return 0 and keep accumulating.
+    if (absAccumulatedDelta > 1) {
+        roundDelta = _accumulatedDeltaY > 0 ? floor(_accumulatedDeltaY) : ceil(_accumulatedDeltaY);
```
gnachman: There's a `RoundTowardZero` function on line 109 that does just want you want here.

gnachman:
Thank you!

PR #369, opened 2018-10-13
Fixed some typos with the GPU Renderer Unavailability reasoning.

KamranMackey:
Nothing more, nothing less. Just a few typo fixes.

The "disconnected from power" reason may be a bit long in terms of line length but I don't exactly know how to concatenate strings in Objective-C, or if you're even able to concatenate strings.

**EDIT:** Concatenated the "disconnected from power" reason. If my changes to it don't fit the normal code style though, I can always revert the commit.


gnachman:
Thanks!

PR #370, opened 2018-10-17
Improve performance and resolve memory leaks

stuartcarnie:
Commit 13091a8 resolves a memory leak which also negatively affected performance as discussed in [issue 7227](https://gitlab.com/gnachman/iterm2/issues/7227). A few additional leaks were identified and fixed along the way.

### DIFF
--- a/sources/Metal/Support/iTermCharacterSource.m
+++ b/sources/Metal/Support/iTermCharacterSource.m
@@ -542,6 +542,7 @@ static const CGFloat iTermCharacterSourceAliasedFakeBoldShiftPoints = 1;
         }
     }
     for (int i = 0; i < _numberOfIterationsNeeded; i++) {
+        CGContextRelease(_cgContexts[i]);
         _cgContexts[i] = [iTermCharacterSource newBitmapContextOfSize:_size];
     }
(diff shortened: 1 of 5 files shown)


sources/Metal/Support/iTermCharacterSource.m
```diff
@@ -542,6 +542,7 @@ - (void)initializeStateIfNeededWithFont:(CTFontRef)runFont {
         }
     }
     for (int i = 0; i < _numberOfIterationsNeeded; i++) {
+        CGContextRelease(_cgContexts[i]);
```
gnachman: This looks wrong to me because this method only gets to run once in the lifetime of the iTermCharacterSource object. If you can catch a nonnull _cgContexts[i] at this point it's a more serious bug than a simple memory leak.

gnachman: Never mind! I was looking at the wrong branch. The fix is to cherrypick 859a1b783d4dee83006772d19bd509ce23a7fa70 to master.

gnachman:
Thanks for the fixes. I've merged them.

PR #371, opened 2018-11-01
Add RequestAttention once to bounce the app icon only one time

ceafdc:


gnachman:
I don't think the proxy icon changes were meant to be in this PR. Take those out and I'll have another look.

gonzula:
Sorry, I messed up here, but I already fixed it.

gnachman:
This looks good. The iTerm2-website repo will also need to be updated. The it2attention script in this repo should probably be deleted; it's now here: https://github.com/gnachman/iterm2-website/blob/master/source/utilities/it2attention

Please send a PR to update that file and I'll accept both.

gonzula:
Ok, did it.
I will also create a PR to remove those old files from this repo.

PR #372, opened 2018-11-06
Remove files that were migrated to iterm2-website

gonzula:
As said in #371, these files have moved to https://github.com/gnachman/iterm2-website


gnachman:
There are links to imgcat and imgls on the internets, so I think their content should be changed to tell the user their new locations (https://iterm2.com/utilities/imgcat and https://iterm2.com/utilities/imgls). isiterm2.sh does not have another home, and should not be removed. The others can safely be deleted.

gonzula:
Isn't [it2check](https://github.com/gnachman/iterm2-website/blob/master/source/utilities/it2check) the same as `isiterm2.sh`? Should we change it's content to tell users the new filename and location instead of deleting it?

gnachman:
You're right, it2check is a better version of isiterm2.sh. There are lots of links out there to isiterm2.sh so that one needs to refer to the new location (https://iterm2.com/utilities/it2check)

gonzula:
Ok, done!

gnachman:
Merged, thank you!

herrtodd:
@gnachman / @gonzula this PR deleted tests/download.sh and I can't find that file in the iterm2-website repo. That file is still live at https://iterm2.com/download.sh and is referenced in sources/iTermTipData.m.

Where should I go to file a bug / make a PR to fix an issue with the download.sh script?
The current script corrupts long filenames because the gnu coreutils version of base64 wraps at 76 characters. It looks like a similar problem was just solved in imgcat here: https://github.com/gnachman/iterm2-website/commit/6da26d526f1ec69f79af3960c9d5c84957fd38c0

Thanks!

gonzula:
The file exists in the website repo, but it was named to`it2dl`. It's located [here](https://github.com/gnachman/iterm2-website/blob/master/source/utilities/it2dl). And it's a newer version than `download.sh` but doesn't have this b64 fix indeed.

gnachman:
Hrm, newlines in the filename shouldn't be a problem any more. Are you on an ancient version of iTerm2?

herrtodd:
The newline is in the base64 encoding of the filename, not the filename itself. When decoded, the filename is garbled. This is iTerm2 3.2.7b3

screen shot 2019-01-08 at 12 54 40 am

PR #373, opened 2018-11-08
Add Proxy Icon for files (not only directories)

gonzula:
At the moment, iTerm shows the proxy icon for current directory, but not for files being edited (eg. in vim). Terminal.app supports that with a escape sequence. I've implemented a similar behaviour in this PR. Here is a [vimscript](https://gist.github.com/gonzula/ee464c5e3c42628dd1c26ba50dc7f288) to test this feature (works better with neovim, but works in vim too).

I'm using the escape sequence `\[]1337;SetProxyIcon=path^G`

Here is a demonstration of this feature example


sources/PTYSession.h
```diff
@@ -276,6 +276,10 @@ typedef enum {
 // should be used.
 @property(nonatomic, readonly) NSString *windowTitle;
 
+// The path to the proxy icon thath should be used when this session is current. If is nil the current directory icon
```
gnachman: thath->that

sources/PTYSession.m
```diff
@@ -8046,6 +8046,18 @@ - (void)screenRaise:(BOOL)flag {
     }
 }
 
+// Sets current session proxy icon.
+- (void) setProxyIcon:(NSString *)value {
```
gnachman: Remove the space after `(void)`. Also, why not name this `setUserDesiredProxyIcon:` since that's really what it's doing?

sources/PTYSession.m
```diff
@@ -8046,6 +8046,18 @@ - (void)screenRaise:(BOOL)flag {
...
+    NSURL* url = nil;
```
gnachman: The `*` should have a space before it, not after it.

sources/VT100ScreenDelegate.h
```diff
@@ -82,6 +82,9 @@
 // If flag is set, bring the window to front; if not, move to back.
 - (void)screenRaise:(BOOL)flag;
 
+// Set the proxy icon of current session window.
+- (void) setProxyIcon:(NSString *)value;
```
gnachman: remove space after `(void)`. Also obey the naming convention (which is admittedly used inconsistently) to begin delegate method names with `screen`, so this would be `screenSetProxyIcon:`

sources/PTYSession.m
```diff
@@ -8046,6 +8046,18 @@ - (void)screenRaise:(BOOL)flag {
...
+    if (value) {
+        url = [NSURL fileURLWithPath:value];
+    }
+    self.userDesiredProxyIcon = url;
+    if (url) {
+        [[_delegate parentWindow] updateProxyIcon];
```
gnachman: Don't go right to the window controller from here. For one thing, this might not be the active session. It also adds dependencies on the window controller, and I'd like to remove as many of them as possible.

Instead, add a delegate method (like `sessionProxyIconDidChange:`) whose implementation in `PTYTab.m` will be similar to `sessionCurrentDirectoryDidChange:`: It should check if the session is the current session, and if so invoke a method on *its* delegate (like `tab:proxyIconDidChange:`). Since all that `updateCurrentLocation` does is mess with the proxy icon, may as well rename it to `updateProxyIcon`.

gonzula:
Ok, I fixed those problems!

sources/PTYSession.h
```diff
@@ -193,6 +193,7 @@ typedef enum {
 
 - (void)sessionCurrentDirectoryDidChange:(PTYSession *)session;
 - (void)sessionCurrentHostDidChange:(PTYSession *)session;
+- (void)sessionSetProxyIcon:(PTYSession *)session;
```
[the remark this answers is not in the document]
gonzula: `sessionDidChangeProxyIcon` or `sessionProxyIconDidChange`? The latter seems more aligned with the conventions

sources/PTYSession.m
```diff
@@ -8046,6 +8046,16 @@ - (void)screenRaise:(BOOL)flag {
     }
 }
 
+// Sets current session proxy icon.
+- (void)screenSetPreferredProxyIcon:(NSString *)value {
+    NSURL *url = nil;
+    if (value) {
```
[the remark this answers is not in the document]
gonzula: Ok, I was counting on that, and checked that produce a nil URL. But explicit is better than implicit.

gonzula:
Ok, Changed the escape sequence to OSC 6. Testing with the script in the link you provided seems to work fine with files without any special character, but it doesn’t work with files containing them. But since the value passed is an URL, not a path, I think that percent escaping should be sender's responsibility.

gonzula:
And what about the empty string to clear current icon? Should "file://" be this empty value? Since an empty string will not fire `[argument hasPrefix:@"file://"]`

gnachman:
> And what about the empty string to clear current icon? Should "file://" be this empty value? Since an empty string will not fire `[argument hasPrefix:@"file://"]`
> 

Should be fine for empty string to clear the icon. It doesn't overlap with the other OSC 7 functionality.

> Also, I think that something should be added to `Disable potentially insecure escape sequences.` description, but not quite sure what exactly to put there.

How about: "OSC 7’s proxy icon-changing feature"

gonzula:
Actually is OSC 6
but that's ok, I added it

sources/PTYSession.m
```diff
@@ -8046,6 +8046,16 @@ - (void)screenRaise:(BOOL)flag {
     }
 }
 
+// Sets current session proxy icon.
+- (void)screenSetPreferredProxyIcon:(NSString *)value {
+    NSURL *url = nil;
+    if (![value isEqualToString:@"file://"]) {  // Empty path should clear preferred proxy icon.
```
gnachman: Pass nil to `screenSetPreferredProxyIcon:` to reset it

sources/VT100Terminal.m
```diff
@@ -2498,6 +2498,12 @@ - (void)executeXtermSetPalette:(VT100Token *)token {
 
 - (void)executeXtermProprietaryExtermExtension:(VT100Token *)token {
     NSString* argument = token.string;
+    if ([argument hasPrefix:@"file://"]) {  // Support for proxy icon, following Terminal.app syntax
```
gnachman: I think it's better to take this route if the argument does not begin with a digit. That way empty string can reset it.

sources/VT100Screen.m
```diff
@@ -3441,6 +3441,10 @@ - (void)terminalStealFocus {
     [delegate_ screenRaise:YES];
 }
 
+- (void)terminalSetProxyIcon:(NSString *)value {
+    [delegate_ screenSetPreferredProxyIcon:value];
```
gnachman: I think this would be the right place to convert empty string to nil.

sources/NSStringITerm.m
```diff
@@ -1799,6 +1799,10 @@ - (BOOL)isNumeric {
     return range.location == NSNotFound;
 }
 
+- (BOOL)startsWithDigit {
+    return YES;
```
gnachman: I think this needs a little work :)

gnachman: I see we were racing. Looks goo dnow.

PR #374, opened 2018-11-14
Support option key for zoom button in compact mode

hampustagerud:
Thank you for making this awesome emulator, been using it since I first got my mac and could not imagine working without it! Also thanks for implementing the compact mode which looks real nice 🙂

This PR fixes on of the bugs with manual placement of the stoplight buttons, on macOS you can press and hold the option key to zoom instead of going fullscreen. This PR reimplements that functionality for the custom buttons.

Demo:
First window: TextEditor
Second window: iTerm2 compiled from ```master```
Third window: iTerm2 compiled from this branch with the fix applied

iterm2zoom


gnachman:
Thank you!

PR #375, opened 2018-11-14
Wrong frame position in highlighted mark (when rendering with Metal)

gonzula:
When using GPU rendering and the window height is not divisible by the character height (eg. by ctrl-dragging window's borders) the mark highlight is rendered out of place.

Here's a demonstration of this issue:

On the left is iTerm 3.2.5 and on the right is my fork

example


gnachman:
Thank you!

PR #376, opened 2018-11-22
use case-sensitive path for Cocoa

scoates:
I have a case-sensitive filesystem. The build was failing saying that it couldn't find `cocoa/cocoa.h` here. Changing the path to `Cocoa/Cocoa.h` seems to have fixed it.


gnachman:
Thanks for the fix!

PR #377, opened 2018-11-24
Add missing DLog import

nhynes:
This fixes an `Undefined symbol _DLog` error when building on Mojave. `DLog` is used in the file but not imported.


gnachman:
Thanks for the PR. I've fixed the issue.

PR #378, opened 2018-11-28
Correct "Terminal.app Compatibility" keyboard mapping "⌥←" and "⌥→"

geoff-codes:
This PR changes the "Terminal.app Compatibility" keyboard mapping to map "⌥←" and "⌥→" to "Escape"+`b` and "Escape"+`f`, respectively, which (for me at least...?) correctly implements the word-jumping functionality in Terminal.app. Seems to work with all shells. I've been manually doing this myself for several years. And I think I saw an open issue on GitLab somewhere about this... perhaps [#2587](https://gitlab.com/gnachman/iterm2/issues/2587), although I'm only fixing the 'option' bits, as 'command' left and right are not implemented in Terminal.app.

---
terminal app

iterm2

### DIFF
--- a/plists/PresetKeyMappings.plist
+++ b/plists/PresetKeyMappings.plist
@@ -727,9 +727,9 @@
       <key>0xf702-0x280000</key>
       <dict>
         <key>Action</key>
-        <integer>11</integer>
+        <integer>10</integer>
         <key>Text</key>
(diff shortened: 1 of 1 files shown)


gnachman:
Glad to see Terminal changed their weird behavior for these keys! Thanks for the fix.

PR #380, opened 2019-01-28
use profile module, not profile name string

tomviner:
These 2 methods take a `profile` param which is a string (or None).

I'm pretty sure the intention here was to access the profile module.


gnachman:
Yep, you're right. Thanks for the fix. I'll make sure to add a test for this.

PR #382, opened 2019-02-01
Make iTerm2 display a message in macOS that tells the user the app is…

pradorocchi:
… requesting elevation to manipulate the system (if iTerm2 is not included in Privacy > Full Disk Access)

This makes iTerm2 acts exactly like Apple's Terminal.app does when it need such access permission.


gnachman:
I merged this manually to tweak the string and update a few other plist files for beta and nightly builds. This has certainly been educational! I wish I could hire you full time to solve impossible problems.

PR #385, opened 2019-05-29
iterm2_git_poll.sh tweaks

rabbitt:
* use `/usr/bin/env sh` for the shebang to allow a user's environment to dictate which `sh` is run
* use locals to avoid leaking function variables
* use lowercase names for locals
* convert `~` in the received directory path to the user's `$HOME` directory
* use the more portable and clearer `test` conditional by using `[ ... ]` notation instead of the arithmetic expression `(( ... ))`
* use `command` to locate the `git` binary

### DIFF
--- a/sources/iterm2_git_poll.sh
+++ b/sources/iterm2_git_poll.sh
@@ -1,26 +1,23 @@
-#!/bin/sh
+#!/usr/bin/env sh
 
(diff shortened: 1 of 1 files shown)


gnachman:
Good catch on the Report Directory issue with tildes. That's a nasty bit of trouble there. Thanks for the fixes :)

PR #387, opened 2019-06-24
Don't assume non-nil for a function which may be nil

grimreaper:
(I couldn't figure out a way to rewrite without the nil return)


gnachman:
See 9915746cf for the right way to mark a function as returning a nil value in this cas.e

grimreaper:
Thank you!

PR #388, opened 2019-07-31
Session.async_set_variable(): Document required `user.` variable prefix

teddywing:
As described in:

proto/api.proto:597


gnachman:
Thank you!

teddywing:
Thanks for the merge!

PR #389, opened 2019-08-03
add postgres icon

will:
I'm not sure what the hex color is for, but I picked the same blue the elephant logo has just in case.


gnachman:
The icons are black and white and the hex color is used to recolor it. Makes it easy to make the color configurable in the future. It would be great to find a monochrome version of the postgres icon so it's not the odd one out.

will:
@gnachman ah okay, I didn't realize that, I'll find and/or make these black and white soon, and ping you when I got it.

will:
@gnachman Is grayscale okay? I just pushed a version that has them in that, since it seems like many of the other icons have some gray instead of just black and white. I can do black and white though if that's better.

gnachman:
I don't think it looks the way you intended after it gets recolored:

image

Changing the outlines to be transparent makes it look a little nicer:

image

klaude:
I wonder if it'd be better to use the [database icon](https://github.com/gnachman/iTerm2/blob/master/ThirdParty/hyper-tab-icons-plus/png/graphic_database%402x.png) for this. 

I was thinking about adding an icon for cockroachdb and others, but I'm worried about adding a bunch of one-off program icons to iTerm, especially when a lot of other database programs already use the generic database icon.

will:
I can see an argument either way for generic db icon vs pg specific,  and I'm not too attached one way or another. I would however make a small argument that Postgres is over 23 years old and still very popular. So maybe if a project is 15-20 years old, and still popular, it can warrant an iterm2 icon.

@gnachman I can try and figure out how to reproduce your screenshot, but I'm also happy for you to just merge what you have yourself, and ignore this PR, if you have that all ready to go and are willing to have a postgres icon. I'm not attached to having the commit in my name.

gnachman:
My icon was a 30 second Photoshop hack and isn't good enough to use in real life. It would be better to start with the vector asset and modify it before downscaling. I'm fine with adding a Postgres icon as a one-off.

pthalla:
Great!

giovannimariacusaro:
A vector version (.svg and .pdf) positive and negative ready to add.
Mark image as [template image](https://developer.apple.com/documentation/appkit/nsimage/1520017-istemplate) and set a tint like classic blue [NSColor colorWithRed:0.25 green:0.40 blue:0.55 alpha:1.00] or better a color set in .xcassets with two colors for light and dark mode.

[PostgreSQLicons.zip](https://github.com/gnachman/iTerm2/files/5012211/PostgreSQLicons.zip)

gnachman:
@giovannimariacusaro Thank you! Where did you source those files? I want to make sure there's no issue with the license.

giovannimariacusaro:
@gnachman I made them with illustrator starting from an ugly gif. Hope they can be useful!

gnachman:
Merged as 8daeb1d62f43faac0b8f82b366948e24716adaf5. Thanks all!

PR #391, opened 2019-08-14
Allow moving tmux windows between sessions by holding option/command

zeroimpl:
The tmux dashboard allows linking windows between different sessions
via drag-and-drop. This change modifies that behaviour, such that if
option or cmd are pressed during the drag, the window will be moved
instead of linked.


gnachman:
Thank you!

PR #392, opened 2019-08-15
Make options for reporting clicks and for reporting mouse wheel events independent of each other

mihao:
I want iTerm to report mouse wheel events, so that I can use scrollback in tmux. However, tmux's handling of mouse clicks is much worse than in iTerm, so I don't want clicks to be reported. This patch makes reporting clicks and reporting mouse wheel events independent of each other, which should be useful to other users of tmux.

### DIFF
--- a/Interfaces/PreferencePanel.xib
+++ b/Interfaces/PreferencePanel.xib
@@ -3612,7 +3612,7 @@ DQ
                     </connections>
                 </button>
                 <button id="EMG-WL-JwE">
-                    <rect key="frame" x="45" y="273" width="215" height="18"/>
+                    <rect key="frame" x="20" y="273" width="215" height="18"/>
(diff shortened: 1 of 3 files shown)


gnachman:
Interesting use case. I'm OK with adding support for the feature, but I don't think we can do it as proposed because it would break existing setups—people's systems would start reporting mouse wheel events where they had not done so before.

I think the best option is to add another checkbox to enable reporting movements & clicks, and have it default to enabled. Something like this:

image

mihao:
I see, your solution does look better, so I'm closing this pull request. Thanks for adding support for this feature!

gnachman:
I haven't implemented it—that was just a quick hack in interface builder to show what it could look like. If you're interested in trying to implement it feel free to re-open the PR.

PR #393, opened 2019-08-20
Specify preferences section

trodrigues:
This message confused me a bit the first time as it wasn't obvious which section of Preferences this would be on. This would make it easier for anyone to find it.


gnachman:
Thanks!

PR #394, opened 2019-08-21
Add some process names for Redis and PostgreSQL

mikl:
To support icons for PostgreSQL and Redis, here's some more process names.

I didn't include all PostgreSQL binaries, since there's a very long list, but just the most common.


gnachman:
👍

PR #396, opened 2019-08-21
Add icons for Elixir and Erlang

mikl:
While I was at it, I added icons for Elixir and Erlang.


gnachman:
My colleagues at WhatsApp will be happy :)

PR #399, opened 2019-09-13
Add a preference to allow Profiles menu to open in new windows by default

zeroimpl:
I find I always want the profile menu to open in a new window, so added an advanced option to swap to this behaviour, instead of having to press option each time.

### DIFF
--- a/sources/iTermApplicationDelegate.m
+++ b/sources/iTermApplicationDelegate.m
@@ -1440,8 +1440,8 @@ static BOOL hasBecomeActive = NO;
     [closeWindow setKeyEquivalentModifierMask:NSEventModifierFlagCommand];
 }
 
-- (void)updateAddressBookMenu:(NSNotification*)aNotification {
-    DLog(@"Updating address book menu"); // aka Profiles menu
+- (void)updateAddressBookMenu:(NSNotification *)aNotification {
+    DLog(@"Updating Profile menu");
     JournalParams params;
     if ([iTermAdvancedSettingsModel openProfilesInNewWindow]) {
         params.selector = @selector(newSessionInWindowAtIndex:);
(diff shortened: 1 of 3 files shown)


sources/iTermApplicationDelegate.m
```diff
@@ -1440,12 +1440,21 @@ - (void)nonTerminalWindowBecameKey:(NSNotification *)aNotification {
...
-    DLog(@"Updating address book menu");
+- (void)updateAddressBookMenu:(NSNotification *)aNotification {
+    DLog(@"Updating Profile menu");
     JournalParams params;
-    params.selector = @selector(newSessionInTabAtIndex:);
+    if ([iTermAdvancedSettingsModel openProfilesInNewWindow]) {
+        params.selector = @selector(newSessionInWindowAtIndex:);
+        params.alternateSelector = @selector(newSessionInTabAtIndex:);
+        [bookmarkMenu itemAtIndex:2].title = [[bookmarkMenu itemAtIndex:2].title stringByReplacingOccurrencesOfString:@"Window" withString:@"Tab"];
+        [bookmarkMenu itemAtIndex:3].title = [[bookmarkMenu itemAtIndex:3].title stringByReplacingOccurrencesOfString:@"Window" withString:@"Tab"];
```
gnachman: I think you'll also want to update the last item, which says "Open All in New Window" when holding option.

zeroimpl: So right now, that item still works the old way - will do tabs unless you hold Option.  Think it makes sense to flip it as well?

gnachman: After taking a second look, I think it's OK to keep it as is. I'll merge this.

gnachman:
Oops, I merged this by accident! I've undone it.

PR #400, opened 2019-09-17
Option to escape strings with quotes instead of backslashes

CouleeApps:
Some shells like [xonsh](https://xon.sh) don't support file paths escaped `with\ backslashes` and require `'quoted strings'` instead. This adds an experimental option to change how iTerm handles escaping to support this alternate format.

### DIFF
--- a/sources/NSStringITerm.m
+++ b/sources/NSStringITerm.m
@@ -2198,28 +2198,21 @@ static TECObjectRef CreateTECConverterForUTF8Variants(TextEncodingVariant varian
 - (void)escapeShellCharactersIncludingNewlines:(BOOL)includingNewlines {
     if ([iTermAdvancedSettingsModel escapeWithQuotes]) {
         // Only need to escape single quote and backslash in a single-quoted string
-        NSString* charsToEscape = @"\\'";
-        NSString* charsToSearch = [NSString shellEscapableCharacters];
+        NSMutableString *charsToEscape = [@"\\'" mutableCopy];
+        NSMutableCharacterSet *charsToSearch = [NSMutableCharacterSet characterSetWithCharactersInString:[NSString shellEscapableCharacters]];
         if (includingNewlines) {
-            charsToEscape = [charsToEscape stringByAppendingString:@"\r\n"];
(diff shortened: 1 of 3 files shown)


sources/NSStringITerm.m
```diff
@@ -2196,11 +2196,34 @@ - (void)escapeShellCharactersExceptTabAndNewline {
 }
 
 - (void)escapeShellCharactersIncludingNewlines:(BOOL)includingNewlines {
-    NSString* charsToEscape = [NSString shellEscapableCharacters];
-    if (includingNewlines) {
-        charsToEscape = [charsToEscape stringByAppendingString:@"\r\n"];
+    if ([iTermAdvancedSettingsModel escapeWithQuotes]) {
+        // Only need to escape single quote and backslash in a single-quoted string
+        NSString* charsToEscape = @"\\'";
```
gnachman: There should be a space before `*`, not after.

sources/NSStringITerm.m
```diff
@@ -2196,11 +2196,34 @@ - (void)escapeShellCharactersExceptTabAndNewline {
...
+        NSString* charsToSearch = [NSString shellEscapableCharacters];
```
gnachman: Instead of creating a string and then iterating over its characters, you can build a NSMutableCharacterSet and insert the characters you want to search for. Then use `-[NSString rangeOfCharacterFromSet:]` to see if any exists.

Iterating over characters in a string has a lot of ways it can fail (surrogate pairs and combining marks, for example) and is also slower than using a character set.

sources/NSStringITerm.m
```diff
@@ -2196,11 +2196,34 @@ - (void)escapeShellCharactersExceptTabAndNewline {
...
+        if (includingNewlines) {
+            charsToEscape = [charsToEscape stringByAppendingString:@"\r\n"];
+            charsToSearch = [charsToSearch stringByAppendingString:@"\r\n"];
+        }
+        bool didEscape = false;
```
gnachman: Booleans literals in objc are `YES` and `NO`

gnachman:
Thanks! Landed as 9aff539d4

PR #402, opened 2019-09-28
Fix some unevenly arranged elements in Appearance preferences

mackuba:
I've been looking at the Appearance section in the Preferences today, and I noticed some elements were badly positioned, which triggered my OCD so badly I needed to make this PR 😉 

This fixes:

- the "Theme:" label being vertically positioned one pixel differently than the other two
- the "Show window number in title bar:" checkbox/label being offset one pixel to the right
- some checkboxes in the stacks having different vertical offsets from each other (now all the spaces should be equal to 6px)

BTW, it looks like you guys might have to switch to AutoLayout in the end, because Xcode 11 seems to just turn it on without any way to turn it off… 😅

Screen Shot 2019-09-28 at 22 19 19
Screen Shot 2019-09-28 at 22 19 23

### DIFF
--- a/Interfaces/PreferencePanel.xib
+++ b/Interfaces/PreferencePanel.xib
@@ -4999,11 +4999,11 @@ DQ
             <point key="canvasLocation" x="2057" y="-22.5"/>
         </customView>
         <customView id="yCt-fi-XI9" userLabel="Prefs - Appearance" customClass="iTermSizeRememberingView">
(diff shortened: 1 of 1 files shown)


gnachman:
Thanks for the fixes! I'll look into the autolayout issue. I really hope there's a workaround, because it's a ton of complexity I can live without in this UI.

PR #404, opened 2019-10-02
Fixed xcodebuild

jkoutavas:
Added SSKeychain as a dependency to the iTerm2Shared target, thus allowing `xcodebuild -project iTerm2.xcodeproj` to successfully build iTerm2.app from the command line without a `can't locate file for: -lSSKeychain` error being generated.


gnachman:
Looks good. Thanks!

PR #406, opened 2019-10-12
More unused header removal

jkoutavas:
Another baby step towards an embeddable iTerm2 library. 

Sorry about the redundant commits in this PR, I had a bit of a cockpit error while rebasing my fork's master branch to your master branch. Though I'm an old hand at git and github, I'm kind of new at cross-fork PR submissions.

### DIFF
--- a/sources/PTYSession.h
+++ b/sources/PTYSession.h
@@ -3,7 +3,10 @@
 #import "DVR.h"
 #import "iTermAPIHelper.h"
 #import "iTermEchoProbe.h"
+#import "iTermFindDriver.h"
+#import "iTermFileDescriptorClient.h"
 #import "iTermMetalUnavailableReason.h"
+#import "iTermWeakReference.h"
 #import "iTermPopupWindowController.h"
 #import "iTermSessionNameController.h"
 
--- a/sources/PTYSession.h
+++ b/sources/PTYSession.h
@@ -1,28 +1,20 @@
 // Implements the model class for a terminal session.
 
-#import "Api.pbobjc.h"
 #import "DVR.h"
 #import "iTermAPIHelper.h"
 #import "iTermEchoProbe.h"
-#import "iTermFindDriver.h"
-#import "iTermFileDescriptorClient.h"
 #import "iTermMetalUnavailableReason.h"
-#import "iTermWeakReference.h"
-#import "ITAddressBookMgr.h"
 #import "iTermPopupWindowController.h"
 #import "iTermSessionNameController.h"
 
 #import "GPBEnumArray+iTerm.h"
-#import "LineBuffer.h"
 #import "PTYTask.h"
 #import "PTYTextView.h"
-#import "ProfileModel.h"
 #import "TextViewWrapper.h"
 #import "TmuxController.h"
 #import "TmuxGateway.h"
 #import "VT100Screen.h"
 #import "VT100ScreenMark.h"
-#import "WindowControllerInterface.h"
 #import <AppKit/AppKit.h>
 #import <Foundation/Foundation.h>
 #include <sys/time.h>
--- a/sources/SessionView.h
+++ b/sources/SessionView.h
@@ -26,7 +26,6 @@
  */
 
 #import <Cocoa/Cocoa.h>
-#import "iTermFindDriver.h"
 #import "iTermMetalDriver.h"
 #import "PTYScrollView.h"
 #import "PTYSession.h"
--- a/Interfaces/PreferencePanel.xib
+++ b/Interfaces/PreferencePanel.xib
@@ -1224,6 +1224,7 @@ DQ
                 <outlet property="_checkUpdate" destination="5048" id="Oi0-3Y-fu1"/>
                 <outlet property="_confirmClosingMultipleSessions" destination="1996" id="1wi-R8-LGb"/>
                 <outlet property="_copyLastNewline" destination="5747" id="y4f-IT-rUT"/>
+                <outlet property="_doubleClickPerformsSmartSelection" destination="wgv-Ah-fFr" id="OZU-W2-nQN"/>
                 <outlet property="_enableAPI" destination="hxX-te-qT5" id="37I-pK-Hmx"/>
                 <outlet property="_enableBonjour" destination="5051" id="lpI-Ub-bAd"/>
                 <outlet property="_evenIfThereAreNoWindows" destination="7pa-hD-WIi" id="laL-N3-lTV"/>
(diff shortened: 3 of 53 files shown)


gnachman:
Some of these shouldn't be removed. My goal is to follow the "include what you use" rule (https://github.com/include-what-you-use/include-what-you-use/blob/master/docs/WhyIWYU.md).

For example, `PTYSession.m` imports `iTermFindDriver.h` because it conforms to `iTermFindDriverDelegate`. This PR proposes to remove it. That means it is imported indirectly from `LineBuffer.h` importing `FindContext.h` importing `iTermFindDriver.h`. If I refactor `FindContext.h` to not depend on that file any more then it will rather mysteriously break code that imports `PTYSession.h`.

jkoutavas:
Okay, good point. I'll apply more due diligence as I progress. Let's leave this PR open for the time being as I delve deeper into dividing "view from app"

sources/PTYSession.h
```diff
@@ -1,28 +1,23 @@
 // Implements the model class for a terminal session.
 
-#import "Api.pbobjc.h"
```
gnachman: This is used in this file. For example, `ITMGetBufferResponse`. Anything with the `ITM` prefix comes from the protobufs. However, since only objects are ever referenced this import could be replaced with forward references like `@class ITMGetBufferResponse` for each referenced symbol.

sources/PTYSession.h
```diff
@@ -1,28 +1,23 @@
...
 #import "DVR.h"
 #import "iTermAPIHelper.h"
 #import "iTermEchoProbe.h"
 #import "iTermFindDriver.h"
 #import "iTermFileDescriptorClient.h"
 #import "iTermMetalUnavailableReason.h"
 #import "iTermWeakReference.h"
-#import "ITAddressBookMgr.h"
```
gnachman: This file uses `iTermSessionEndAction` from this include.

sources/PTYSession.h
```diff
@@ -1,28 +1,23 @@
...
 #import "iTermPopupWindowController.h"
 #import "iTermSessionNameController.h"
 
 #import "GPBEnumArray+iTerm.h"
-#import "LineBuffer.h"
 #import "PTYTask.h"
 #import "PTYTextView.h"
-#import "ProfileModel.h"
```
gnachman: We use the `Profile` macro in this file, which is defined in `ProfileModel.h`

sources/PTYSession.h
```diff
@@ -1,28 +1,23 @@
...
 #import "TextViewWrapper.h"
 #import "TmuxController.h"
 #import "TmuxGateway.h"
 #import "VT100Screen.h"
 #import "VT100ScreenMark.h"
-#import "WindowControllerInterface.h"
```
gnachman: Can replace this with `@interface WindowControllerInterface`, which is used below

sources/SessionView.h
```diff
@@ -26,7 +26,6 @@
  */
 
 #import <Cocoa/Cocoa.h>
-#import "iTermFindDriver.h"
```
gnachman: SessionView conforms to the `iTermFindDriverDelegate` protocol declared in this header

jkoutavas:
Okay, I'll roll these back.

The "big picture plan" has evolved since I started. I originally was hoping to separate-out a "ViewKit" library, which had everything necessary to provide an iTerm2 view without any window dressing or auxiliary view controllers (Preferences, WindowController, etc.), even foregoing Python scripting capabilities, along the lines of "Phase 2" that I brought up in PR #366. As I've dived in deeper, now I'm starting to see how tightly coupled these things are, and am considering just "lumping it all together" like PR #66 tried to do (and what my Phase 1 described). And then, maybe abstract-out the auxiliary view controllers after that.

The bulk of my work is currently in a local git repo called "IT2ViewKit" which uses cmake to divide things up into nice layers like "IT2Foundation", "IT2Scripting", and IT2ViewKit". Using cmake is great because it doesn't require I touch iTerm's existing xcode project at all, nor even have to commit to an "iTerm2Lib" xcode project. Cmake can generate projects at will.

Maybe, even further down the road, when my project is well underway, I may submit a PR that uses cmake to build everything, including iTerm2.app. A PR that does not require committing even iTerm2.xcodeproj to the repo. But, that's jsut pie-in-the-sky for me right now.

gnachman:
Decoupling things will be a necessary part of making it possible to create a library. I'm fine with using cmake but I don't know how to prevent breaking things as changes are made. I suppose when a new file is added both the xcode project and the makefile will need to be updated together?

jkoutavas:
> I suppose when a new file is added both the xcode project and the makefile will need to be updated together?

Yeah, pretty much like that. I'm hoping to achieve the goal of making that really easy and understandable when the time comes to present that PR.  

At the moment, I'm trying to find a clean way to avoid inclusion of iTermAPIHelper in the view library. `iTermAPIHelper.m` imports modules from all over the place.

gnachman:
> At the moment, I'm trying to find a clean way to avoid inclusion of iTermAPIHelper in the view library. iTermAPIHelper.m imports modules from all over the place. I think I can nip it off at ITermObject.

I think any attempt at cleanly separating views out from other stuff will require adding a lot of new protocols. Right now the dependency graph is full of cycles because I've been optimizing for velocity over flexibility. I could imagine a world where the API helper is a delegate of the various classes that call directly into it, and then they could tolerate its non-existence.

jkoutavas:
@gnachman, it's been a few days since my last comment re:Big Picture. Let me fill you in on what I've been to.

I paused my work on sorting-out the headers and focused on ensuring that "iTerm2Lib" will be the right solution for my project. The area I'm exploring right now is "how to trick iTerm2Lib into using a pty driver that does a remote connection to a server." You see, I want to use /just/ the text rendering aspects of iTerm2, I'm not interested in opening a local shell, I'm interested in opening a connection to a remote MUD server. I'm thinking I could make a custom pty process/driver that connects to the server and then have iTerm2Lib open that pty session. Problem is, I'm new to dealing with tty/pty dev concepts and am not sure if even such a thing is possible. I'm a little out of my experience zone here. Do you know much about this subject? Do you think I'm taking the right approach of redirecting at the pty level like this? Is there an easier approach?

gnachman:
I made a rough sketch of how to refactor PTYTask to enable this in the `refactor_ptytask` branch. That branch is not shippable because it has bugs. If we're going to make this real the refactor will need to proceed methodically and probably add some unit tests. I'll continue to work on it because I will need this for an upcoming change to reduce the number of daemon processes.

jkoutavas:
> I made a rough sketch of how to refactor PTYTask to enable this in the `refactor_ptytask` branch.

I looked over this branch a few days ago, but I didn't see how I might be able to reuse things to do a remote socket. Is such a thing possible?

Maybe instead of integrating a socket, I have the session do the remote connect as part of its "shell" script?

gnachman:
Good news, I've landed that change.

You need to write a class that implements `iTermJobManager`.  You don't need to implement `attachToServer:withProcessID:task:` as long as `isSessionRestorationPossible` returns NO.

The most important thing is to vend a file descriptor in your `-fd` method and to implement `forkAndExecWithTtyState:…` to create a new connection. Use `iTermLegacyJobManager` as your model, since it's simple-ish.

> Maybe instead of integrating a socket, I have the session do the remote connect as part of its "shell" script?

What do you mean by "'shell' script"? If you mean replacing its `shell` ivar with an implementation other than PTYTask I don't recommend that.

soheil (commented):
ok nice

mikeknack86 (approved):

mikeknack86:
Ok

mikeknack86:
<details><summary>Details</summary>
<p>



</p>
</details>

PR #407, opened 2019-10-14
Add Nixpkgs-specific make flag to fix build with nix

tricktron:
### Problem
[Nix](https://nixos.org/nix/) has its default home directory under `/private/var/empty`. As a result, Xcode sets the build directory relative to this home directory. Since Mojave the user has no write permission in that directory and the build fails.

### Solution
Set the build directory to a path where Nix has write permission. This is done with the 
`-derivedDataPath` flag to `xcodebuild`.

### Other
 - I also updated the path where the executable lies.
Before: `build/Deployment`
New: `Build/Products/Deployment`

You may want to change them for the other Makefile flags as well.

- Added empty `CODE_SIGN_IDENTITY`and `DEVELOPMENT_TEAM` to fix codesign issue.

### My Setup
- macOS Catalina 10.15
- Xcode 11.1

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -40,6 +40,10 @@ Deployment:
 	xcodebuild -parallelizeTargets -target iTerm2 -configuration Deployment && \
 	chmod -R go+rX build/Deployment
 
+Nix:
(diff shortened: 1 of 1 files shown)


gnachman:
The reason for the funny build path is that I configure xcode that way so it's easy to find builds, but it's not the default. But this looks fine, and Nix looks interesting!

PR #409, opened 2019-12-01
Handle synthetic directories correctly when opening new tabs

erikolofsson:
In Catalina there is support for creating symlinks in the read only root directory through `/etc/synthetic.conf`. See `man synthetic.conf` for details.

This pull request makes "Reuse previous session's directory" use the remapped directory (`/Dir`) instead of the target (usually `/System/Volumes/Data/Dir`)


gnachman:
Landed as d2aacb1e7. Thank you!

PR #410, opened 2019-12-15
Adds command+shift+t keyboard shortcut to 'Duplicate Tab' under 'Shell'

einichi:
This shortcut appears to not otherwise be in use. I use Duplicate Tab for opening multiple tabs to work on different things in the same directory, but have to click several times to get the tabs I want. A keyboard shortcut would make this experience nicer and I suspect there may be other users who would also benefit from this.

### DIFF
--- a/Interfaces/MainMenu.xib
+++ b/Interfaces/MainMenu.xib
@@ -144,8 +144,8 @@
                                 </connections>
                             </menuItem>
                             <menuItem isSeparatorItem="YES" id="1199"/>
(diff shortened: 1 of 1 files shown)


gnachman:
Cmd-Shift-T is already bound to toggling whether tabs are shown in full screen:

image

If it were available, I'd use it for undo close tab/window to match web browsers.

I recommend binding it in keyboard prefs:

image

einichi:
Good point regarding undo close tab/window, and thanks for pointing out key binding.

PR #411, opened 2020-01-16
Tooltip documentation for tab title setting

jjb:
The tab title setting has identical tooltip documentation to the setting above it.

Let me know what the copy should be and I'll update the PR!

Screen Shot 2020-01-15 at 9 35 05 PM
Screen Shot 2020-01-15 at 9 35 32 PM


gnachman:
Looks like this one never got documented! I've added it to the 3.3 docs with the following text:

> The OSC 0 and OSC 2 control sequences set the window title. This setting controls whether such a control sequence changes the window title associated with all sessions in the window, or only with the one in which it was received.

gnachman:
0871441d319239ade785b00151d387b34f0a8319

PR #412, opened 2020-01-25
Highlight text and keep foreground color

kjmph:
When highlighting text, it doesn't look nice to change all the foreground colors to black. Here is a proposed patch that will keep the color of the font.

I'm not a regular iTerm2 dev, I just wanted to fix the colors.. If there is a need to make this into an option in user preferences, I would need help.

### DIFF
--- a/sources/iTermMetalPerFrameState.m
+++ b/sources/iTermMetalPerFrameState.m
@@ -1245,10 +1245,6 @@ ambiguousIsDoubleWidth:(BOOL)ambiguousIsDoubleWidth
         // Blue link text.
         rawColor = VectorForColor([_configuration->_colorMap colorForKey:kColorMapLink]);
         caches->havePreviousCharacterAttributes = NO;
-    } else if (selected) {
-        // Selected text.
-        rawColor = VectorForColor([colorMap colorForKey:kColorMapSelectedText]);
-        caches->havePreviousCharacterAttributes = NO;
     } else if (_configuration->_reverseVideo &&
                ((c->foregroundColor == ALTSEM_DEFAULT && c->foregroundColorMode == ColorModeAlternate) ||
                 (c->foregroundColor == ALTSEM_CURSOR && c->foregroundColorMode == ColorModeAlternate))) {
--- a/sources/iTermTextDrawingHelper.m
+++ b/sources/iTermTextDrawingHelper.m
@@ -1891,10 +1891,6 @@ NSColor *iTermTextDrawingHelperGetTextColor(iTermTextDrawingHelper *self,
         // Blue link text.
         rawColor = [context->colorMap colorForKey:kColorMapLink];
         context->havePreviousCharacterAttributes = NO;
-    } else if (context->hasSelectedText) {
-        // Selected text.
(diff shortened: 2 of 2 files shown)


kjmph:
It appears to be an unrelated error that causes the build checks to fail.

gnachman:
You can change the selection color without modifying the code:

Screen Shot 2020-01-27 at 9 09 48 PM

gnachman:
Oops, circled the wrong one. Look at the color under the one I circled.

kjmph:
Hello @gnachman, Yep, I appreciate the pointer there. However, this profile setting changes the background and foreground color to rendered with these colors. This would override any attributes being displayed in the terminal output via escape codes. The patch actually does something different, it leaves alone the color of the foreground, so the outputted coloring shows through the selection (while the background color is still rendered). With iTerm2 being the premier way to get 24bit color support on a Mac, it is a shame to lose all coloring when selecting items.

Note, I recognize this isn't a complete patch, and I'm not surprised it is closed, I think more needs to be changed in the user preferences before it would be accepted. However, note that it is modifying the user experience to leave alone their foreground, and not have a Selected Text color at all.

Another way to express this change is to allow the user to set an alpha channel in the Selected Text color picker, and someone with my preferences could set opacity to 0%, and have the terminal's output color be rendered.

gnachman:
I think the right way to solve this is to make "selected text" into a checkbox, just like "tab color" is. When disabled, the default color would be used.

I'm concerned that this would cause text to be illegible when it's similar to the selection background color, though. How would you deal with that problem?

kjmph:
Yes, the checkbox would be fabulous. iTerm can provide the default, and users like me can disable it.

RE: foreground color matching selection's background color. In practice, it doesn't come up a lot for typical settings, however, you are correct that this would make it hard to read when it does come up. I know this discussion goes back a long way in terminal software. I know some other terminals invert foreground and background as the selection highlight to avoid this issue, (it is a nice choice, I just don't like all the white filling the screen). I've long thought that automatic contrast adjustment would be the best choice here, I believe I've used a few terminals that did that in the past. I think with the configurability of iTerm as it stands today, users that want a dark grey text and want a selection text color not applied (this new user preference proposed in this thread), they can choose a different selection background color.

matthesoundman:
Since this is closed, any chance there is another feature targeted to allow foreground preservation?

## Terminal
Terminal


## iTerm
iTerm

gnachman:
Hm, interesting idea. Yes, please file a feature request at https://iterm2.com/bugs

kjmph:
Wouldn't the pull request resolve that bug @matthesoundman ?

matthesoundman:
@kjmph I am not sure. I am very new to this. What is the status of your pull request?

@gnachman I did as you requested and filed bug [#8761](https://gitlab.com/gnachman/iterm2/-/issues/8761)

kjmph:
As is, it preserves foreground colors when selecting text. Some optionality is requested to extend the pull request, yet I get lost tracing through the iTerm2 preference panes.

cyal1:
Is there a way to display the length of the selected text

PR #413, opened 2020-02-02
Observe effectiveAppearance instead of AppleInterfaceStyle

saagarjha:
Tested on macOS Catalina.


gnachman:
I decided to refactor this code a bit more, incorporating these ideas, in commit 0fe504caa. Thanks!

PR #414, opened 2020-02-15
fix(keys): Correctly map 'increase height' shortcut.

jessestuart:
### What does this PR do?
Fixes a bug in the "Keys" menu, where assigning a keybinding to "Increase Height" instead assigns to "Decrease Height" — looks like this was [introduced](https://github.com/gnachman/iTerm2/commit/2a58778cb62c8318ba003f8991fcdeab122cedbc#diff-d685851ee8f00202cb41d7229dba8234R135) as part of a refactoring a couple weeks ago.

Screengrab demonstrating issue:

Screen Capture on 2020-02-15 at 17-32-58


gnachman:
Thank you!

jessestuart:
You bet! Thank *you* for all the work you put into iTerm — can't imagine my workflow without it. 👍

PR #417, opened 2020-04-11
Add iTerm variables for user's Selection

aid:
Purpose of this pull request is to permit iTerm2 Python scripts to act upon the user's selection; which does not previously seem to be possible.

This is performed by creating two new iTerm Variables that a Python script is able to watch:

```
selectionLength   - The length of the user's original selection.
selection         - A constrained version of the user's section.
```

There is also a new Advanced option defined:

```
maximumBytesToProvideToPythonAPI	100
```

The `selection` iTerm variable is constrained to a maximum length as defined by the above option.

This allows for a Python API `coro` function of the form:

```python
async def statusbar_selection_coro(
        knobs,
        selection=iterm2.Reference('selection'),
        selection_length=iterm2.Reference('selectionLength')):
```


gnachman:
Have you seen these?

 * https://iterm2.com/python-api/selection.html
 * https://iterm2.com/python-api/session.html#iterm2.Session.async_get_selection
 * https://iterm2.com/python-api/session.html#iterm2.Session.async_get_selection_text
 * https://iterm2.com/python-api/session.html#iterm2.Session.async_set_selection

The variable would give you an easy way to tell when the selection changes.

aid:
Yes, I did see those function calls – an earlier version of my python script used `async_get_selection_text` within the click handler to grab the selection.

_(Although I don't think I worked out how to update the text within the selection bar from within the click handler – I was just creating a popup containing the data at that point.)_

This was inspired by Vicky Singh's recent post to the iTerm mailing list regarding presenting date stamps in BST.  I'd never noticed iTerm's decode of selected data in the content menu – and I have recently been decoding such values hundreds of times manually.  I thought presenting such decoded data down in the status bar would be an interesting alternative – and a good excuse to learn the Python API.

Having the user need to click on the status bar item to ensure the status bar item is up-to-date would not lead to good user experience – so, as you suggest, the primary reason for these new variables is as a trigger for the python `coro` function to ensure the status bar always shows timely information.

sources/iTermController.m
```diff
@@ -247,6 +248,28 @@ - (PTYSession *)anyTmuxSession {
...
+    if (current_session) {
+        iTermVariableScope * variablesScope = current_session.variablesScope;
+        if (variablesScope) {
+            if (selection && selection.length > 0) {
+                int max_length = [iTermAdvancedSettingsModel maximumBytesToProvideToPythonAPI];
+                [variablesScope setValue:[selection substringToIndex:MIN(max_length,selection.length)] forVariableNamed:iTermVariableKeySessionSelection];
+                [variablesScope setValue:[NSNumber numberWithInt:selection.length] forVariableNamed:iTermVariableKeySessionSelectionLength];
+            } else {
+                [variablesScope setValue:@"" forVariableNamed:iTermVariableKeySessionSelection];
+                [variablesScope setValue:[NSNumber numberWithInt:0] forVariableNamed:iTermVariableKeySessionSelectionLength];
```
gnachman, line 267:
```suggestion
                const int maxLength = [iTermAdvancedSettingsModel maximumBytesToProvideToPythonAPI];
                [variablesScope setValue:[selection substringToIndex:MIN(maxLength, selection.length)] forVariableNamed:iTermVariableKeySessionSelection];
                [variablesScope setValue:@(selection.length) forVariableNamed:iTermVariableKeySessionSelectionLength];
            } else {
                [variablesScope setValue:@"" forVariableNamed:iTermVariableKeySessionSelection];
                [variablesScope setValue:@0 forVariableNamed:iTermVariableKeySessionSelectionLength];
```

sources/iTermController.m
```diff
@@ -247,6 +248,28 @@ - (PTYSession *)anyTmuxSession {
...
 
+- (void)updateSelectionVariables:(NSString *)selection {
+    // Assign the whole selection to the internal _lastSelection variable but only
+    // maximumBytesToProvideToPythonAPI characters to the "selection" iTerm Variable
+    //
+    // The "selectionLength" iTerm variable contains the full length of the original
+    // selection; not the restricted length assigned to the "selection" iTerm Variable
+    PTYSession * current_session = [_frontTerminalWindowController currentSession];
+    if (current_session) {
+        iTermVariableScope * variablesScope = current_session.variablesScope;
```
gnachman, line 259:
```suggestion
        iTermVariableScope *variablesScope = current_session.variablesScope;
```

sources/PTYTextView.m
```diff
@@ -4697,6 +4697,9 @@ - (void)selectionDidChange:(iTermSelection *)selection {
             selection = nil;
         }
         [[iTermController sharedInstance] setLastSelection:selection];
+        [[iTermController sharedInstance] updateSelectionVariables:selection];
+    } else {
+        [[iTermController sharedInstance] updateSelectionVariables:nil];
```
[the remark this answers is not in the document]
aid: `selection` – as an `NSString *` – doesn't exist in the else code path.  This function starts with `selection` as an `iTermSelection *` which is later shadowed by an `NSString *` – but not in the else path.

I think either it needs to stay as is with the repeated call (but using your delegate suggestion) or the function starts off with an `NSString * selectionString = nil` - with that possibly being updated in the if statement and then call to the delegate outside of the if/else block – picking up either nil or the string of the selection.  The shadowing of `selection` feels bad to me; so I'll go this second route for now...

sources/PTYTextView.m
```diff
@@ -4697,6 +4697,9 @@ - (void)selectionDidChange:(iTermSelection *)selection {
             selection = nil;
         }
         [[iTermController sharedInstance] setLastSelection:selection];
+        [[iTermController sharedInstance] updateSelectionVariables:selection];
```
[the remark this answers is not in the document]
aid: Thanks for the hint; assigning the variables within `PTYSession` is far neater.

sources/PTYSession.h
```diff
@@ -728,6 +728,9 @@ backgroundColor:(NSColor *)backgroundColor;
 // Call refresh on the textview and schedule a timer if anything is blinking.
 - (void)refresh;
 
+// Update PTYSession variables on selection change
+- (void)selectionDidChange:(NSString *)selection;
```
gnachman, line 732: It’s good enough that the declaration is in `PTYTextViewDelegate`. It shouldn’t be here since it’s not meant to  be a public API.

sources/PTYSession.m
```diff
@@ -7556,6 +7558,24 @@ - (void)mitigateTouchBarStupidityForEvent:(NSEvent *)event {
...
+    //
+    // The "selectionLength" iTerm variable contains the full length of the original
+    // selection; not the restricted length assigned to the "selection" iTerm Variable
+    DLog(@"selectionDidChange to %@", selection);
+    
+    if (selection && selection.length > 0) {
+        const int maxLength = [iTermAdvancedSettingsModel maximumBytesToProvideToPythonAPI];
+        [self.variablesScope setValue:[selection substringToIndex:MIN(maxLength, selection.length)] forVariableNamed:iTermVariableKeySessionSelection];
+        [self.variablesScope setValue:@(selection.length) forVariableNamed:iTermVariableKeySessionSelectionLength];
+    } else {
```
gnachman, line 7573: If we guarantee that `selection` is nonnull then the else clause can be eliminated.

sources/PTYTextView.h
```diff
@@ -78,6 +78,7 @@ typedef NS_ENUM(NSInteger, PTYCharType) {
 - (void)sendHexCode:(NSString *)codes;
 - (void)sendText:(NSString *)text;
 - (void)sendTextSlowly:(NSString *)text;
+- (void)selectionDidChange:(NSString *)selection;
```
gnachman, line 81: New code should use the modern naming convention for delegates. Some day I'll fix the old code :)

```suggestion
- (void)textViewSelectionDidChangeToTruncatedString:(NSString *)selection;
```

sources/PTYTextView.m
```diff
@@ -4688,16 +4688,18 @@ - (void)_settingsChanged:(NSNotification *)notification {
...
     DLog(@"selectionDidChange to %@", selection);
     [_delegate refresh];
     if (!_selection.live && selection.hasSelection) {
         const NSInteger MAX_SELECTION_SIZE = 10 * 1000 * 1000;
-        NSString *selection = [self selectedTextWithCappedAtSize:MAX_SELECTION_SIZE minimumLineNumber:0];
-        if (selection.length == MAX_SELECTION_SIZE) {
-            selection = nil;
+        selectionString = [self selectedTextWithCappedAtSize:MAX_SELECTION_SIZE minimumLineNumber:0];
+        if (selectionString.length == MAX_SELECTION_SIZE) {
+            selectionString = nil;
```
gnachman, line 4698: The reason `setLastSelection:` takes nil if it’s oversize is to avoid pasting a truncated value later on (the “last selection” is used when pasting from last-selected-text rather than from pasteboard). This isn’t a problem for the variable, though, so let's make sure the value passed to the delegate is never nil.

sources/iTermAdvancedSettingsModel.m
```diff
@@ -327,6 +327,7 @@ + (void)enumerateDictionaries:(void (^)(NSDictionary *))block {
 DEFINE_FLOAT(idleTimeSeconds, 2, SECTION_GENERAL @"Time in seconds before a session is considered idle.\nUsed for updating icons and activity indicator in tabs.");
 DEFINE_FLOAT(findDelaySeconds, 1, SECTION_GENERAL @"Time to wait before performing Find action on 1- or 2- character queries.");
 DEFINE_INT(maximumBytesToProvideToServices, 100000, SECTION_GENERAL @"Maximum number of bytes of selection to provide to Services.\nA large value here can cause performance issues when you have a big selection.");
+DEFINE_INT(maximumBytesToProvideToPythonAPI, 100, SECTION_GENERAL @"Maximum number of bytes of selection to provide to PythonAPI.\nA large value here can cause performance issues when you have a big selection.");
```
gnachman, line 330:
```suggestion
DEFINE_INT(maximumBytesToProvideToPythonAPI, 100, SECTION_GENERAL @"Maximum number of bytes of selection to provide to Python API.\nA large value here can cause performance issues when you have a big selection.");
```

sources/PTYSession.m
aid: I think that rather than ensuring this delegate is never called with a nil – which is impossible in ObjC anyway – we could accept the nil as an indicator of no selection.

We can then use `selection = selection ? selection : @"";` towards the start of the delegate to ensure that a valid `NSString` is available when the iTerm variables are set – allowing the for the removal of the if/else and the almost-duplicate `setValue` lines.  Seem OK?

sources/PTYTextView.m
aid: I was purposely leaving the operation of lastSelection alone, assuming something along those lines (with just the variable name changed so not to shadow the method's parameter).

With regards to ensuring the delegate isn't provided nil:  That would be awkward as if the `(!_selection.live && selection.hasSelection)` clause is not true there is no NSString available to pass to the delegate –  we'd need to create a new empty string for the delegate call should that clause be false.

I guess I could do:

```
[_delegate textViewSelectionDidChangeToTruncatedString:selectionString ? selectionString : @""];
```

Which is basically the same as what I have now in the delegate – but I really dislike knowing that the delegate would SEG fault if it were called with nil – perhaps by some future function.

aid:
Updated version based upon latest comments now pushed into this branch...

sources/PTYTextView.m
gnachman: If you initialize selectionString to `@""` on line 4691 and then avoid nilling it prior to the call to `setLastSelection:` you can satisfy both the need to pass `nil` to `setLastSelection:` on truncate and the need to always pass the selection to `textViewSelectionDidChangeToTruncatedString:`. Something like
```
NSString *selectionString = @"";
…
  if (!_selection.live && selection.hasSelection) {
    …
    selectionString = [self selectedTextWithCappedAtSize:MAX_SELECTION_SIZE minimumLineNumber:0];
    if (selectionString.length == MAX_SELECTIONSIZE) {
      [[iTermController sharedInstance] setLastSelection:nil];
    } else {
      [[iTermController sharedInstance] setLastSelection:selectionString];
    }
  }
  [_delegate textViewSelectionDidChangeToTruncatedString:selectionString];
```

sources/PTYSession.h
```diff
@@ -728,6 +728,8 @@ backgroundColor:(NSColor *)backgroundColor;
 // Call refresh on the textview and schedule a timer if anything is blinking.
 - (void)refresh;
 
+- (void)textViewSelectionDidChangeToTruncatedString:(NSString *)selection;
```
gnachman, line 731: Delete this line. It's sufficient to declare it in the delegate protocol.

sources/PTYSession.m
```diff
@@ -7556,6 +7558,20 @@ - (void)mitigateTouchBarStupidityForEvent:(NSEvent *)event {
...
 
+- (void)textViewSelectionDidChangeToTruncatedString:(NSString *)selection {
+    // Assign a maximum of maximumBytesToProvideToPythonAPI characters to the "selection"
+    // iTerm Variable
+    //
+    // The "selectionLength" iTerm variable contains the full length of the original
+    // selection; not the restricted length assigned to the "selection" iTerm Variable
+    DLog(@"textViewSelectionDidChangeToTruncatedString: %@", selection);
+
+    selection = selection ? selection : @"";
```
gnachman, line 7569: Don’t assign to an argument. Rename the argument to `maybeSelection` and then do `NSString *selection = maybeSelection ?: @""`.

sources/PTYTextView.m
aid: OK, updated as per your suggestion – although I used a ternary operator  on `setLastSelection` to remove duplication.

sources/PTYSession.h
aid: Weird; I thought I'd already removed that.   Gone for sure now....

sources/PTYSession.m
aid: Done.  I didn't know of the `<cond> ?: <fallback>` syntax - thanks.

aid:
Updates now committed to github branch.

gnachman:
Thank you very much!

PR #420, opened 2020-06-17
Only apply 'exclude from app switcher' setting when all windows are hotkey windows

andystanton:
Hi,

This is a prototype to only apply the 'exclude from app switcher' setting when all windows are hotkey windows. Why?

I have two modes of operation for iTerm:
1. To quickly enter transient commands.
2. For long lived vim/ssh/log sessions.

Hotkey windows are ideally suited for the first case, and standard windows are great for the second. The pedant in me dislikes having iTerm appear in the app switcher when I only have a hotkey window open, but having iTerm in the app switcher is required for the second use case. I want the best of both worlds, so I implemented this prototype PR to achieve that:

**iTerm2 excluded from app switcher when only a hotkey window is open:**
iTerm2 excluded from app switcher when only a hotkey window is open

**iTerm2 included in app switcher when a non-hotkey window is open:**
iTerm2 included in app switcher when a non-hotkey window is open

I labeled this prototype because I am not well-versed in either Objective C or Apple UI apps, so before I commit more time to learning how to make the change properly, I wanted to ask a few questions:

1. Is there an existing way to achieve what I'm trying to do here?
2. Is this something you would accept as part of iTerm2?
If so:
3. Is the approach along the right lines?
4. Should it be configured separately from the existing settings?

Thanks,

-Andy

### DIFF
--- a/sources/iTermAdvancedSettingsModel.m
+++ b/sources/iTermAdvancedSettingsModel.m
@@ -315,7 +315,7 @@ DEFINE_BOOL(dockIconTogglesWindow, NO, SECTION_HOTKEY @"If the only window is a 
 DEFINE_BOOL(hotkeyWindowFloatsAboveOtherWindows, NO, SECTION_HOTKEY @"The hotkey window floats above other windows even when another application is active.\nYou must disable “Prefs > Keys > Hotkey window hides when focus is lost” for this setting to be effective.");
 DEFINE_FLOAT(hotKeyDoubleTapMaxDelay, 0.3, SECTION_HOTKEY @"The maximum amount of time allowed between presses of a modifier key when performing a modifier double-tap.");
 DEFINE_FLOAT(hotKeyDoubleTapMinDelay, 0.01, SECTION_HOTKEY @"The minimum amount of time required between presses of a modifier key when performing a modifier double-tap.");
-DEFINE_BOOL(automaticallyToggleLSUI, YES_IF_BETA_ELSE_NO, SECTION_HOTKEY @"Automatically sets iTerm2 as LSUIElement when the “Exclude iTerm2 from dock and app switcher” setting is enabled AND only Hotkey windows are open. Unsets iTerm2 as LSUIElement when the “Exclude iTerm2 from dock and app switcher” setting is enabled AND at least one non-Hotkey window is open.");
+DEFINE_BOOL(automaticallyToggleLSUI, YES_IF_BETA_ELSE_NO, SECTION_HOTKEY @"Disable “Exclude iTerm2 from Dock and ⌘-Tab App Switcher” when any windows are not hotkey windows.");
 
 #pragma mark General
 
--- a/sources/iTermAdvancedSettingsModel.m
+++ b/sources/iTermAdvancedSettingsModel.m
@@ -315,7 +315,7 @@ DEFINE_BOOL(dockIconTogglesWindow, NO, SECTION_HOTKEY @"If the only window is a 
 DEFINE_BOOL(hotkeyWindowFloatsAboveOtherWindows, NO, SECTION_HOTKEY @"The hotkey window floats above other windows even when another application is active.\nYou must disable “Prefs > Keys > Hotkey window hides when focus is lost” for this setting to be effective.");
 DEFINE_FLOAT(hotKeyDoubleTapMaxDelay, 0.3, SECTION_HOTKEY @"The maximum amount of time allowed between presses of a modifier key when performing a modifier double-tap.");
 DEFINE_FLOAT(hotKeyDoubleTapMinDelay, 0.01, SECTION_HOTKEY @"The minimum amount of time required between presses of a modifier key when performing a modifier double-tap.");
-DEFINE_BOOL(automaticallyToggleLSUI, YES_IF_BETA_ELSE_NO, SECTION_HOTKEY @"Automatically sets iTerm2 as LSUIElement when the 'Exclude iTerm2 from dock and app switcher' setting is enabled AND only Hotkey windows are open. Unsets iTerm2 as LSUIElement when the 'Exclude iTerm2 from dock and app switcher' setting is enabled AND at least one non-Hotkey window is open.");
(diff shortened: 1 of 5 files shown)


gnachman:
> Is there an existing way to achieve what I'm trying to do here?

No.

> Is this something you would accept as part of iTerm2?

I think the idea makes sense. My concern is that it's hard to predict how people will use a feature, so this will almost certainly break someone's workflow in ways that will only become apparent after it lands. If that turns out to be a lot of people, then we have a problem.

The risk can be mitigated by enabling it in beta first and seeing if anyone hollers.

> Is the approach along the right lines?

Kind of. You want to change the mode when a window is created or closed, but the existing implementation affects tabs. The easiest place to add the change is in `iTermController.m`. It manages an array called `_terminalWindows` that gets updated when a window is created or destroyed. Conveniently, it doesn't include accessory windows like prefs or the profile opener. So that's where the check for which mode to be in ought to be triggered.

> Should it be configured separately from the existing settings?

Add an advanced pref to iTermAdvancedSettingsModel.[hm] called `automaticallyDisableLSUI` in the "experimental" section. Set its default value to `YES_IF_BETA_ELSE_NO`. There are lots of other boolean prefs to use as a model. Then gate the feature on `[iTermAdvancedSettingsModel automaticallyDisableLSUI]`.

andystanton:
Many thanks for the pointers George. I've implemented your suggestions.

sources/iTermAdvancedSettingsModel.m
```diff
@@ -315,6 +315,7 @@ + (void)enumerateDictionaries:(void (^)(NSDictionary *))block {
 DEFINE_BOOL(hotkeyWindowFloatsAboveOtherWindows, NO, SECTION_HOTKEY @"The hotkey window floats above other windows even when another application is active.\nYou must disable “Prefs > Keys > Hotkey window hides when focus is lost” for this setting to be effective.");
 DEFINE_FLOAT(hotKeyDoubleTapMaxDelay, 0.3, SECTION_HOTKEY @"The maximum amount of time allowed between presses of a modifier key when performing a modifier double-tap.");
 DEFINE_FLOAT(hotKeyDoubleTapMinDelay, 0.01, SECTION_HOTKEY @"The minimum amount of time required between presses of a modifier key when performing a modifier double-tap.");
+DEFINE_BOOL(automaticallyToggleLSUI, YES_IF_BETA_ELSE_NO, SECTION_HOTKEY @"Automatically sets iTerm2 as LSUIElement when the 'Exclude iTerm2 from dock and app switcher' setting is enabled AND only Hotkey windows are open. Unsets iTerm2 as LSUIElement when the 'Exclude iTerm2 from dock and app switcher' setting is enabled AND at least one non-Hotkey window is open.");
```
gnachman, line 318: Use curly quotes instead of straight single quotes. Opening quote: “ Closing quote: ”

sources/iTermApplicationDelegate.m
```diff
@@ -2063,7 +2063,17 @@ - (IBAction)loadRecording:(id)sender {
 #pragma mark - Private
 
 - (void)updateProcessType {
-    [[iTermApplication sharedApplication] setIsUIElement:[iTermPreferences boolForKey:kPreferenceKeyUIElement]];
+    bool enableLSUI = [iTermPreferences boolForKey:kPreferenceKeyUIElement];
```
gnachman, line 2066: `BOOL` is how Objective C does it, while `bool` is a c++ism that snuck into a header file somewhere

sources/iTermApplicationDelegate.m
```diff
@@ -2063,7 +2063,17 @@ - (IBAction)loadRecording:(id)sender {
...
+
+    if ([iTermAdvancedSettingsModel automaticallyToggleLSUI]) {
+        bool onlyHotKeyWindowsOpen = true;
```
gnachman, line 2069: `bool`->`BOOL`
`true` -> `YES`

sources/iTermApplicationDelegate.m
```diff
@@ -2063,7 +2063,17 @@ - (IBAction)loadRecording:(id)sender {
...
+        for (PseudoTerminal* term in [self terminals]) {
```
gnachman, line 2070: The `*` should touch `term`, not `PseudoTerminal`

sources/iTermApplicationDelegate.m
```diff
@@ -2063,7 +2063,17 @@ - (IBAction)loadRecording:(id)sender {
...
+            onlyHotKeyWindowsOpen &= [term isHotKeyWindow];
```
gnachman, line 2071: A BOOL is a char, which means there are 255 nonzero values that are true. That makes bitwise and unsafe, unfortunately.

A better way to do this is:
```
const BOOL onlyHotKeyWindowsOpen = [self.terminals allWithBlock:^BOOL(PseudoTerminal *term) { return term.isHotKeyWindow; }];
```

sources/iTermApplicationDelegate.m
```diff
@@ -2063,7 +2063,17 @@ - (IBAction)loadRecording:(id)sender {
...
+        }
+        enableLSUI &= onlyHotKeyWindowsOpen;
```
gnachman, line 2073: `&=` unsafe

sources/iTermController.m
```diff
@@ -228,6 +230,11 @@ - (void)updateWindowTitles {
     }
 }
 
+- (void)updateProcessType {
+    [[NSNotificationCenter defaultCenter] postNotificationName:iTermProcessTypeDidChangeNotification
+    object:nil];
```
gnachman, line 235: Align the colon on this line with the colon on the preceding line

sources/iTermApplicationDelegate.m
andystanton: Thanks - I went with your functional suggestion below.

andystanton:
Thanks very much for taking the time to review the PR and suggest idiomatic Objective C @gnachman

gnachman:
Thank you!

PR #423, opened 2020-07-05
Allow full range of blurRadius for 10.14+ and add option to create a new window with the current profile

anthroid:
Full range of blurRadius is working on 10.14 without the artifacts noted in issue #6115 and #6138. No way that I know of to create a new window with the current profile (other than directly linking a specific profile to a keyboard shortcut), so I added this option in the Shell menu because I find myself always needing it. Right now, you have to create a 'New Tab with Current Profile' and then drag the tab off the window, and this is cumbersome.

### DIFF
--- a/Interfaces/PreferencePanel.xib
+++ b/Interfaces/PreferencePanel.xib
@@ -3248,7 +3248,7 @@ DQ
                 <slider verticalHuggingPriority="750" fixedFrame="YES" translatesAutoresizingMaskIntoConstraints="NO" id="5363">
                     <rect key="frame" x="36" y="283" width="218" height="21"/>
                     <autoresizingMask key="autoresizingMask" flexibleMaxX="YES" flexibleMinY="YES"/>
-                    <sliderCell key="cell" continuous="YES" alignment="left" minValue="0.10000000000000001" maxValue="30" doubleValue="0.10000000000000001" tickMarkPosition="below" sliderType="linear" id="5368"/>
+                    <sliderCell key="cell" continuous="YES" alignment="left" minValue="0.10000000000000001" maxValue="100" doubleValue="25" tickMarkPosition="below" sliderType="linear" id="5368"/>
                     <connections>
                         <action selector="settingChanged:" target="BMA-OA-Kzc" id="AJB-o1-kDf"/>
                         <outlet property="nextKeyView" destination="4973" id="5373"/>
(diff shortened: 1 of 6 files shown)


sources/PTYTab.m
```diff
@@ -2460,10 +2460,12 @@ - (double)blurRadius {
         }
     }
     if (count > 0) {
-        if (@available(macOS 10.13, *)) {
+        if (@available(macOS 10.14, *)) {
+            return sum / count;
```
anthroid, line 2464: This is intended to cascade from macOS versions highest -> lowest due to the nature of `@available`. If we have 10.14, then we are assumed to be on 10.14-10.16. Clip the value for 10.13, and allow normal range for 10.12 and under.

Interfaces/PreferencePanel.xib
```diff
@@ -3248,7 +3248,7 @@ DQ
                 <slider verticalHuggingPriority="750" fixedFrame="YES" translatesAutoresizingMaskIntoConstraints="NO" id="5363">
                     <rect key="frame" x="36" y="283" width="218" height="21"/>
                     <autoresizingMask key="autoresizingMask" flexibleMaxX="YES" flexibleMinY="YES"/>
-                    <sliderCell key="cell" continuous="YES" alignment="left" minValue="0.10000000000000001" maxValue="30" doubleValue="0.10000000000000001" tickMarkPosition="below" sliderType="linear" id="5368"/>
+                    <sliderCell key="cell" continuous="YES" alignment="left" minValue="0.10000000000000001" maxValue="100" doubleValue="25" tickMarkPosition="below" sliderType="linear" id="5368"/>
```
anthroid, line 3251: This renders fine with full blur radius (100) and drop shadow without any chopping or visible performance issues on 10.14. Can't test on any other versions.

sources/iTermController.m
```diff
@@ -275,6 +275,34 @@ - (void)newWindow:(id)sender possiblyTmux:(BOOL)possiblyTmux {
     }
 }
 
+- (void)newWindowWithSameProfile:(id)sender possiblyTmux:(BOOL)possiblyTmux {
+    Profile *bookmark = nil;
+    if (_frontTerminalWindowController) {
```
gnachman, line 280: There's a bunch of duplicate code with the previous method and with `newSessionWithSameProfile`. Please refactor it out.

anthroid: Will do.

anthroid:
Closing this and will resubmit each feature as a separate PR.

PR #426, opened 2020-07-22
Allow switching tabs using '^ Number'

nhynes:
Same request as https://gitlab.com/gnachman/iterm2/-/issues/7387, but now atop master. Allowing tab switch using `^` makes the shortcuts more consistent with browsers.


gnachman:
Merged with changes as 9f071a0d709f853af8ac2fad6ab75c020d7a57e4

PR #427, opened 2020-08-14
Extract usernames from txt-record entries in bonjour/mDNS advertisements

zero0cool0:
Maybe worth making this configurable, too. Currently extracts usernames that are set in mDNS txt-records with u=[username] or username=[username].

I've tested this against AVAHI running on Ubuntu and Raspberian. The default AVAHI sshd.service file needs to be modified and a txt-record needs to be added manually.

The use case here is that the ssh advertisement sends a default username which makes it easier for server administrators to log in remotely if the local username is not the same as the remote one.

### DIFF
--- a/sources/ITAddressBookMgr.m
+++ b/sources/ITAddressBookMgr.m
@@ -387,7 +387,8 @@ iTermWindowType iTermThemedWindowType(iTermWindowType windowType) {
 - (void)_addBonjourHostProfileWithName:(NSString *)serviceName
                        ipAddressString:(NSString *)ipAddressString
                                   port:(int)port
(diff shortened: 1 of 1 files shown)


gnachman:
Is `u=` or `username=` a standard of some sort?

zero0cool0:
I don't know to be honest.

It's referenced [here](https://kodi.wiki/view/Avahi_Zeroconf) and [here (in German)](https://wiki.ubuntuusers.de/Avahi/)

The txt-record itself is conform to the mDNS standard which uses standard DNS elements as payload.

gnachman:
Merged, thank you!

PR #429, opened 2020-10-07
Added support for floating point timestamps

vkalinsky:
Right-click on timestamp shows its human-readable value in local timezone. It's a great feature! Huge thanks to someone who added it.

This PR adds parsing of floating point timestamps, like 1602085036.828, where the whole number represents seconds since epoch.


gnachman:
Thanks for the suggestion. I think it's time to use regexes for this. Done in commit 05ad9b573f97be11002f8c1e768d47ff4f4da280.

PR #430, opened 2020-12-20
speed up decoding small images

Cyberbeni:
Relates to: https://gitlab.com/gnachman/iterm2/-/issues/9379
I tested it and this makes my prompt faster with images inside it even in Debug. They are all around 2kB (1 tall, 2 wide PNG images), so I set the limit to 4kB.
I guess I'll leave that issue open as using image path could still be beneficial to some.

### DIFF
--- a/sources/iTermImage.m
+++ b/sources/iTermImage.m
@@ -14,13 +14,13 @@
 
 static const CGFloat kMaxDimension = 10000;
 
-#define DECODE_IMAGES_IN_PROCESS 0
+#define FORCE_DECODE_IMAGES_IN_PROCESS 0
 
 #if defined(__has_feature)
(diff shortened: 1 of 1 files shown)


gnachman:
Out-of-process decoding is for security reasons. Probably could make it faster using a better form of IPC, or using XPC instead of forking a new subprocess each time (although I'm not sure how it could be run in a custom sandbox).

Cyberbeni:
We are using XPC to communicate between our app and our Finder Extension. I'll probably have some time this year to make at least a PoC to be able to test the performance. (But I can't make promises about it, family is always #1 for me)

I'll close this PR as these changes would have to be reverted anyways.

PR #432, opened 2020-12-23
XPC image decoding

Cyberbeni:
@gnachman This is where I got to today. Feel free to give any inputs (or continue my work), I don't know when I will have time to continue this.

It works really well in terms of performance: 
Screenshot 2020-12-23 at 20 07 09

IIRC these kind of XPC services don't need connection validation, we had to use a Login Item XPC service, so the Finder Extension can connect to it and then we had to add audit token validation.


gnachman:
Thanks for putting this together.

I think my main question is how restrictive this sandbox is as compared to the current sandboxing rules:

https://github.com/gnachman/iTerm2/blob/386cbe063615dc8a46063b3c6dc228cad62db4d5/OtherResources/image_decoder.sb

I understand the standard app sandbox is pretty restrictive, but is it enough to run untrusted code if someone gets an RCE?

Cyberbeni:
Oh yeah, it can read files (and write to certain locations) and make network requests with the most restrictive setup from Xcode. Luckily sandbox_init can add additional restrictions. (Deprecated in 10.8 and I assume the currently used sandbox-exec was deprecated at the same time)

`kSBXProfilePureComputation` is perfect for regular images but I assume those extra file access permissions are for sixel support. It seems to not like calling sandbox_init twice with different profiles to pass both NoNetwork and NoWrite.

Would "NoNetwork" with the default sandboxed write permissions be sufficient or should we just accept that decoding sixel will stay the same as it is now?

Cyberbeni:
`kSBXProfilePureComputation` is not good for all images, it is consistently crashing with this gif: 
[pepoCheer.txt](https://github.com/gnachman/iTerm2/files/5742488/pepoCheer.txt)

```console
* thread #4, queue = 'com.apple.NSXPCConnection.user.anonymous.9308'
    frame #0: 0x00007fff2032a462 libsystem_kernel.dylib`__pthread_kill + 10
    frame #1: 0x00007fff20358610 libsystem_pthread.dylib`pthread_kill + 263
    frame #2: 0x00007fff202ab720 libsystem_c.dylib`abort + 120
    frame #3: 0x00007fff256c2496 HIServices`_RegisterApplication + 13699
    frame #4: 0x00007fff256bee28 HIServices`GetCurrentProcess + 23
    frame #5: 0x00007fff286a5abf HIToolbox`MenuBarInstance::GetAggregateUIMode(unsigned int*, unsigned int*) + 63
    frame #6: 0x00007fff22c3a733 AppKit`_NSScreenConfigurationUpdateSharedInfoForReason + 531
    frame #7: 0x00007fff22c3a4d6 AppKit`___NSScreenConfigurationEnsureInitialUpdateOccurred_block_invoke + 105
    frame #8: 0x00007fff201ad7c7 libdispatch.dylib`_dispatch_client_callout + 8
    frame #9: 0x00007fff201ae96b libdispatch.dylib`_dispatch_once_callout + 20
    frame #10: 0x00007fff22c8a9d9 AppKit`+[_NSScreenConfiguration latestGreatestBackingScaleFactor] + 135
    frame #11: 0x00007fff22db1c35 AppKit`NSImageCTMForReferenceContext_hints_ + 763
    frame #12: 0x00007fff22db179a AppKit`-[NSSnapshotContextSignature initWithDrawingRect:applicableForRect:context:hints:] + 96
    frame #13: 0x00007fff22db16e5 AppKit`-[NSCGImageSnapshotRep initWithCGImage:drawingRect:applicableForRect:context:hints:] + 314
    frame #14: 0x00007fff22db1576 AppKit`-[NSImage initWithCGImage:size:] + 312
  * frame #15: 0x000000010a6dbfee iTerm2SandboxedWorker`-[iTermImage initWithData:](self=0x00007ffa32705240, _cmd="initWithData:", data=2174 bytes) at iTermImage.m:154:38
    frame #16: 0x000000010a6d7b98 iTerm2SandboxedWorker`-[iTerm2SandboxedWorker decodeImageFromData:withReply:](self=0x00007ffa32509170, _cmd="decodeImageFromData:withReply:", imageData=2174 bytes, reply=0x00007fff21434ff0) at iTerm2SandboxedWorker.m:16:11
    frame #17: 0x00007fff211e74be Foundation`__NSXPCCONNECTION_IS_CALLING_OUT_TO_EXPORTED_OBJECT_S2__ + 10
    frame #18: 0x00007fff211e653e Foundation`-[NSXPCConnection _decodeAndInvokeMessageWithEvent:flags:] + 2213
    frame #19: 0x00007fff2119dd49 Foundation`message_handler + 206
    frame #20: 0x00007fff2009cc28 libxpc.dylib`_xpc_connection_call_event_handler + 56
    frame #21: 0x00007fff2009ba9c libxpc.dylib`_xpc_connection_mach_event + 935
    frame #22: 0x00007fff201ad867 libdispatch.dylib`_dispatch_client_callout4 + 9
    frame #23: 0x00007fff201c4a47 libdispatch.dylib`_dispatch_mach_msg_invoke + 441
    frame #24: 0x00007fff201b34a7 libdispatch.dylib`_dispatch_lane_serial_drain + 263
    frame #25: 0x00007fff201c55b8 libdispatch.dylib`_dispatch_mach_invoke + 498
    frame #26: 0x00007fff201b34a7 libdispatch.dylib`_dispatch_lane_serial_drain + 263
    frame #27: 0x00007fff201b40fe libdispatch.dylib`_dispatch_lane_invoke + 426
    frame #28: 0x00007fff201bdc5d libdispatch.dylib`_dispatch_workloop_worker_thread + 819
    frame #29: 0x00007fff20355499 libsystem_pthread.dylib`_pthread_wqthread + 314
    frame #30: 0x00007fff20354467 libsystem_pthread.dylib`start_wqthread + 15
```

Cyberbeni:
Found another way to decode GIFs that work with `kSBXProfilePureComputation`

Cyberbeni:
Okay, seems like sixel can work with the stricter sandbox too. When trying to decode PDF, I got a similar crash as with the old GIF parsing, so now normal images that don't use `NSBitmapImageRep` will be routed through the current image_decoder, which supposedly also works on macOS 10.12 now.

Cyberbeni:
Almost missed that images should be transported as raw data and OFC the original solution crashed here too. Luckily there was an easy solution and it's even faster than encoding NSImage normally.

Cyberbeni:
Decoding my test pdf file 100 times went from ~20 seconds to ~15 seconds by switching to using NSSecureCoding. If I understand it correctly, base64 encoding here is not necessary here, right?

Cyberbeni:
Bundle identifier still has to be fixed by you but otherwise it should be ready to :shipit:

gnachman:
Great job on a really high quality PR. Merged as commit 1585ba8dbf0a46c852b050e93c5d89f9c9b8e6eb. Thank you very much :D

PR #435, opened 2021-01-19
Recheck blur when changing window styles

anthroid:
When changing window styles using `Window > Window Style`, the window's blur setting is temporarily lost until the window regains focus. This re-enables the blur after modifying the window style.


gnachman:
Thank you!

PR #437, opened 2021-02-14
Recheck blur when updating tab bar theme

anthroid:
When changing the tab bar theme in `Preferences > Appearance > Theme`, all active windows lose their blur until each window is clicked individually or regains focus. This re-enables blur after changing the tab bar theme.


gnachman:
Thanks!

PR #438, opened 2021-03-03
provide gitlab description templates

gl-njanz:
added three GitLab [description (issue) templates](https://docs.gitlab.com/ee/user/project/description_templates.html), for three different cases:

- Bug reports
- Crash report
- Perfomance issues

this could replace the default issue template and could clear up some confusion over what information is required by the reporter for each issue type.
 it also automatically applies labels on issue creation. the labels would need to be created on the [project (gitlab.com/gnachman/iterm2) manually once](https://docs.gitlab.com/ee/user/project/labels.html#project-labels), though.

take it for a test drive over here 😉 : https://gitlab.com/niklasjanz/iterm2/-/issues/new
the templates can also be referenced in the URL directly, like so: https://gitlab.com/niklasjanz/iterm2/-/issues/new?issuable_template=Crash

### DIFF
--- a/.gitlab/issue_templates/Bug.md
+++ b/.gitlab/issue_templates/Bug.md
@@ -1,15 +1,15 @@
 # Bug report
 
-<!-- Thanks for filing an issue! Please answer the questions below so I can help you. -->
+Thanks for filing an issue! Please answer the questions below so I can help you.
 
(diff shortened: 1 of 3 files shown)


gnachman (approved):

gnachman:
Thanks!

PR #440, opened 2021-03-20
Fix "show border" description in preferences panel

netzverweigerer:
The setting has zero effect when not using transparency, so I propose changing the description for the setting to make clear this only applies to transparent windows.


gnachman:
It's also effective in light mode. It is only in dark mode + all panes in current tab are opaque that it has no effect.

Since this is too hard to fit into a label, maybe put it in the tooltip instead?

netzverweigerer:
If it doesn't fit into the label because it's too long, then I +1 to your suggestion to reject my PR and let's put it into the tooltip instead, sure.

gnachman:
Done in commit 7315afd61

PR #443, opened 2021-04-26
Fix status bar shows incorrect network throughput when network interface change

GetToSet:
The network throughputs displayed on the status bar are calculated in a way that sums up bytes  sent/received from all interfaces and then calculates its diff with previous sample. In cases where user adds or removes a network interface (which can happen when toggling Wi-Fi or VPN / proxy client), the instant value calculated and displayed can be incorrect, even with negative values.

2021-04-26 AM 10 02 10

This pull request solves the above issue by comparing significant network interfaces count with previous value when looping through interfaces and if they don't match, which signals an interface change has happened, same throughputs as previous will be displayed rather than trying to calculate them from bytes, which will definitely produce wrong result.

### DIFF
--- a/sources/iTermNetworkUtilization.m
+++ b/sources/iTermNetworkUtilization.m
@@ -12,6 +12,7 @@
 #include <sys/sysctl.h>
 #include <netinet/in.h>
 #include <net/if.h>
+#include <net/if_dl.h>
 #include <net/route.h>
 
 typedef struct {
@@ -49,7 +50,7 @@ typedef struct {
     iTermPublisher<iTermNetworkUtilizationSample *> *_publisher;
     iTermNetworkUtilizationStats _last;
     // Used to detect network interface change
-    NSUInteger lastInterfaceCount;
+    NSSet *lastInterfaceAddrs;
     BOOL interfaceHasChanged;
 }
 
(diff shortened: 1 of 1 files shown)


gnachman:
Would it be better to look at the set of MAC addresses instead? A count could be misleading if an interface is added and another is removed between updates.

GetToSet:
> Would it be better to look at the set of MAC addresses instead? A count could be misleading if an interface is added and another is removed between updates.

Yeah maybe it's a better approach, I'll try to figure it out.

GetToSet:
I've updated to take MAC address into account when detecting if there is any changes in network interfaces.

I used a the way same with `getifaddrs (3)` ([getifaddrs.c](https://opensource.apple.com/source/Libinfo/Libinfo-542.40.3/gen.subproj/getifaddrs.c.auto.html))  to get `sockaddr_dl` structure from interface information returned as `RTM_IFINFO2` type, this structure contains interface name and (optionally) associated MAC address, which is then cached into a NSSet to compare against.

For potential performance cost, time consumed by this additional check should be tolerable according to  following profile result.
Screnshot 2021-05-06 PM 2 30 21

GetToSet:
Further investigation shows that interface change that resulting this bug can only happen when actions like toggling a VPN or plugging in external network adapters such iOS hotspot. Toggling Wi-Fi won't result such sensorial since Wi-Fi interface is still there (but switched to/from the active one).

gnachman (approved):

PR #444, opened 2021-04-27
 Fix preferences pane search bar style on Big Sur.

GetToSet:
The search bar on iTerm preference pane has a hover background like normal toolbar items, which is slightly different from search bar seen on toolbars is other Apps like Finder.
Screenshot 2021-04-27 AM 10 38 47

After some time diving into is issue, the solution is quite simple, just exclude it from `toolbarSelectableItemIdentifiers`.

This pull request removes `searchFieldToolbarItem` from `toolbarSelectableItemIdentifiers` to make it non-selectable, thus removes the hover background. A `NSToolbarFlexibleSpaceItemIdentifier` is also added between other items and the search bar item to achieve proper layout on Big Sur.

Since this change may introduce compatibility issue with older systems. I've tested it on Big Sur 11.2, Catalina 10.15.7 & Mojave 10.14.6.

- On Big Sur 11.2

Screenshot 2021-04-27 AM 10 47 45

- On Catalina 10.15.7

Screenshot 2021-04-26 AM 9 53 35

### DIFF
--- a/sources/PreferencePanel.m
+++ b/sources/PreferencePanel.m
@@ -744,7 +744,7 @@ andEditComponentWithIdentifier:(NSString *)identifier
 }
 #endif
 
-- (NSArray *)orderedToolbarIdentifiers {
+- (NSArray *)orderedToolbarIdentifiersWithExcludesSearch:(BOOL)excludesSearch {
     if (!_globalToolbarItem) {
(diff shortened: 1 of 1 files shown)


gnachman:
Thanks!

PR #447, opened 2021-07-09
Update 'nightly' and 'beta' icons

peterlewis:
I love riding on the betas, but I don't think that the additional text fits with the rest of the iTerm icon.

I would like to propose making a [totally subjective] subtle tweak to the 'beta' and 'nightly' icons, ridding-them of their additional text and instead having a little fun with the `$`.

I appreciate that the icon used to feature the beta symbol, but [was changed so as to remove potential vulgarity](https://github.com/gnachman/iTerm2/commit/79d52a1edef78a3874d23c4794d1c237729833d7#diff-1a5a23bdc23cc5049bb62fb67a184909585a64b02998a984ec647a5e70e5796e). This revision should avoid the potential for  offence due to being a only a single symbol.

I would love your thoughts on this and THANK YOU to @gnachman and all contributors for making the most-used application of my entire life! x


peterlewis:
Oops! Sorry, had a sudden panic that I had used a different typeface—so had closed whilst I confirmed—but all looks good on that front! :)

gnachman:
I agree that these look better. Can you generate iconsets for them as well? These go in `Media.xcassets/AppIcon-{Beta,Nightly}.appiconset`.

Instructions on creating an icon set are here: https://www.codingforentrepreneurs.com/blog/create-icns-icons-for-macos-apps

peterlewis:
Ah thank you for being so receptive to the suggestion! :) Oh yes of course, good point! Those have now been added!

As an aside, just FYI - by running the `.png`s through [ImageOptim](https://github.com/ImageOptim/ImageOptim) it was possible to reduce (lossless) the file size by approximately 50%. Granted, that's only 100kb or so, but every little helps, right?

Anyway, if this is of interest to you then I've drafted a PR, updating the release assets, that you can find [here](https://github.com/gnachman/iTerm2/pull/448). I'm happy to close if it's not required/desired, though! :)

peterlewis:
Ah blast, I've just noticed there's an issue with the fill. Closing whilst I resolve (so as to prevent merge).

peterlewis:
Sorry, realised that closing the PR significantly reduced ease-of-access, so I'll just keep it open with a big ol' DO NOT MERGE message, here.

So, looks that the issue lies with the `iTermIcon.sketch` file, which has the grey listed as `#202B2E` whereas the actual app icon appears to be  `#131C1F`.

I would just update the gray, but it appears that the same colour discrepancy issue is present with the greens as well (and I haven't yet looked into the gradients!). Is there something awry/outdated with the `.sketch` file or am I doing something wrong, here?

Thanks!

Peter

gnachman:
It is entirely possible that the sketch file is out of date. Let's use the current production icon colors are the source of truth.

The background should be SRGB #131c1f and then green should be SRGB #04e518 for the $, ɑ, and β and #076310 for the cursor.

peterlewis:
Thanks for that, and a darn good job that the colours were out of whack, because I realised that my Sketch had kindly substituted-out Source Code Pro for a different font! :-/

Anyway, the font has now been resolved (it is now Source Code Pro) and the colours fixed (as per your comment, above). However, the gradient on the bezel is still not right and not as refined as the production icon (as can be seen, below):

Current `beta-1024.png`:

beta-1024

New `beta-1024.png`:

beta-1024

gnachman:
Yikes, what a mess. I don't know why the colors are different but we should stick with the actual colors in use. Commit cb2a0a73db5b6758e74d510bbe4236b022b2f2af updates the sketch file to match the production icons.

peterlewis:
Ah, gimmie two mins. My ImageOptim settings were stripping the colour profile. Just re-running, leaving the profile intact.

peterlewis:
Right, sorted. Apologies for that.

gnachman:
Merged as 4abdefa9814e266d24935d3edb03138c38f01341. Thank you!

PR #448, opened 2021-07-10
Optimise release app icon

peterlewis:
Reduced icon size by approx 100kb by running the source `.png`s through [ImageOptim](https://github.com/ImageOptim/ImageOptim)


gnachman:
The iconset looks good but release.png has the same color discrepancy as your other PR

peterlewis:
Ah yes, so it does! I have re-added the discrepancy to `release.png`, so as to align with how it is currently!

peterlewis:
This PR is confusing me, what with the discrepancy that is present in the existing `release.png` but not in the `release-resolution.png`. Am going to close this for now. I can happily re-open for optimisation of release icons, if needs-be.

PR #449, opened 2021-07-19
Fix a PTYSession retain cycle

saagarjha:
We forgot to use weakSelf in the observer block, keeping the entire session (and all its lines) alive even after the window had closed.


gnachman:
Thanks for this.

PR #450, opened 2021-07-21
Fix bad indentation

goranjviv:
a few tabs were breaking the entire `iterm2` package. `itomate` was also broken because of this


gnachman:
Thank you!

PR #451, opened 2021-08-24
Fix vertical cursor height

hetranger:
The height of the vertical cursor does not change when line spacing < 1.
image
However, the height of the box cursor does change.
image
Trying to make them behave consistently.


gnachman (approved):

gnachman:
Thank you!

PR #452, opened 2021-09-08
Improve Fish shell integration script

kidonng:
Hi George, recently I created a [standalone plugin](https://github.com/kidonng/iterm2.fish) based on the Fish integration script. I'd like to submit some of the changes back to iTerm2:

- Replace `[` `]` with `test`
- Replace `sh -c "exit $last_status"` with `return $last_status`
- Replace `grep` with Fish's `string match` builtin
- Replace `wc` with Fish's `count` builtin
- Simplify `iterm2_has_fish_mode_prompt` by directly using its return status

I hope you will like these improvements!


gnachman:
Committed as 2808b6ae5d2cc1426709d0a09792248887bfa3b1, sorry for the delay!

PR #454, opened 2021-11-04
Support opening to line number in all JetBrains IDEs

shawwn:
From iTerm's documentation:

> If you hold cmd and click on a URL it will be opened. If you hold cmd and click on a filename, it will be opened. There is special support for MacVim, TextMate, and BBEdit when you cmd-click on a text file's name: if it is followed by a colon and line number, the file will be opened at that line number.

Currently, iTerm can open to line number for IDEA, but it doesn't support any other IntelliJ editors like WebStorm, PyCharm, etc.

This PR implements support for every JetBrains IDE, since they can all handle the `--line` argument on command line.

The one tricky change was:

```objc
- (void)launchAppWithBundleIdentifier:(NSString *)bundleIdentifier args:(NSArray *)args {
    args = [@[ @"-nb", bundleIdentifier, @"--args" ] arrayByAddingObjectsFromArray: args];
    [self launchTaskWithPath:@"/usr/bin/open" arguments:args completion:nil];
}
```

In order to open JetBrains IDEs correctly, it's necessary to use e.g. `open -nb com.jetbrains.WebStorm --args --line 4 index.js`. Simply invoking the executable directly will result in strange errors. (Note that their official docs use `open -n`: https://www.jetbrains.com/help/pycharm/working-with-the-ide-features-from-command-line.html#toolbox)

### DIFF
--- a/sources/iTermSemanticHistoryController.m
+++ b/sources/iTermSemanticHistoryController.m
@@ -207,24 +207,9 @@ NSString *const kSemanticHistoryColumnNumberKey = @"semanticHistory.columnNumber
     return executable;
 }
 
-- (NSString *)intelliJIDEALauncherInApplicationBundle:(NSBundle *)bundle {
-    DLog(@"Trying to find IntelliJ IDEA launcher in %@", bundle.bundlePath);
-    return [[[[bundle.bundleURL URLByAppendingPathComponent:@"Contents"] URLByAppendingPathComponent:@"MacOS"] URLByAppendingPathComponent:@"idea"] path];
-}
-
 - (void)launchAppWithBundleIdentifier:(NSString *)bundleIdentifier args:(NSArray *)args {
-    NSBundle *bundle = [self applicationBundleWithIdentifier:bundleIdentifier];
-    if (!bundle) {
-        DLog(@"No bundle for %@", bundleIdentifier);
-        return;
-    }
-    NSString *executable = [self executableInApplicationBundle:bundle];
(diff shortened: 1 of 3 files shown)


gnachman:
It's hard to know what effect this change will have. I would recommend using the new method of opening only for JetBrains IDEs and leave it alone for all other apps.

lydell:
This would be amazing! I’m using JetBrains Rider and would love to have goto-line.

gnachman:
I've merged a commit based on this, but `open -nb com.jetbrains.WebStorm --args --line 4 file.js` doesn't seem to work. It's fine for Rider though. Let me know if there's another JetBrains IDE I should add support for.

lydell:
Awesome! Rider is the only one I need currently.

josb:
Could `--column 123` be added if there's another colon and number present? That is, with a file name like `/path/to/file:LINE:COLUMN`. Some tools emit this enhanced format.

Unfortunately, to implement this cleanly a breaking change would have to be made which should be clearly announced in advance, given that the `\3` substitution token (which would be the natural candidate for this improvement) sadly wasn't reserved for this purpose and is already in use. `\6` could be used instead but that's ugly given that line and column numbers naturally belong together.

gnachman:
@josb Sorry,for which editor(s)?

josb:
@gnachman I imagine all JetBrains IDEs support this option, but in my case I am specifically talking about IntelliJ IDEA Ultimate.

PR #455, opened 2021-12-07
Allow new window from dock when daily tip is present.

GetToSet:
Allow opening new window from dock when all iTerm windows are closed while 'Tip of the Day' is present, by making `iTermTipWindowController`'s window a subclass of `NSPanel`.


gnachman:
Thank you!

PR #457, opened 2022-01-10
Allow requesting Bluetooth permission

dlech:
Command line programs that use CoreBluetooth will crash with SIGABRT if this is not included in info.plist. The message doesn't seem to actually be shown, but [CentralManager.didUpdateState:](https://developer.apple.com/documentation/corebluetooth/cbcentralmanagerdelegate/1518888-centralmanagerdidupdatestate?language=objc) will be called with the state set to [CBManagerStateUnauthorized](https://developer.apple.com/documentation/corebluetooth/cbmanagerstate/cbmanagerstateunauthorized?language=objc) unless the users enables Bluetooth permission in *System Preferences*.


gnachman:
LOL apple didn't document this as supported on macOS but I guess it is…sorry for the delay in getting to this PR, looks safe to merge.

PR #459, opened 2022-01-19
Improve input speed for tmux integration

lonnywong:
# make tmux integration input faster

## Issue

More technical details at [10179: Improve input speed for tmux integration](https://gitlab.com/gnachman/iterm2/-/issues/10179).

## tmux send-keys

The `-l` flag disables key name lookup and processes the keys as literal UTF-8 characters.

Most input can be sent directly. No need to convert to hex, which is 5 times slower.

## simple test code

I wrote a simple project to test the piece of code: [TmuxSendKeys](https://github.com/lonnywong/TmuxSendKeys/blob/main/TmuxSendKeys/TmuxGateway.m)

Make sure that any bytes from `0x00` to `0xFF`  is processed correctly.

## test with iTerm2

1. login to the server via `telnet`
2. use `tcpdump` to capture tcp packets
3. run `tmux -CC` and do a some works
4. analyze the tcp packets

## tested tmux version
* 1.8
* 2.1
* 2.6
* 3.0a

### DIFF
--- a/sources/TmuxGateway.m
+++ b/sources/TmuxGateway.m
@@ -856,13 +856,48 @@ error:
     }
 
     // Send multiple small send-keys commands because commands longer than 1024 bytes crash tmux 1.8.
+    const NSUInteger maxLiteralCharacters = 1000;  // 1024 - len('send -lt %123456789 ') = 1004
+    const NSUInteger maxHexCharacters = maxLiteralCharacters / 8;  // len(' C-Space') = 8
+
+    BOOL asLiteralCharacters = NO;
     NSMutableArray *commands = [NSMutableArray array];
-    const NSUInteger stride = 80;
-    for (NSUInteger start = 0; start < codePoints.count; start += stride) {
-        NSUInteger length = MIN(stride, codePoints.count - start);
-        NSRange range = NSMakeRange(start, length);
-        NSArray *subarray = [codePoints subarrayWithRange:range];
-        [commands addObject:[self dictionaryForSendKeysCommandWithCodePoints:subarray windowPane:windowPane]];
+    NSMutableArray *subarray = [NSMutableArray arrayWithCapacity:MIN(codePoints.count, maxLiteralCharacters)];
+
+    for (NSNumber *number in codePoints) {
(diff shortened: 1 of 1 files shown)


gnachman:
Thank you for sending this PR. I'm currently up to my neck in a huge change and won't have time to look at it for a few days, but I promise to come back to this as soon as I'm able.

gnachman:
Merged with changes as 4aeb3647e183749b1336bc694fa22df5cf1d5286

lonnywong:
Thanks very much. I learn a lot from this.

And I want to send `Base64` directly, how to code it?

Please support the standard base64 ( aka `+` `/` `=` ) at least.

My [trzsz](https://github.com/trzsz/trzsz) ( similar to lrzsz ) is counting on that. Since it is using `Base64`.

```c
- (BOOL) canSendAsLiteralCharacter:(NSNumber *)codePoint {
    const unichar c = codePoint.unsignedShortValue;
    return isascii(c) && isalnum(c);
}
```

https://en.wikipedia.org/wiki/Base64
Screen Shot 2022-02-13 at 11 01 35 AM

gnachman:
I can't add `-` because tmux will confuse it with an option to `send-keys`. Others added in commit abc777966

lonnywong:
> I can't add `-` because tmux will confuse it with an option to `send-keys`. Others added in commit [abc7779](https://github.com/gnachman/iTerm2/commit/abc77796656c7ae489731dc532d9ca65887d34b9)

Thanks @gnachman . `+`  `/`  `:` is a big help.

`-` is not a big deal for me.

Why there is `c == ')'` ?

gnachman:
> > I can't add `-` because tmux will confuse it with an option to `send-keys`. Others added in commit [abc7779](https://github.com/gnachman/iTerm2/commit/abc77796656c7ae489731dc532d9ca65887d34b9)
> 
> Thanks @gnachman . `+` `/` `:` is a big help.
> 
> `-` is not a big deal for me.
> 
> Why there is `c == ')'` ?

I misread your original request, but I think it's harmless to have it there.

PR #467, opened 2022-04-02
Document how to join font name and size in the Python API

fxn:
By reading the docs, it was not clear how to pass the size to `set_normal_font()` and friends. I've done a pass specifying that they have to be separated by a space.

Took the opportunity to normalize "font's name" vs "font name", since you could find both. I believe in English "font name" would be the one, so picked that one. Please let me know if you prefer "font's name" instead.

### DIFF
--- a/api/library/python/iterm2/iterm2/profile.py
+++ b/api/library/python/iterm2/iterm2/profile.py
@@ -1916,8 +1916,8 @@ class LocalWriteOnlyProfile:
         Sets the normal font.
 
         The normal font is used for either ASCII or all characters depending on
-        whether a separate font is used for non-ascii. The value is a font's
-        name and size as a string.
+        whether a separate font is used for non-ascii. The value is a string with
+        the font name and size separated by a space.
 
         .. seealso::
           * Example ":ref:`increase_font_size_example`"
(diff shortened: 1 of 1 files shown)


gnachman:
This is a good idea, but you shouldn't modify `profile.py` directly because it is generated code. Instead, change `api/library/python/iterm2/gen_profile.py/gen_profile.py` and run `cd api/library/python/iterm2/gen_profile.py; make iterm2/profile.py`.

fxn:
@gnachman oh, awesome! I'll do that.

I was thinking also about detailing what is a font name exactly, because I had to try a bit for some. For example, for "DejaVu Sans Mono", you need to pass "DejaVuSansMono", but for "Noto Sans Mono" you have to pass "NotoSansMono-Regular" (its PostScript name).

Is there a way to pass the font name as it is? If not, how can we describe to the user what does the API understands as "font name"?

gnachman:
> @gnachman oh, awesome! I'll do that.
> 
> I was thinking also about detailing what is a font name exactly, because I had to try a bit for some. For example, for "DejaVu Sans Mono", you need to pass "DejaVuSansMono", but for "Noto Sans Mono" you have to pass "NotoSansMono-Regular" (its PostScript name).
> 
> Is there a way to pass the font name as it is? If not, how can we describe to the user what does the API understands as "font name"?

It's from the "PostScript language code", whatever that is :). I'm following Apple's API here: https://developer.apple.com/documentation/appkit/nsfont/1526183-fontname?language=objc

PR #468, opened 2022-04-02
Documents how to clear a background image

fxn:
I have documented this by trial and error, not sure it is correct.

Thought you should pass `None` to clear, but an empty string is what seems to work.

Have also made the description match type hinting.

### DIFF
--- a/api/library/python/iterm2/iterm2/profile.py
+++ b/api/library/python/iterm2/iterm2/profile.py
@@ -1939,9 +1939,10 @@ class LocalWriteOnlyProfile:
 
     def set_background_image_location(self, value: str):
         """
-        Sets the path to the background image.
+        Sets or clears the path to a background image.
 
-        The value is a Path.
+        The value is a string with the path to the image file,
+        or an empty string to clear the setting.
 
         :param value: A str
         """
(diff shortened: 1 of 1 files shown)


gnachman (approved):

gnachman:
Thank you!

fxn:
BTW, let me share my use case, maybe you find it interesting!

I have a little utility to do presentations in iTerm. Have been presenting with my little program for years. Slides live in one Ruby file, they are pure text, no Markdown. Bold face, cursor positioning, etc., is done with ANSI escape sequences. Source code highlighting via Pygments. (Early prototype [here](https://github.com/fxn/tkn), I do not maintain that repo).

I configure background images to include images in the presentation. So neat! Works like a charm. Clearing the background image is used to transition to the next slide. (Tried `imgcat`, but like this trick better.)

I did this with AppleScript time back, but now this Python API is really cool. Thanks for working on this API ❤️.

gnachman:
Cool applicaiton! Glad you found the API useful :)

fxn:
Followup: I opened RailsConf this year. The talk has so much images displayed with total normality thanks to your API, that it is not even apparent it is running on iTerm ([video](https://www.youtube.com/watch?v=DzyGdOd_6-Y&list=PLbHJudTY1K0f1WgIbKCc0_M-XMraWwCmk)).

I thought you might like to know :).

PR #469, opened 2022-04-11
[python API] Change the open/save panels

lonnywong:
1. Open the last directory by default.
2. Remove the exception if the user cancels.

### DIFF
--- a/api/library/python/iterm2/iterm2/filepanel.py
+++ b/api/library/python/iterm2/iterm2/filepanel.py
@@ -105,7 +105,9 @@ class OpenPanel:
              f'extensions: {json.dumps(self.extensions)},' +
              f'prompt: {json.dumps(self.prompt)},' +
              f'message: {json.dumps(self.message)})'))
-        return OpenPanel.Result(response) if response else None
+        if response:
+            return OpenPanel.Result(response)
+        return None
 
 
 class SavePanel:
@@ -242,4 +244,6 @@ class SavePanel:
                 f'message: {json.dumps(self.message)},' +
                 f'name_field_label: {json.dumps(self.name_field_label)},'
                 f'default_filename: {json.dumps(self.default_filename)})'))
-        return SavePanel.Result(response) if response else None
+        if response:
+            return SavePanel.Result(response)
+        return None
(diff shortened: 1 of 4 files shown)


gnachman:
That being said, what about returning None instead of a Result if the user cancels?

lonnywong:
> I think the exception was useful:
> 
> * We can guarantee that the list is never empty
> * If people forget to handle it, the breakage happens immediately

1. The developer should know that the user may cancels.
2. If catch the exception, maybe a real exception happens, instead of the user cancels. It's hard to identify.

lonnywong:
> That being said, what about returning None instead of a Result if the user cancels?

Return None is better then exception.

api/library/python/iterm2/iterm2/filepanel.py
```diff
@@ -89,13 +89,12 @@ def message(self) -> typing.Optional[str]:
     @message.setter
     def message(self, value: str):
         self.__message = value
-        
+
     async def async_run(self, connection) -> 'iterm2.OpenPanel.Result':
```
gnachman, line 93: The return type is now `typing.Optional['iterm2.OpenPanel.Result']`

Similar comment for save panel

api/library/python/iterm2/iterm2/filepanel.py
```diff
@@ -106,7 +105,9 @@ async def async_run(self, connection) -> 'iterm2.OpenPanel.Result':
              f'extensions: {json.dumps(self.extensions)},' +
              f'prompt: {json.dumps(self.prompt)},' +
              f'message: {json.dumps(self.message)})'))
-        return OpenPanel.Result(response)
+        if response:
```
gnachman, line 108: fyi I tweaked this and you'll need to pull to get it

api/library/python/iterm2/iterm2/filepanel.py
lonnywong: > The return type is now `typing.Optional['iterm2.OpenPanel.Result']`
> 
> Similar comment for save panel

done

lonnywong:
@gnachman 
Is the PR ready to merge? Thanks for your time.

PR #470, opened 2022-05-02
Add Apple Dark and Apple Light color theme

SebDanielsson:
New theme that I did since I couldn't find something that match Apple Human Interaction Interface color scheme. Colors from https://developer.apple.com/design/human-interface-guidelines/ios/visual-design/color/#system-colors

They match what Apple uses for their UI for dark and lightmode respectively. I made this theme for warp terminal as well and pictures of the color schemes can be found here:
Apple Dark
Apple Light

### DIFF
--- a/plists/ColorPresets.plist
+++ b/plists/ColorPresets.plist
@@ -4683,5 +4683,739 @@
 		<real>0.70196080207824707</real>
 	</dict>
     </dict>
+	<key>Apple Dark</key>
+	<dict>
+	<key>Ansi 0 Color</key>
+	<dict>
+		<key>Alpha Component</key>
+		<real>1</real>
+		<key>Blue Component</key>
+		<real>0.11764705926179886</real>
+		<key>Color Space</key>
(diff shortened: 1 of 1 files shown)


gnachman:
Thanks for taking this on! Since color schemes now have built-in support for light and dark mode I think it'd be better if this was a single color scheme.  I'd name it something like `Apple HIG`. The individual keys would be like `Ansi 0 Color (Dark)` and `Ansi 0 Color (Light)`. No need for the naked `Ansi 0 Color` — that's only for backward compatibility with older versions of iTerm2.

SebDanielsson:
Thanks for the tip. Will try to get this fixed tomorrow.

So I do the same with the background color as well? How do I change between the dark and light mode in the GUI? Will the same theme automatically get two entries in the theme list?

SebDanielsson:
ColorPresets.plist is a real pain to work on. If anyone else want this theme feel free to create a new PR😉

PR #472, opened 2022-05-28
Add escape code for rich notifications.

gawashburn:
This pull request adds support for a "rich" notification escape code.
There might a better choice than calling them "rich", but the feature
allows more control over notifications than the existing escape code.

I had considered whether it might be better to try to implement this
functionality via the "custom" escape code and Python API, and then
using another program such as terminal-notifier
(https://github.com/julienXX/terminal-notifier) to present the
notifications.  However, it does not seem possible to get an external
tool to focus the correct window/tab when clicking on the notification.

To keep parsing simple, each of the fields for the notification is
base64 encoded.

It might make sense to lock this down further with add an allow/deny
list for the notification image, similar to what there appears to be for
setting a background image.  However, I wasn't able to determine where
those lists were defined.

It also appears that there the code is using an older deprecated
notifications API.  It seems better to address that in a separate change
though.  However, it looks the new API was introduced in Mojave, so I'm
not sure if there backwards compatibility implications.

Finally, it has also been quite a long time since I have written much
Objective-C, so definitely let me know if there are any improvements to
style that I should make.

### DIFF
--- a/sources/PTYSession.m
+++ b/sources/PTYSession.m
@@ -13343,23 +13343,92 @@ scrollToFirstResult:(BOOL)scrollToFirstResult {
     [self screenNeedsRedraw];
 }
 
-- (void)screenPostUserNotification:(NSString * _Nonnull)message {
+- (void)screenPostUserNotification:(NSString * _Nonnull)message rich:(BOOL)rich {
     if (![self shouldPostTerminalGeneratedAlert]) {
         DLog(@"Declining to allow terminal to post user notification %@", message);
         return;
     }
     DLog(@"Terminal posting user notification %@", message);
     [self incrementBadge];
-    NSString *description = [NSString stringWithFormat:@"Session %@ #%d: %@",
-                             [[self name] removingHTMLFromTabTitleIfNeeded],
-                             [_delegate tabNumber],
-                             message];
-    [[iTermNotificationController sharedInstance]
-     notify:@"Alert"
-     withDescription:description
-     windowIndex:[self screenWindowIndex]
-     tabIndex:[self screenTabIndex]
-     viewIndex:[self screenViewIndex]];
+
+    iTermNotificationController* controller = [iTermNotificationController sharedInstance];
+
+    if (rich) {
+        NSString *title = nil;
+        NSString *subtitle = nil;
+        NSString *description = nil;
+        NSString *image = nil;
+
+        // Parse the notification payload.
+        NSString* remainingText = message;
+        // Continue parsing until we run out of text.
+        while (remainingText.length > 0) {
+            // First search for the equals token dividing a key from value.
+            NSRange eqRange = [remainingText rangeOfString:@"="];
+            if (eqRange.location != NSNotFound) {
+                NSString *key = [remainingText substringToIndex:eqRange.location];
+                remainingText = [remainingText substringFromIndex:eqRange.location+1];
+                // Next search to see if there is a termination token indicating additional
+                // key-value pairs.
+                NSRange termRange = [remainingText rangeOfString:@";"];
+                NSString *value;
+                if (termRange.location != NSNotFound) {
+                    value = [remainingText substringToIndex:termRange.location];
+                    remainingText = [remainingText substringFromIndex:termRange.location+1];
+                } else {
+                    value = remainingText;
+                    remainingText = @"";
+                }
+
+                // Decode the value from base64.  If the data cannot be decoded continue to next key-value pair.
+                NSString *decodedValue = [value stringByBase64DecodingStringWithEncoding:NSUTF8StringEncoding];
+                if (decodedValue == nil) {
+                    DLog(@"Rich notification value not valid base64 data: %@", value);
+                    continue;
+                }
+
+                // Handle they key and value.
+                if ([key isEqualToString:@"message"]) {
+                    description = decodedValue;
+                } else if ([key isEqualToString:@"title"]) {
+                    title = decodedValue;
+                } else if ([key isEqualToString:@"subtitle"]) {
+                    subtitle = decodedValue;
+                } else if ([key isEqualToString:@"image"]) {
+                    image = [decodedValue stringByTrimmingTrailingCharactersFromCharacterSet:[NSCharacterSet whitespaceAndNewlineCharacterSet]];
+                } else {
+                    DLog(@"Unexpected key-value pair while parsing rich notification payload: %@=%@", key, value);
+                }
+            // No token, so exit.
+            } else {
+                DLog(@"No '=' token found while parsing rich notification payload: %@", remainingText);
+                break;
+            }
+        }
+
+        // description must not be nil.  If it is the empty string, the title will be used.
+        if (description == nil) {
+            description = @"";
+        }
+
+        [controller notifyRich:title
+                  withSubtitle:subtitle
+               withDescription:description
+                     withImage:image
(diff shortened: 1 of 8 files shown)


gnachman:
Thank you for taking this on and I'm sorry it took so long for me to get to it. The only concern I have is about the image. We have to treat input as malicious (since when you ssh somewhere or git push, a remote computer controls the input you receive). We also have to treat image decoding a remote code execution. Since the image has to be local it's less of a concern, but this could be one step in an exploit chain.

To make image decoding safe, use iTermImage instead of creating the NSImage directly:

```
NSData *data = [NSData dataWithContentsOfFile:image];
notification.contentImage =  = [[[iTermImage imageWithCompressedData:data] images] firstObject];
```

gawashburn:
No problem, I've had plenty to keep me busy in the meantime :-)

I've updated the pull request as suggested.  Let me know if you think of anything else.

sources/iTermNotificationController.m
```diff
@@ -84,6 +85,53 @@ - (BOOL)notify:(NSString *)title
...
+    }
+    if (subtitle != nil) {
+        notification.subtitle = subtitle;
+    }
+    if (image != nil) {
+        NSData *data = [NSData dataWithContentsOfFile:image];
+        notification.contentImage = [[[iTermImage imageWithCompressedData:data] images] firstObject];
+    }
+    if (sound != nil) {
+        notification.soundName = [sound isEqualToString:@"default"] ? NSUserNotificationDefaultSoundName : sound;
```
gnachman, line 127: How is sound supposed to work? I see in NSUserNotification.h that it's supposed to be the name of a sound in the app bundle:

```
// The name of the sound file in the resources of the application bundle to play when the notification is delivered. NSUserNotificationDefaultSoundName can be used to play the default Notification Center sound. A value of 'nil' means no sound.
@property (nullable, copy) NSString *soundName;
```

But there are no sounds in the app bundle. I'd be fine with adding a few if they're not too big and would be generally useful, but otherwise we should remove the sound field since it won't work.

sources/PTYSession.m
```diff
@@ -13343,23 +13343,84 @@ - (void)screenDidUpdateReturnCodeForMark:(id<VT100ScreenMarkReading>)mark
...
+        // Description
+        NSString* part = decoded_parts[2];
+        if ([part length] > 0) {
+            description = part;
+        } else {
+            DLog(@"Rich notification requires a non-empty description field.");
+        }
+
+        // Title
+        part = decoded_parts[0];
```
gnachman, line 13391: I recommend using key-value pairs rather than assuming an order. It'll make it easier to modify this in the future. Like this:

```OSC 1337 ; Notification=title=[base64] ; subtitle=[base64] ; image=[base64] ST```

File: sources/PTYSession.m [hunk 3]
```diff
@@ -13343,23 +13343,84 @@ - (void)screenDidUpdateReturnCodeForMark:(id<VT100ScreenMarkReading>)mark
...
+            return;
+        }
+
+        NSString *title = nil;
+        NSString *subtitle = nil;
+        NSString *description = nil;
+        NSString *image = nil;
+        NSString *sound = nil;
+
+        NSMutableArray *decoded_parts = [[NSMutableArray alloc] init];
```
gnachman, line 13371: Avoid using _ in variable names, preferring camelCase. Also this file is manual retain-release so this is a memory leak. This would be correct:

```
NSMutableArray<NSString *> *decodedParts = [NSMutableArray];
```

sources/PTYSession.m
```diff
@@ -13343,23 +13343,84 @@ - (void)screenDidUpdateReturnCodeForMark:(id<VT100ScreenMarkReading>)mark
...
+        // Decode all the arguments
+        for (NSString* part in parts) {
```
gnachman, line 13373: Keep the `*` next to the variable name, not the type. So `NSString *part`. Same on line 13374.

sources/PTYSession.m
```diff
@@ -13343,23 +13343,84 @@ - (void)screenDidUpdateReturnCodeForMark:(id<VT100ScreenMarkReading>)mark
...
+        NSString* part = decoded_parts[2];
+        if ([part length] > 0) {
+            description = part;
+        } else {
+            DLog(@"Rich notification requires a non-empty description field.");
+        }
+
+        // Title
+        part = decoded_parts[0];
+        if ([part length] > 0) { title = part; }
```
gnachman, line 13392: if bodies need to go on the next line so you can set breakpoints on them. Same on line 13395 and 13403.

sources/iTermNotificationController.m
```diff
@@ -84,6 +85,53 @@ - (BOOL)notify:(NSString *)title
...
+    }
+
+    // Send to NSUserDefaults directly
+    NSUserNotification *notification = [[NSUserNotification alloc] init];
+    notification.informativeText = description;
+
+    NSDictionary *context = nil;
+    if (windowIndex >= 0) {
+        context = @{ @"win": @(windowIndex),
+                @"tab": @(tabIndex),
```
gnachman, line 111: Inconsistent indentation. I suggest moving `@"win":…` onto the next line and then xcode will do the right thing.

sources/iTermNotificationController.m
```diff
@@ -84,6 +85,53 @@ - (BOOL)notify:(NSString *)title
...
+
+    if (title != nil) {
+        notification.title = title;
+    }
+    if (subtitle != nil) {
+        notification.subtitle = subtitle;
+    }
+    if (image != nil) {
+        NSData *data = [NSData dataWithContentsOfFile:image];
+        notification.contentImage = [[[iTermImage imageWithCompressedData:data] images] firstObject];
```
gnachman, line 124: What if `data` is nil?

gnachman (changes requested):

sources/iTermNotificationController.m
gawashburn: Ah, I had misunderstood the documentation.  I had assumed it would search system sounds as well.  But if the sound must be bundled with the application, it seems like the `NSUserNotification` API is not general enough to be worth exposing here.

sources/PTYSession.m
gawashburn: I was trying to keep the code simpler by avoiding the need to do more complex parsing.  I've rewritten it to parse key-value pairs and it proved to not be that much more code and effort.

sources/PTYSession.m
gawashburn: Thanks for pointing out the memory leak here!  In any event, this code no longer exists in my revised version.

sources/PTYSession.m
gawashburn: I think I've corrected any occurrences of this in my revised version.

sources/PTYSession.m
gawashburn: That code is no longer present, but I think I've avoided this in my revised version.

sources/iTermNotificationController.m
gawashburn: Fixed.

sources/iTermNotificationController.m
gawashburn: Ah, I had overlooked that.  It would appear `iTermImage` will handle the `nil` case correctly, but I've added a conditional to avoid the call altogether.

gnachman:
Merged as 8ef95e1b0d9c849a68d76500e91901e97b27b635. Thank you!

PR #473, opened 2022-06-20
Attempt to disclaim ownership of child processes. 

gnachman:
Doesn't work because they are just denied permission for everything.

### DIFF
--- a/sources/iTermFileDescriptorMultiServer.c
+++ b/sources/iTermFileDescriptorMultiServer.c
@@ -186,6 +186,31 @@ static void RemoveChild(int i) {
 
 #pragma mark - Launch
 
+static int CountArgs(const char **argv) {
+    int i = 0;
+    while (argv[i] != NULL) {
+        i += 1;
+    }
+    return i;
+}
+
+static char **DuplicateNullTerminatedArrayOfCStrings(const char **inputArray) {
+    const int count = CountArgs(inputArray);
+    char **outputArray = malloc(sizeof(char *) * (count + 1));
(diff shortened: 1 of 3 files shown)


rolfbjarne:
What does `log stream` show about the process that was denied permission?

gnachman:
Can you remind me what this was all about?

rolfbjarne:
> Can you remind me what this was all about?

https://gitlab.com/gnachman/iterm2/-/issues/10360

gnachman:
I stared at it for a while but not seeing anything useful here, but maybe you will. It'll be before this line
```
2023-08-11 11:54:04.638662-0700 0x27299f   Default     0x5bc32f             76509  0    bluetoothapp: CBManagerStateUnauthorized
```
[log.txt.zip](https://github.com/gnachman/iTerm2/files/12324394/log.txt.zip)

rolfbjarne:
I just built iTerm2 locally with this patch and it works for me - in that I get the same (somewhat unintuitive) behavior as Terminal.

That is, I get the following dialog when an executable asks for permission (in this case to access the Music library):

Screenshot 2024-02-20 at 11 48 16

which is the exact same dialog I get with Terminal:

Screenshot 2024-02-20 at 11 47 38

Note that the previous behavior was for the executable to crash with:

> This app has crashed because it attempted to access privacy-sensitive data without a usage description.

gnachman:
Merged in 1d3e09f225ec5a15e1aeb341da28ad6397c831d9. It's off by default for now but I expect to turn it on after 3.5 is release. See the commit message for details.

PR #475, opened 2022-07-07
Fix 1password retrieval via `op` when more than one password in the record

hughescr:
1Password items can have more than one field with the label `"password"`; in this case, the `op` command seems to pick one arbitrarily and return it; there is no clean way that I could find to get `op` to return the one which has `id` set to `"password"`, which is the primary password field on the item.  Instead, I return all the fields, and then find the one with `id` set to `"password"` and return that.

Fixes https://gitlab.com/gnachman/iterm2/-/issues/10488


gnachman:
Merged as 8b74fcc7c242e1095bf1dc090c116ac0c3d59cd3, thank you!

PR #477, opened 2022-08-26
Fix View menu offset on macOS Big Sur and later

dnicolson:
This pull request fixes the alignment of the `ColorsMenuItemView` view when there are no menu items with an `on` state in the View menu on macOS Big Sur and later.

Before and after:
Before After

The `viewMenuHasSubmenuOnState` method is perhaps too coupled to this class so it may be better placed elsewhere.


gnachman:
Dang, I didn't know about this! Good catch. Merged with tweaks as fd2d42416ff8d0be19fb12b9673e70a51fa51ad4

PR #479, opened 2022-10-11
Small additions to settabcolor example and connection.py documentation

brewingcode:
In Python, I was naively passing the `ITERM_SESSION_ID` envvar into `app.get_session_by_id()` but not getting any sessions back. I then realized the env var is prefixed a window+tab+pane identifier which is not part of the session ID. I wasn't sure if this prefix always exists, so the example code regexes for the session id in a more flexible way than simply `.split(':')`.

I didn't see this mentioned anywhere, and couldn't find a great place to add it, so I just added it to the code sample on the example that I was basing my work on in the first place. I was hitting this because I was appending a tab color change to the end of long-running scripts, and the tab change would get applied to my CURRENT tab instead of the previous tab where I had left the long-running script running.

I also found the `connection.py:async_create()` documentation without finding the `running.rst:REPL` section, so a link from `async_create` to that doc section would have been helpful. :-)

### DIFF
--- a/api/library/python/iterm2/docs/examples/settabcolor.rst
+++ b/api/library/python/iterm2/docs/examples/settabcolor.rst
@@ -7,8 +7,6 @@ Set Tab Color
 
 This script sets the tab color for the current session to a hard-coded value. It also turns on the use of tab color for that session. It does not modify the underlying profile, so only the current session is affected.
 
-It also shows how to parse the ITERM_SESSION_ID environment variable to set the tab color for the tab that originally ran this script, instead of the currently active tab.
-
 .. code-block:: python
 
     #!/usr/bin/env python3.7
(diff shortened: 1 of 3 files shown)


gnachman:
Thanks for the submission! I don't think it makes sense to add SESSION_ID parsing to the example code since there are many examples where it'd make sense to put it. Furthermore, it wouldn't work when launched not from the command line. But the update to the docs is excellent and worth keeping.

brewingcode:
> I don't think it makes sense to add SESSION_ID parsing to the example code

Agreed. I just wanted to put it somewhere, because I couldn't find mention anywhere of the structure of `SESSION_ID` being `<prefix>:<actual session ID>`. Having it in a PR is close enough for googling purposes, though. :-)

Thanks for merging, and for iTerm!

PR #480, opened 2022-11-19
Automatically set Alert on Next Mark on all unfocused sessions

tiborvass:
I love the "Alert on Next Mark" feature so much so that I became annoyed to have to constantly enable it. I realized I always want to be notified when any session other than the active one, has yielded. Maybe it is useful for others too hence the PR.

Not sure if this could be accepted upstream in some shape or form (gated by a setting?).

And thank you for this amazing project!

### DIFF
--- a/sources/PTYSession.m
+++ b/sources/PTYSession.m
@@ -6823,6 +6823,7 @@ scrollToFirstResult:(BOOL)scrollToFirstResult {
         DLog(@"Is tmux gateway");
         return;
     }
+    [self setAlertOnNextMark:!focused];
     _focused = focused;
     if (_screen.terminalReportFocus) {
         DLog(@"Will report focus");


gnachman:
I think it's too big a change for the default behavior. But it might make sense to extend the Python API to make this doable through a script.

tiborvass:
@gnachman Would you be okay with making it a profile setting? Or do you want to add a special RPC for this boolean ?

tiborvass:
Thank you @gnachman ! Correct me if I'm wrong, but this is not exactly what I was suggesting in this ill-formed pull request: although I do want all sessions to have alert on next mark, I also don't want to have to do the keybinding every time. I think users would greatly benefit by setting the profile setting once. What do you think?

gnachman:
1f4a0dc1dc6f67e1f56da03770fea5396896467b adds a menu item `Add Alert on Marks in Offscreen Sessions`, which is persistent.

PR #482, opened 2023-01-13
Change minimum value of blending level from 5 to 0

theoctober19th:
This fixes the bug of not being able to enter values like 10, 20, etc. when setting the blending level from the textbox alongside the slider in Preferences -> Window tab.

Fixes Issue: https://gitlab.com/gnachman/iterm2/-/issues/10772


gnachman:
Thank you very much!

PR #483, opened 2023-01-15
Fix XCode 12 compile error

mklein-de:
```
error: property 'safeAreaInsets' not found on object of type 'NSScreen *'
```


gnachman:
Sorry, I don't try to maintain support for older versions of xcode. It makes a real mess out of the codebase to have all the conditionals.

PR #481, opened 2023-01-07
RFC: Report extra mouse buttons, issue 10689

aymanbagabas:
This is a POC to add support for extra mouse buttons support. Things still need to be done:

* report multiple scroll left & right events
* handle scroll left & right switching to different tabs collision
* add button 10 & 11

Fixes: https://gitlab.com/gnachman/iterm2/-/issues/10689

PR #484, opened 2023-02-04
Report mouse button 10 & 11

aymanbagabas:
Handle pty mouse buttons 10 and 11

Following up on https://github.com/gnachman/iTerm2/pull/481
Fixes: https://gitlab.com/gnachman/iterm2/-/issues/10689

### DIFF
--- a/sources/PTYMouseHandler.m
+++ b/sources/PTYMouseHandler.m
@@ -1271,6 +1271,12 @@ static double EuclideanDistance(NSPoint p1, NSPoint p2) {
                     return MOUSE_BUTTON_BACKWARD;
                 case 4:
                     return MOUSE_BUTTON_FORWARD;
+                case 5:
+                    return MOUSE_BUTTON_10;
+                case 6:
+                    return MOUSE_BUTTON_11;
+                default:
+                    return MOUSE_BUTTON_UNKNOWN;
             }
 
 
--- a/sources/VT100Output.h
+++ b/sources/VT100Output.h
@@ -2,6 +2,7 @@
 #import "VT100GridTypes.h"
 
 typedef enum {
+    MOUSE_BUTTON_UNKNOWN = -1,   // unknown button
     // X11 button number
     MOUSE_BUTTON_LEFT = 0,       // left button
     MOUSE_BUTTON_MIDDLE = 1,     // middle button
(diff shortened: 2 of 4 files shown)


aymanbagabas:
The other commit changes the `updateVersion.py` tool to use `/usr/bin/env` to look for `python3`. I believe `/opt/homebrew/bin/python3` is only valid under M1 homebrew.

sources/PTYMouseHandler.m
```diff
@@ -1271,6 +1271,12 @@ - (MouseButtonNumber)mouseReportingButtonNumberForEvent:(NSEvent *)event {
                     return MOUSE_BUTTON_BACKWARD;
                 case 4:
                     return MOUSE_BUTTON_FORWARD;
+                case 5:
+                    return MOUSE_BUTTON_10;
+                case 6:
+                    return MOUSE_BUTTON_11;
+                default:
+                    return MOUSE_BUTTON_UNKNOWN;
```
gnachman, line 1279: Maybe `MOUSE_BUTTON_NONE` would be better here because then we could still report movement while an unrecognized button is pressed.

sources/VT100Output.h
```diff
@@ -13,6 +14,8 @@ typedef enum {
     MOUSE_BUTTON_SCROLLRIGHT = 7,// scroll right
     MOUSE_BUTTON_BACKWARD = 8,   // backward (4th button)
     MOUSE_BUTTON_FORWARD = 9,    // forward (5th button)
+    MOUSE_BUTTON_10,             // extra button 1
+    MOUSE_BUTTON_11,             // extra button 2
```
gnachman, line 18: Please add explicit values (` = 10`, ` = 11`) for these. It's a signal to future maintainers that the value of the enum is important and shouldn't be changed.

gnachman (changes requested):

sources/PTYMouseHandler.m
aymanbagabas: The [xterm specs](https://invisible-island.net/xterm/ctlseqs/ctlseqs.html#h3-Extended-coordinates:~:text=the%0AalternateScroll%20resource.-,Other%20buttons,-Some%20wheel%20mice) don't support buttons beyond 11 :/
`MOUSE_BUTTON_NONE` would create confusion for applications listening to it since it corresponds to button release in (1000) mode. I think unrecognized buttons should be ignored.

gnachman:
Thank you!

PR #485, opened 2023-08-11
Python API: Making *Knob classes subclasses of Knob and moving Knob over to BaseKnob

djpadz:
This will allow static type checkers to not complain when it runs across code like this:

```python
    python_mycomponent = StatusBarComponent(
        ...
        knobs=[
            CheckboxKnob("Debug", False, "mycomponent_debug"),
        ],
    )
```


gnachman:
Thank you very much!

PR #488, opened 2023-10-29
Optimize PromptStateMachine.handle(token:)

saagarjha:
Some code used for debugging was actually quite expensive, so reduce it a bit when not relevant.


gnachman:
Thanks!

PR #489, opened 2023-11-22
Only restore previously active app if iTerm is active

mallexxx:
Fixes a bug when iTerm restores previously active app when another app is already active:
- Activate some application (e.g. `TextEdit`) then open another app window in front of it (e.g. `Calculator`)
- Open iTerm hotkey window
- With iTerm hotkey window open click a first activated app (`TextEdit`)
- iTerm restores previously open app (`Calculator`) in place of just activated `TextEdit`

Same bug can be reproduced when running an `open` command from a hotkey window: e.g. running `open .` opens Finder window and then reactivates another app that was open previously.

### DIFF
--- a/sources/iTermPreviousState.m
+++ b/sources/iTermPreviousState.m
@@ -85,9 +85,9 @@
 
     NSRunningApplication *app = [self appToSwitchBackToIfAllowed];
     BOOL result = NO;
-    if (app) {
+    if (app && [NSApp isActive]) {
         DLog(@"Restore app %@", app);
-        DLog(@"** Restor previously active app from\n%@", [NSThread callStackSymbols]);
+        DLog(@"** Restore previously active app from\n%@", [NSThread callStackSymbols]);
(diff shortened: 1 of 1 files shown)


sources/iTermPreviousState.m
```diff
@@ -85,9 +85,9 @@ - (BOOL)restorePreviouslyActiveApp {
 
     NSRunningApplication *app = [self appToSwitchBackToIfAllowed];
     BOOL result = NO;
-    if (app) {
+    if (app && [NSApp isActive]) {
```
mallexxx, line 88: ensure iTerm app is still active before restoring a previously active app

gnachman:
I suspect that commit 9c5be76be0a4fb09f55640c277121ca71d9f468d fixes the issue in a different way. Can you check if this is still needed?

mallexxx:
@gnachman, yes, that worked, thanks!

PR #490, opened 2023-11-22
Support opening to line number in Xcode

mallexxx:
Adds support opening to a file:line in Xcode IDE (using `xed` command)


gnachman:
Thank you very much!

PR #491, opened 2023-11-25
Add forgotten VSCode Insiders ⌘-click open handling

jevinskie:
The bundle identifiers for both VSCode and VSCode Insiders (beta builds) were already present but only the VSCode non-Insiders was used in the ⌘-click open logic. This adds the Insiders bundle ID to be handled the same way as standard VSCode.

Fixes [issue 11236](https://gitlab.com/gnachman/iterm2/-/issues/11236).


gnachman:
Thanks!

PR #494, opened 2024-01-13
Add dark appearance support for AI setup image

dnicolson:
This adds support for a dark appearance image, there's probably a better way of achieving this but there didn't seem to be assets available to add to.

Before and after:


gnachman:
FYI a simpler solution can be found in commit 364b78e3aeca1fab794a328acccdea7ea1fd2584

PR #496, opened 2024-02-08
Optionally disable Cmd-Click opening Command Info

robbieg8s:
This adds an advanced setting which defaults to the current behaviour, but when switched off, disables the feature that Cmd-Click in a prompt opens the Command Info window. The Cmd-Click instead falls back to Smart Selection. The window can still be accessed via right click as normal.

### DIFF
--- a/sources/PTYTextView.m
+++ b/sources/PTYTextView.m
@@ -5875,7 +5875,8 @@ scrollToFirstResult:(BOOL)scrollToFirstResult
 
 - (void)mouseHandlerOpenTargetWithEvent:(NSEvent *)event
                            inBackground:(BOOL)inBackground {
-    if ([self showCommandInfoForEvent:event]) {
+    if ([iTermAdvancedSettingsModel enableCmdClickPromptForShowCommandInfo]
+        && [self showCommandInfoForEvent:event]) {
         return;
     }
     [_urlActionHelper openTargetWithEvent:event inBackground:inBackground];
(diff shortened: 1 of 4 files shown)


sources/iTermURLActionFactory.m
```diff
@@ -309,6 +309,9 @@ - (void)trySecureCopy {
 #pragma mark - Sub-factories
 
 - (URLAction *)urlActionForPrompt {
+    if (![iTermAdvancedSettingsModel enableCmdClickPromptForShowCommandInfo]) {
+        return nil;
+    }
```
robbieg8s, line 314: For my use case this felt cleaner than reordering the phases in the factory (so that Smart Selection is found first), but I'm open to the other path also.

sources/PTYTextView.m
```diff
@@ -5875,7 +5875,8 @@ - (void)mouseHandlerDidInferScrollingIntent:(PTYMouseHandler *)handler trying:(B
 
 - (void)mouseHandlerOpenTargetWithEvent:(NSEvent *)event
                            inBackground:(BOOL)inBackground {
-    if ([self showCommandInfoForEvent:event]) {
+    if ([iTermAdvancedSettingsModel enableCmdClickPromptForShowCommandInfo]
+        && [self showCommandInfoForEvent:event]) {
```
robbieg8s, line 5879: I'm not sure this is quite the right place for this - maybe in `showCommandInfoForEvent` would be cleaner - but without something here, the click goes to the new feature even though the fix in `sources/iTermURLActionFactory.m` below results in the UX showing the Smart Selection as the highlight.

gnachman: I can't recall why this is here, but this is the right place for it because you don't want to disable showing the command info when you right click on the offscreen command line view.

gnachman (approved):

PR #497, opened 2024-02-19
Add advanced setting to exclude utun from network utilization

hguandl:
Possible implementation for [Issue 11369](https://gitlab.com/gnachman/iterm2/-/issues/11369).


gnachman (approved):

gnachman:
I think it's ok to enable this by default.

PR #499, opened 2024-04-22
Nit: use arrayWithCapacity initializer in fix 799afde

jevinskie:
I have no idea how hot this path is but I figure it can't hurt. Only spotted this because I was happy to see the Cmd-K issue resolved - thanks!

Forgive my editor for trimming trailing whitespace, if you don't want that.


gnachman (approved):

gnachman:
Thank you!

PR #500, opened 2024-05-03
Display correct window restoration instructions for macOS 14

ed-flanagan:
TL;DR: I believe the "Window Restoration" help window has the outdated text for macOS 14, but not macOS 13.

---

I don't have much (any) experience with Objective-C or Apple app dev, so apologies ahead of time if I'm way off base with anything:

I noticed on my Mac Book (macOS 14.3 (I know I need to update)) that the "Window Restoration Disabled" help window references `System Preferences > General > ...`. Which, based on conditional, should be `Desktop & Dock` for macOS `>= 13`.

I'm guessing https://github.com/gnachman/iTerm2/commit/f5617a1089541f1bdcb4edcbbfec54a492304188#diff-3a2c028b85d15b2ff41c606edba92e67d8e1fd252936a12ee185ed859005c809R779 was "reverted" since the message was incorrect for macOS 13, after the change to 14. 

So I'm further guessing `@available(macOS 13, *)` behaves more like `ver < 14 && ver >= 13.0.0` or `ver ~> 13.0` instead of just `ver >= 13.0.0` or `ver ~> 13`. 

So I just extended the conditional to include `macOS 14`, to see if that fixes it.

I haven't tried building it myself yet, so sorry for submitting before testing, but hoping to try in the next morning when I have a moment. But figured on the off-chance the small change is correct, can at least post.

Screenshot 2024-05-02 at 11 34 56 PM

Screenshot 2024-05-02 at 11 44 25 PM

### DIFF
--- a/sources/GeneralPreferencesViewController.m
+++ b/sources/GeneralPreferencesViewController.m
@@ -816,7 +816,7 @@ enum {
     NSString *message;
     NSString *action;
     NSString *path;
-    if (@available(macOS 13, *)) {
+    if (@available(macOS 13, *) || @available(macOS 14, *)) {
         message = @"System window restoration has been disabled, which prevents iTerm2 from respecting this setting. Disable ”System Settings > Desktop & Dock > Close windows when quitting an application“ to enable window restoration.";
         action = @"Open System Settings";
         path = @"/System/Library/PreferencePanes/Dock.prefPane";


gnachman:
The existing code handles any version of macOS greater than or equal to 13, so 14 is already included. Are you sure you were testing the latest version of the code when you saw the outdated message?

ashugg:
FWIW today I am sitting at a system with macOS 14.8.1 and iTerm2 3.6.5, and the "Window Restoration Disabled" warning message is certainly correct when that system setting is enabled.

ed-flanagan:
Apologies for never following up (haven't really been GH active since around then) - moot now.

PR #501, opened 2024-05-08
Fix horizontal scrolling in preferences

dnicolson:
This prevents the instances of `NSScrollView` under Preferences > Profiles (General, Terminal, Advanced) from being able to be horizontally elastically scrolled. This is achieved by subtracting the width of the scroll bar from the view.

This related code only prevents the animation:
`scrollView.horizontalScrollElasticity = NSScrollElasticityNone;`


gnachman:
This is definitely an improvement. It's broken if you change the scroller style while the settings window is open, though. Not sure it's worth the effort to fix that! Thanks :)

PR #504, opened 2024-07-23
Add py.typed file

chris-janidlo:
This ensures conformity with [PEP-561][1], which is [required for mypy type-checking to work][2].

[1]: https://peps.python.org/pep-0561/
[2]: https://mypy.readthedocs.io/en/stable/installed_packages.html#creating-pep-561-compatible-packages


gnachman:
Thanks!

PR #506, opened 2024-08-13
Default AI model to new gpt-4o-mini

fiddler:
Switch the default AI model to the new gpt-4o-mini, which is cheaper and better than the current default gpt-3.5-turbo.


gnachman:
Thanks!

PR #509, opened 2024-11-07
Fix async_set_broadcast_domains example

ajdavis:
The example was missing the required first arg, "connection".


gnachman:
Thanks!

PR #512, opened 2024-12-10
Fix memory leak in metalGetGlyphKeysData

charlievieth:
This commit fixes a leak in metalGetGlyphKeysData due to the `positions` vector, which is always allocated, only being freed when the `bidiInfo` argument is not null. The fix is to always free the `positions` vector.

https://gitlab.com/gnachman/iterm2/-/issues/12029


charlievieth:
I've been running iTerm2 with this change for the past 2 days and so far it has been stable and not leaking memory.

gnachman:
Thank you!

PR #513, opened 2025-01-09
Fixed powerline pdf glyph rendering

riarheos:
Hello.

Some time ago the builtin powerline glyph rendering was broken and the glyphs are rendered full-size even when the scaled option is set to NO.

I couldn't find the commit that actually broke the rendering but I could fix it so that the glyphs render correctly again.

The before-after image:
iterm-pr-demo

P.S. Thank you for the awesome terminal! :)

### DIFF
--- a/sources/iTermBoxDrawingBezierCurveFactory.m
+++ b/sources/iTermBoxDrawingBezierCurveFactory.m
@@ -555,7 +555,6 @@ offset:(CGPoint)offset {
                   color:(CGColorRef)color
             antialiased:(BOOL)antialiased 
                  offset:(CGPoint)offset {
-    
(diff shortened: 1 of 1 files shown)


riarheos:
Hmm, these extended symbols seem to be handled separately indeed. Added a `stretch:YES` to handle them (all the glyphs in the extended powerline directory need the stretch from what I can see).

So now it goes like this:
- we load and cache images with their original size (scaled down to twice the cell size)
- in `drawPDFWithName` we respect the `stretch` argument to actually stretch it

Does this look right for you?

изображение

gnachman:
Merged. Thanks!

PR #517, opened 2025-05-26
Fix maximizing window on title bar double click when system-wide action is set to "Fill"

fnesveda:
In macOS Sequoia, there’s a new feature that allows users to double-click a window's title bar to make the window fill the screen (rather than just "zoom" it as in older macOS versions).
Screenshot 2025-05-26 at 16 22 12

However, when this option is selected, the double-click title bar to maximize feature in iTerm2 stops working. This PR fixes that by checking whether the `AppleActionOnDoubleClick` system-wide user default maybe contains the new "Fill" option when deciding what to do after a title bar double-click.

### DIFF
--- a/sources/NSWindow+iTerm.m
+++ b/sources/NSWindow+iTerm.m
@@ -24,7 +24,7 @@ void *const iTermDeclineFirstResponderAssociatedObjectKey = (void *)"iTermDeclin
         [self performMiniaturize:nil];
         return;
     }
-    if (doubleClickAction == nil || [doubleClickAction isEqualToString:@"Maximize"]) {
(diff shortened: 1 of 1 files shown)


gnachman:
Thanks for the PR. I'm going to try to hew closer to the OS's own behavior in this case by calling a private API (commit 2e4690623c379e47aab0b18828a76b4ca52bf76e)

gnachman:
Thanks for the PR. I'm going to try to hew closer to the OS's own behavior in this case by calling a private API (commit 2e4690623c379e47aab0b18828a76b4ca52bf76e)

PR #518, opened 2025-06-10
fix parse error while using third-party llm api

Ylarod:
While using qwen2.5-max

```
Failed to decode API response: keyNotFound(CodingKeys(stringValue: "id", intValue: nil), Swift.DecodingError.Context(codingPath: [], debugDescription: "No value associated with key CodingKeys(stringValue: \"id\", intValue: nil) (\"id\").", underlyingError: nil)). Data is: {"choices":[{"message":{"role":"assistant","content":"echo Hello"},"index":0,"finish_reason":"stop"}],"usage":{"completion_tokens":6,"prompt_tokens":48,"total_tokens":54,"cache_creation_input_tokens":null,"cache_read_input_tokens":null}}
```


gnachman:
Done in 6f69aef6ca027be65e6f7b08b42fc10a009e3889.

PR #521, opened 2025-09-11
Add "Move Tab to New Window" menu item

cbowns:
Adds a menu item to Window → Tab → Move Tab to move a tab to a new window. It calls the existing selector in PseudoTerminal that's used by the tab title contextual menu, but now that it's in the menu bar, there's a keyboard shortcut for it!

This action is similar to existing tab management menu items in Safari, Chrome, and others. (I also added a default shortcut that I *think* is commonly assigned by them, and didn't have any base conflicts in the default iTerm shortcuts that I could find?)

Also extended `validateMenuItem` to disable the item unless there's more than one tab in the active window.

(Been a minute since I mucked around on macOS! Let me know if there's anything I'm missing.)

## UI

Before | After
--- | ---
before | after

### DIFF
--- a/Interfaces/MainMenu.xib
+++ b/Interfaces/MainMenu.xib
@@ -1579,6 +1579,12 @@ DQ
                                                             <action selector="moveTabLeft:" target="-1" id="1410"/>
                                                         </connections>
                                                     </menuItem>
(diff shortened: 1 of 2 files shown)


gnachman:
Merged without a shortcut in commit b9397b0f74ecfb35b7bb978c643f2b156894ec5b. I'm not ready to commit to cmd-opt-T.

PR #523, opened 2025-09-23
Move Tab to New Window: conditionally enable the menu item

cbowns:
(A cherry-pick of 2fb1f24ff7417c33b2d08861ecbb39fc7329b3ca from https://github.com/gnachman/iTerm2/pull/521)

Enable "Move Tab to New Window" only when there's > 1 tab in the active window


gnachman:
Thanks! Sorry for the delay in getting to it.

PR #525, opened 2025-10-05
Shift+meta function key offset correction

cwmaguire:
This fixes some function key (i.e. F1-F12) modifier key combination offsets in VT100Output.m to output the correct escape sequences.
https://invisible-island.net/xterm/xterm-function-keys.html

Index 6 of `offsets` was set to 74, which was skipping 2 entries.
Shift + Meta F5 and F6 were duplicated (`\e[15;4~`, F6 should be `\e[17;4~`)
Shift + Meta F7 to F10 were off by 1
Shift + Meta F11 and F12 were off by 2
Cases `73` and `74` were ignored (60 offset + 12 gives 72 max)

### DIFF
--- a/sources/VT100Output.m
+++ b/sources/VT100Output.m
@@ -574,7 +574,7 @@ typedef enum {
         case 71:
             return _optionIsMetaForSpecialKeys ? @"\e[23;10~" : @"\e[23;4~";
         case 72:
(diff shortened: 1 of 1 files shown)


gnachman:
Merged as 0a9372c4af9df2503f89e0fd6f6d06debeb7269f

PR #529, opened 2025-10-30
Fix inconsistent bottom padding in split panes

csmarshall:
When a terminal window is split into panes, the text positioning is now calculated from the bottom up instead of top down. This ensures that all panes have consistent bottom padding equal to the configured margin (kPreferenceKeyTopBottomMargins), with any excess pixels that don't fit evenly into line heights placed at the top rather than the bottom.

Previously, integer division when calculating rows from height would place excess pixels at the bottom, causing visual misalignment between panes when heights didn't divide evenly by line height.

Changes:
- Modified visibleRectExcludingTopMargin:startingAtRow: in PTYTextView.m to add a top offset equal to (excess - bottomMargin), aligning text from the bottom
- Updated comments in the excess method to reflect its new usage for calculating top offset

### DIFF
--- a/sources/PTYTextView.m
+++ b/sources/PTYTextView.m
@@ -1529,6 +1529,14 @@ static NSString *iTermStringForEventPhase(NSEventPhase eventPhase) {
     NSRect rect = self.enclosingScrollView.documentVisibleRect;
     // Subtract the top margin's height.
     rect.origin.y = row * _lineHeight;
+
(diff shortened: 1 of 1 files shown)


gnachman:
I suspect this will cause many bugs. For example, anywhere that `excess` is used, because there is a widespread assumption that the excess area is at the bottom.

PR #574, opened 2026-02-04
Optimize iTermProcessCache with coalesced incremental refresh

chall37:
## Summary

Refactoring to reduce CPU usage, frequency and duration of blocking operations on main thread from process monitoring with many tabs open.  The process monitoring code path dominates in workloads that generate process churn across multiple tabs (e.g., compiling in two or more visible tabs/windows).

- **Event Coalescing**
  - Legacy: Each monitor callback can trigger a full refresh
  - Refactor: Uses DISPATCH_SOURCE_TYPE_DATA_OR coalescer - multiple events merge into one refresh cycle. 50
  monitors firing = 1 coalescer callback.

- **Foreground/Background Split**
  - Legacy: All monitors run equally, no distinction
  - Refactor: Adds pauseMonitoring()/resumeMonitoring() + setForegroundRootPIDs():
    - Foreground tabs: Fast path via coalescer
    - Background tabs: Monitors paused (dispatch source suspended), updates via 0.5s cadence timer (only 2 roots per tick)

- **Incremental Refresh**
  - Legacy: reallyUpdate() enumerates all processes via lsof on every update - O(all_pids)
  - Refactor: Collection-first fast path walks cached collection - O(depth). Falls back to
   proc_listchildpids() only if stale.

- **Root PID Tracking**
  - Legacy: No trackedRootPID property
  - Refactor: Each monitor knows its root PID, enabling O(1) routing of callbacks to the right tab

- **Data Structures**
  - Legacy: Flat _trackedPidsLQ dictionary
  - Refactor: Adds iTermTrackedRootInfo with epoch tracking, _dirtyHighRootsLQ/_dirtyLowRootsLQ split, staleness detection

- Includes comprehensive unit tests

### Supporting changes

- `iTermProcessMonitor`: Add `trackedRootPID` property and pause/resume methods
- `iTermLSOF`: Add `childPidsForPid:` using `proc_listchildpids` with retry loop
- `PseudoTerminal`: Call `setForegroundRootPIDs:` on tab selection
- Add comprehensive unit tests for process cache and process monitor

### Comparison using process-generating stress tests across 5 tabs over 30 seconds:

| Metric | Legacy | Refactored | Change |                                                                           
|--------|--------|------------|--------|                                                                         
| proc_pidinfo | 634,151 | 19,666 | **-96%** |                                                                      
| listAllPids | 638 | 20 | **-96%** |                                                                               
| reallyUpdate | 638 | 20 | **-96%** |                                                                              
| newProcessCollection | 638 | 20 | **-96%** |                                                                    
| highPriorityRefresh | 0 | 1,110 | (new) |                                                                       
| backgroundRefreshTick | 0 | 15 | (new) |

### DIFF
--- a/sources/iTermLSOF.m
+++ b/sources/iTermLSOF.m
@@ -246,6 +246,50 @@
     }
 }
 
+// Returns direct child PIDs for a given parent PID.
+// Note: proc_listchildpids returns PID count (not bytes like proc_listpids).
++ (NSArray<NSNumber *> *)childPidsForPid:(pid_t)parentPid {
+    int count = proc_listchildpids(parentPid, NULL, 0);
+    if (count <= 0) {
+        return @[];
+    }
+
+    pid_t *pids = NULL;
+    int returnedCount = 0;
+
+    // Retry loop: child list can change between size query and fetch
+    for (int attempt = 0; attempt < 3; attempt++) {
+        int capacity = count + 16;  // Headroom for new children
+        size_t bufferSize = capacity * sizeof(pid_t);
+        pids = (pid_t *)iTermMalloc(bufferSize);
+
+        returnedCount = proc_listchildpids(parentPid, pids, (int)bufferSize);
+        if (returnedCount < 0) {
+            free(pids);
+            return @[];
+        }
+        if (returnedCount <= capacity) {
+            break;
+        }
+        // Buffer too small, retry with larger capacity
+        free(pids);
+        pids = NULL;
+        count = returnedCount;
+    }
+
+    // If all retries exhausted (child count kept growing faster than we could allocate), bail out
+    if (pids == NULL) {
+        return @[];
+    }
+
+    NSMutableArray<NSNumber *> *result = [NSMutableArray arrayWithCapacity:returnedCount];
+    for (int i = 0; i < returnedCount; i++) {
+        [result addObject:@(pids[i])];
+    }
+    free(pids);
+    return result;
+}
+
 // Use sysctl magic to get the name of a process and whether it is controlling
 // the tty. This code was adapted from ps, here:
 // http://opensource.apple.com/source/adv_cmds/adv_cmds-138.1/ps/
--- a/sources/iTermProcessCache.m
+++ b/sources/iTermProcessCache.m
@@ -67,6 +67,9 @@ typedef NS_OPTIONS(unsigned long, iTermProcessCacheCoalescerEvent) {
     NSMutableIndexSet *_dirtyHighRootsLQ;   // High-priority roots needing refresh (_lockQueue)
     NSMutableIndexSet *_dirtyLowRootsLQ;    // Low-priority (background) roots (_lockQueue)
     NSUInteger _currentEpoch;               // Incremented on each refresh cycle (_workQueue)
+
+    // Throttle for background refresh (ensures 0.5s minimum between background refreshes)
+    NSTimeInterval _lastBackgroundRefreshTime;
 }
 
 + (instancetype)sharedInstance {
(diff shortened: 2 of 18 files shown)


sources/iTermLSOF.m
```diff
@@ -246,6 +246,50 @@ + (pid_t)ppidForPid:(pid_t)childPid {
     }
 }
 
+// Returns direct child PIDs for a given parent PID.
+// Note: proc_listchildpids returns PID count (not bytes like proc_listpids).
++ (NSArray<NSNumber *> *)childPidsForPid:(pid_t)parentPid {
+    int count = proc_listchildpids(parentPid, NULL, 0);
```
gnachman, line 252: It might make sense to give yourself some headroom automatically to avoid having to loop in 99% of cases. Even doubling the count would have basically no performance or memory impact.

sources/iTermProcessCache.m
```diff
@@ -360,6 +468,375 @@ - (void)reallyUpdate {
...
+    __block iTermProcessCollection *collection = nil;
+    __block NSUInteger epoch = 0;
+
+    dispatch_sync(_lockQueue, ^{
+        if (self->_dirtyLowRootsLQ.count == 0) {
+            return;
+        }
+
+        NSMutableArray<NSNumber *> *picked = [NSMutableArray array];
+        __block NSUInteger remaining = 2;  // Process at most 2 background roots per tick
```
gnachman, line 772: If you have a hundred tabs does this mean it'll take up to 100 seconds to refresh each tab? At some point doing a single global refresh becomes worthwhile since with a large number of tabs you'll probably have to look at the vast majority of processes on the system anyway.

gnachman:
This is definitely an area that needs improvement.

My first reaction is to wonder if this whole thing would benefit from a fresh start instead of trying to bolt on an optimization. It would be nice to call out the tradeoffs we want to make in terms of responsiveness vs CPU churn and to think explicitly about how to handle having a small number of sessions that update rarely (where we can afford to be more aggressive) vs having a large number of sessions with a few that are churning hard (what I think this change was aimed at).

Here's what I've noticed so far (just read the code, haven't tried it yet):

## It looks like the fast path won't pick up new child processes until the half-second cadence tick:
  - A process forks in the foreground tab
  - The dispatch source fires, the coalescer runs refreshDirtyHighPriorityRoots
  - It walks the old collection, doesn't find the new child
  - Falls back to refreshRootWithKernelFallback which calls proc_listchildpids but returns nil for the deepest foreground job
  - The existing cache entry is preserved (stale data kept alive by processIsAlive:)
  - The new child's info only appears after the next full reallyUpdate, whenever that happens to run

  So the "fast path" for the foreground tab actually delivers stale data for structural changes (fork/exec). It only reliably detects process death. The cache preserves old data rather than showing nothing, but tab titles / job indicators won't update promptly for new processes.

## setForegroundRootPids is global but PseudoTerminal assumes it is per-window
  PseudoTerminal.m calls setForegroundRootPIDs: in both windowDidBecomeKey: and tabView:didSelectTabViewItem:. The implementation (setForegroundRootPIDsLQ:) iterates all roots and demotes anything not in the
  provided set. This means:

  - Switching to Window A sets Window A's current tab as foreground and demotes Window B's visible tab to background
  - Selecting a tab in a non-key window fires tabView:didSelectTabViewItem:, which sets that tab foreground and demotes the key window's tab

  If you have two windows side-by-side both showing compiles, only one gets fast updates at any time.

## _lastBackgroundRefreshTime has no synchronization

updateIfNeeded is documented as runnable on "Any queue" and reads/writes _lastBackgroundRefreshTime without locks (iTermProcessCache.m new lines ~388-392). This is a data race.

## _currentEpoch races between _workQueue and _lockQueue

  _currentEpoch is incremented on _workQueue in refreshDirtyHighPriorityRoots but read under _lockQueue in backgroundRefreshTick. These are different serial queues with no ordering guarantee.

## evictUnseenNodesForRoot:epoch: is a no-op
Remove it?

## cachedDescendants grows unbounded
cachedDescendants grows monotonically -- PIDs of dead children are never removed.

## iTermTrackedRootInfo.monitor comment says "nil if suspended" but the code doesn't nil it
suspendMonitorForRootLQ: pauses the monitor but keeps the reference. resumeMonitorForRootLQ: checks if (info.monitor) and creates a new one if nil. The comment and code disagree about the design intent.

## Testing infrastructure compiled into production
Can we limit the +Testing.m files to the ModernTests target?

PR #580, opened 2026-02-12
Add per-PTY dispatch sources and TaskNotifier bypass

chall37:
## Summary

This PR replaces TaskNotifier's single-threaded `select()` loop with per-PTY GCD dispatch sources for sessions using the fairness scheduler. It completes the I/O pipeline started in PR #567 (backpressure) and PR #568 (scheduler + executor), closing the circuit so that backpressure actually controls reading.

## Motivation

With PR #568, the scheduler controls *execution* order, but *reading* still goes through TaskNotifier's monolithic select() loop. Sessions accumulate tokens faster than the scheduler drains them. Per-PTY dispatch sources let each session's I/O be individually suspended/resumed based on backpressure level, so a session under heavy backpressure stops reading from its PTY until the scheduler catches up.

## Changes

- **PTYTask dispatch sources** — `setupDispatchSources` creates per-fd read/write GCD sources on a private serial queue (`_ioQueue`). Read source starts suspended and is controlled by `updateReadSourceState` based on `shouldRead` (not paused, I/O allowed, backpressure < heavy). Write source activates when `writeBuffer` has data.
- **Backpressure-gated reading** — `shouldRead` checks `tokenExecutor.backpressureLevel < .heavy`. When backpressure becomes heavy, the read source suspends; when the `backpressureReleaseHandler` fires (wired in PR #567), it resumes.
- **Registration ordering** — `didRegister` calls `taskDidRegister:` (which wires `tokenExecutor`) *before* `setupDispatchSources`, ensuring `shouldRead` sees backpressure state from the first event.
- **TaskNotifier bypass** — sessions using dispatch sources (`useDispatchSource` protocol method) are excluded from `select()` fd sets. TaskNotifier still handles legacy sessions, process status monitoring, and the run loop.
- **Coprocess I/O** — dispatch source read handler routes PTY output to coprocess (same as legacy `readTask:length:`).
- **Lifecycle** — `teardownDispatchSources` cancels both sources with proper suspension-count handling (GCD requires sources to be resumed before cancel). Called from `deregisterFileDescriptor:` and `closeFileDescriptor`.

## Benchmarks (5 tabs, 15s, Deployment build)

| Metric | Legacy | Fairness | Delta |
|---|---|---|---|
| Iteration rate | 18,392/s | 17,738/s | -3.6% |
| Metal FPS | 23 | 37 | **+61%** |
| Refresh FPS | 24 | 28 | +17% |
| Content frames | 43,443 | 50,335 | +16% |

With interaction injection (keyboard + scroll + tab switch):

| Metric | Legacy | Fairness | Delta |
|---|---|---|---|
| Iteration rate | 20,145/s | 18,893/s | -6.2% |
| Metal FPS | 10 | 36 | **+260%** |
| Refresh FPS | 17 | 29 | +71% |

The fairness scheduler trades ~4-6% raw throughput for significantly higher frame rates, especially under interactive load.

## Seam

All dispatch source code is gated by `useFairnessScheduler`:
- Flag OFF: TaskNotifier select() loop handles all I/O (identical to PR #568 behavior)
- Flag ON: Dispatch sources handle reading/writing; TaskNotifier skips those fds

## Test Coverage

- **PTYTaskDispatchSourceLifecycleTests** — source creation, teardown, double-teardown safety
- **PTYTaskReadStateTests** — shouldRead predicate under all backpressure levels
- **PTYTaskWriteStateTests** — write source activation with buffer data
- **PTYTaskEventHandlerTests** — read handler pipeline, data delivery to delegate
- **PTYTaskBackpressureIntegrationTests** — backpressure suspend/resume, registration ordering, preloaded-data race test
- **PTYTaskEOFTests** — broken pipe detection, EOF with buffered data
- **TaskNotifierDispatchSourceProtocolTests** — fd exclusion from select()
- **IntegrationTests** — end-to-end registration through TaskNotifier + real session wiring

## PR Sequence

| PR | Content | Status |
|----|---------|--------|
| #567 | Backpressure infrastructure | Closed |
| #568 | FairnessScheduler + TokenExecutor integration | Closed |
| **This** | Per-PTY dispatch sources + TaskNotifier bypass | Open |

Depends on #568.


gnachman:
This note is mostly to jog my own memory: we plan to a new class instead of modifying TaskNotifier. It can be written in Swift and be nice and clean. We'll still need to switch between the two based on the advanced setting since all this stuff has to be enabled gradually. Also, let's make that PR stand alone — this has gotten big enough that my head hurts.

PR #586, opened 2026-02-20
Zero-copy sync for single partial line blocks

chall37:
Evolves the incremental merge from 0ee322c into zero-copy sync for the eligible single-partial-line case. Instead of memcpy'ing deltas on each sync, progenitor and copy share the character buffer across sync cycles; sync updates only CLL and metadata.

The buffer is append-only: progenitor writes past the copy's CLL boundary, and the copy reads only within its current boundary. Merge advances the copy's CLL to include newly appended data, making sync O(1) instead of O(delta).

Non-append mutations (line completion, removal, drop, setPartial) break sharing via existing mutation guards and fall back to COW behavior. Buffer relocation on resize also disables zero-copy and uses the existing fallback path.

### DIFF
--- a/sources/LineBlock.mm
+++ b/sources/LineBlock.mm
@@ -348,8 +348,18 @@ NS_INLINE void iTermLineBlockDidChange(__unsafe_unretained LineBlock *lineBlock,
     return self;
 }
 
-// NOTE: You must not acquire a lock in dealloc. Assume it is reentrant.
+// NOTE: dealloc may be called reentrantly (releasing the lock can trigger another dealloc).
+// gLineBlockMutex is a recursive_mutex, so reentrant acquisition is safe.
 - (void)dealloc {
+    // If this block shares a character buffer with another block (shareCount > 1),
+    // decrement the count so the surviving block knows it is the sole owner
+    // and can skip unnecessary clones on mutation.
+    if (_characterBuffer.isShared) {
+        std::lock_guard<std::recursive_mutex> lock(gLineBlockMutex);
+        if (_characterBuffer.isShared) {
+            [_characterBuffer decrementShareCount];
+        }
+    }
     if ([self deinitializeClients]) {
         if (cumulative_line_lengths) {
             free((void *)cumulative_line_lengths);
@@ -778,6 +788,16 @@ static int iTermLineBlockNumberOfFullLinesImpl(const screen_char_t *buffer,
     return result;
 }
 
+// Clone the character buffer if it is shared (shareCount > 1), ensuring
+// exclusive ownership before a non-append mutation. The check-and-clone
+// is atomic under gLineBlockMutex.
+- (void)cloneCharacterBufferIfZeroCopyShared {
+    std::lock_guard<std::recursive_mutex> lock(gLineBlockMutex);
+    if (_characterBuffer.isShared) {
+        _characterBuffer = [_characterBuffer cloneAndDecrementShareCount];
+    }
+}
+
 - (BOOL)reallyAppendLine:(const screen_char_t *)buffer
                   length:(int)length
                  partial:(BOOL)partial
@@ -817,10 +837,7 @@ static int iTermLineBlockNumberOfFullLinesImpl(const screen_char_t *buffer,
         // If completing the line (hard EOL), the copy still thinks it's partial,
         // so zero-copy sharing must end and append tracking is invalidated.
         if (!partial) {
-            if (_zeroCopyShared) {
-                _characterBuffer = [_characterBuffer clone];
-                _zeroCopyShared = NO;
-            }
+            [self cloneCharacterBufferIfZeroCopyShared];
             _appendOnlySinceLastCopy = NO;
         }
         // update the numlines cache with the new number of full lines that the updated line has.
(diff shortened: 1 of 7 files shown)


chall37:
Actually, I think we can go even further.  The only mutations that actually need buffer isolation are the ones that change the logical view destructively (dropLines, removeLastRawLine, popLastLine), and even those don't modify buffer bytes, just CLL/metadata.

So the split could be:
- Appends (partial extend, line completion, new line): buffer stays shared, merge copies the CLL delta
- Destructive (drop, remove, pop): break sharing

That would keep zero-copy alive through line breaks, multi-line output, anything that's fundamentally appending.  Which is the overwhelmingly common case during normal terminal output, not just the pathological 50MB-no-newline scenario.

But I think that scope should be its own PR (assuming this one is compelling).

Edit: On second thought, I think actually modifying the existing COW to be append-aware is simpler and avoids all of the complexity of bypassing legacy cow to basically do parallel cow under a different name.

gnachman:
If you really want to take it to the limits, you don't need to copy the character buffer until characters would be overwritten. That means dropping from the start of a block is always safe (it just updates offsets, doesn't touch actual characters). Popping from the end is safe until a subsequent append.

But, yes please in another PR :)

sources/LineBlock.mm
gnachman, line 1824: Related to my comment about wasRelocated, this is only safe if we've already cloned the character buffer.

sources/LineBlock.mm
```diff
@@ -1813,6 +1817,10 @@ - (void)setPartial:(BOOL)partial {
     if (partial == is_partial) {
         return;
     }
+    if (_zeroCopyShared) {
```
gnachman: Factor out this if statement, which is repeated in many places.

sources/LineBlock.mm
```diff
@@ -1974,104 +1986,77 @@ - (BOOL)isSynchronizedWithProgenitor {
...
-    LineBlock *progenitor = _progenitor;
-    const int progenitorLen = progenitor->cumulative_line_lengths[progenitor->_firstEntry] - progenitor.bufferStartOffset;
-    const int myLen = cumulative_line_lengths[_firstEntry] - self.bufferStartOffset;
-    const int appendedLen = progenitorLen - myLen;
+- (void)zeroCopyMergeFromProgenitor {
+    // Merge is thread-confined to the main thread and intentionally does not
+    // take gLineBlockMutex. The mutex protects the COW ownership graph
+    // (cowCopy/validMutationCertificate on the mutation thread); the merge
+    // path is serialized with mutations at a higher level (screen lock).
+    ITAssertWithMessage(NSThread.isMainThread, @"Zero-copy merge must run on the main thread");
```
gnachman: I think we should take the lock. I don't believe there is much lock contention (since lock-holding mutations would happen concurrentlyvery rarely while in this code path). Not assuming this runs on the main thread frees us up to use LineBlock across threads in other contexts without limitation.

sources/iTermCharacterBuffer.m
```diff
@@ -70,14 +71,22 @@ - (instancetype)initWithChars:(screen_char_t *)source size:(int)size {
 }
 
 - (void)resize:(int)newSize {
+    screen_char_t *oldBuffer = _buffer;
     _buffer = iTermRealloc(_buffer, newSize, sizeof(screen_char_t));
+    if (_buffer != oldBuffer) {
+        _wasRelocated = YES;
```
gnachman: I don't think wasRelocated can be done correctly. It's never safe to resize a shared character buffer. You can't guarantee that the other reference isn't in use (for example, search runs concurrenctly in its own thread). The only way to safely realloc is to first make a copy of the character buffer so you know you're only copying your own instance and no other LineBlock could be affected by it.

sources/LineBlock.mm
```diff
@@ -826,9 +814,13 @@ - (BOOL)reallyAppendLine:(const screen_char_t *)buffer
...
         // append to an existing line
         ITAssertWithMessage(cll_entries > 0, @"is_partial but has no entries");
-        // If completing the line (hard EOL), invalidate incremental merge eligibility
-        // since the copy will still think it's partial.
+        // If completing the line (hard EOL), the copy still thinks it's partial,
+        // so zero-copy sharing must end and append tracking is invalidated.
         if (!partial) {
+            if (_zeroCopyShared) {
+                _characterBuffer = [_characterBuffer clone];
+                _zeroCopyShared = NO;
```
gnachman: It's odd that zeroCopyShared remains true in one line block after a copy clones the character buffer and sets _zeroCopyShared=NO. That leads to unnecessary copies:

```
a = LineBlock()
a.append("partial line")
b = a.cowCopy()
// a and b both have _zeroCopyShared=YES
b.pop() // b copies characterBuffer, resets _zeroCopyShared to NO
a.pop() // a copies characterBuffer, resets _zeroCopyShared to NO
```

Instead of holding the zeroCopyShared flag, it might make more sense for _characterBuffer to carry a reference count. Prior to making a destructive change, if the refcount > 1, clone it (and all of that should happen atomically).

gnachman (changes requested):

sources/LineBlock.mm
chall37: _zeroCopyShared boolean is completely gone, iTermCharacterBuffer now carries _shareCount and isShared checks _shareCount > 1.

sources/LineBlock.mm
chall37: Addressed as a consequence of removing wasRelocated.

sources/LineBlock.mm
chall37: Done — canZeroCopyMergeFromProgenitor acquires gLineBlockMutex at entry

sources/iTermCharacterBuffer.m
chall37: Removed wasRelocated. The approach is now clone-before-resize in changeBufferSize:cert:

sources/LineBlock.mm
chall37: factored into -cloneCharacterBufferIfZeroCopyShared

PR #587, opened 2026-02-20
Add per-PTY dispatch-source I/O path as alternative to TaskNotifier select loop

chall37:
## Summary

- Adds PTYTaskIOHandler: one serial dispatch queue per PTY with dispatch sources for read/write/process events, as
  an alternative to the shared TaskNotifier select() loop.
- Gates this path behind UsePerPTYDispatchSources (advanced setting, default off, restart required).
- Introduces routing helpers PTYTask.registerTaskWithNotifier: / deregisterTaskFromNotifier: so job managers are
  agnostic to the active I/O path.
- Adds backpressure-aware read control in per-PTY mode:
    - TokenExecutor now exposes a BackpressureLevel derived from available token-buffer slots.
    - PTY reads are suppressed under heavy pressure and resumed via a release callback wired from
      PTYSession.taskDidRegister:.
- Ensures suspended-read EOF delivery remains reliable by monitoring child exit and resuming reads to drain final
  data / broken-pipe.
- Keeps legacy behavior safe by restricting backpressure-gated read suppression to per-PTY mode (legacy TaskNotifier
  path continues unchanged in this area).

## Relationship To Fairness Scheduler

This PR was split out of #580 to stand alone per discussion. A few choices reflect that extraction:

- **BackpressureLevel granularity** — `none/light/moderate/heavy/blocked` is introduced even though this PR
  currently gates only on `>= heavy`. The extra levels are intended to support future scheduler policy/budget
  decisions.
- **Backpressure gating is per-PTY only** — `wantsRead` applies backpressure gating only when an `ioHandler` exists.
  Legacy `TaskNotifier` mode has no pressure-release wakeup, so gating there can stall reads. Future scheduler wiring
  can unify this behavior across both paths.

### DIFF
--- a/ModernTests/DispatchSources/DispatchSourceConcurrencyTests.swift
+++ b/ModernTests/DispatchSources/DispatchSourceConcurrencyTests.swift
@@ -1,0 +1,157 @@
+//
+//  DispatchSourceConcurrencyTests.swift
+//  ModernTests
+//
+//  Regression tests for data races in dispatch source management.
+//  These verify that concurrent operations on PTYTaskIOHandler don't crash.
+//
+
+import XCTest
+@testable import iTerm2SharedARC
+
+// MARK: - Coprocess Setup Serialization Tests
+
+/// Regression test: setupCoprocessSources must serialize source
+/// reference assignments on the ioQueue to prevent data races with event
+/// handlers and updateCoprocess*SourceState.
+final class DispatchSourceCoprocessSerializationTests: XCTestCase {
+
+    /// Rapid sequential setup/teardown of coprocess sources while the primary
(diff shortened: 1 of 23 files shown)


gnachman:
Thanks for taking the time to put this together! I'm trying to review it but I don't think I can do a good job on a PR of this  size. This is critical infrastructure so it's better to be slow and methodical. Probably it would make the most sense to chat about it instead of trying to discuss it asynchronously in a PR. LMK what's a good time when you have an hour or so and we can talk through what's going on here and what would be the best way to proceed. I'm pretty slammed until Tuesday, so any time after that should be fine.

PR #607, opened 2026-03-05
Migrate Python API to websockets new asyncio API (>=14.0)

mkusaka:
Follow-up to #604. Migrates `connection.py` from the legacy websockets API to the new asyncio-based API introduced in websockets 14.0, as discussed with @gnachman.

### Why websockets>=14.0?

websockets 14.0 is the first version where the new asyncio implementation became the default ([changelog](https://websockets.readthedocs.io/en/stable/project/changelog.html)). In 13.0, the new implementation was introduced but opt-in; in 14.0, `websockets.connect` and `websockets.unix_connect` point to the new API by default, and the legacy implementation was deprecated.

### Migration references

- Upgrade guide: https://websockets.readthedocs.io/en/stable/howto/upgrade.html
- New API `connect()` docs (`additional_headers`, `response.headers`): https://websockets.readthedocs.io/en/stable/reference/asyncio/client.html
- `InvalidStatus` vs `InvalidStatusCode` (legacy): https://websockets.readthedocs.io/en/stable/reference/exceptions.html

## Changes

**`connection.py`**
- Remove the legacy/new API compatibility shim (`try: import websockets.legacy.client` / `except`)
- `InvalidStatusCode` → `InvalidStatus` (access via `e.response.status_code` instead of `e.status_code`) — `InvalidStatusCode` is [legacy-only](https://websockets.readthedocs.io/en/stable/reference/exceptions.html)
- `extra_headers=` → `additional_headers=` — [new API parameter name](https://websockets.readthedocs.io/en/stable/reference/asyncio/client.html)
- `response_headers` → `response.headers` — new API uses `ClientConnection.response` object
- `websockets.legacy.client.connect` → `websockets.connect`
- `websockets.legacy.client.unix_connect` → `websockets.unix_connect`
- Remove all `# type: ignore` comments (no longer needed)

**`setup.py`**
- Pin `websockets>=14.0` (first version where new asyncio API is the default)

**`tests/test_connection.py`** (new, 12 tests)
- `TestIterm2ProtocolVersion` — version header parsing (4 tests)
- `TestInvalidStatusHandling` — 406/500/401 error handling in `async_connect` (4 tests)
- `TestAsyncCreate` — 406/500 error handling in `async_create` (2 tests)
- `TestConnectCoroutineArgs` — `additional_headers` parameter passing for TCP and Unix (2 tests)

## Test plan

- [x] `pytest tests/test_connection.py` — 12 passed
- [x] `pytest tests/` — 154 passed (no regressions)
- [x] Smoke-tested against a running iTerm2 instance with the following script:
  ```python
  import iterm2

  async def main(connection):
      app = await iterm2.async_get_app(connection)
      print(f"Protocol version: {connection.iterm2_protocol_version}")
      print(f"Windows: {len(app.windows)}")
      for w in app.windows:
          for tab in w.tabs:
              for session in tab.sessions:
                  print(f"  Session: {session.session_id}")

  iterm2.run_until_complete(main)
  ```
  ```sh
  cd api/library/python/iterm2
  uvx --with . python tests/smoke_test.py
  ```
  Result: connected successfully, protocol version `(1, 11)`, 4 windows / 16 sessions listed

### DIFF
--- a/api/library/python/iterm2/iterm2/connection.py
+++ b/api/library/python/iterm2/iterm2/connection.py
@@ -5,23 +5,11 @@ import iterm2.auth
 import os
 import sys
 import traceback
-import types
 import typing
 import websockets
+import websockets.exceptions
 
 gDisconnectCallbacks: typing.List[typing.Callable[[], None]] = []
-
-# websockets 9.0 moved client into legacy.client and didn't document how to
-# migrate to the new API :(. Stick with the old one until I have time to deal
-# with this.
-websockets_client: types.ModuleType
-try:
-  import websockets.legacy.client
-  websockets_client = websockets.legacy.client
-  from websockets.legacy.client import connect as websockets_connect
-except:
-  websockets_client = websockets.client
-  from websockets import connect as websockets_connect  # type: ignore[assignment]
 
 import iterm2.api_pb2
 from iterm2._version import __version__
@@ -114,8 +102,8 @@ class Connection:
                     connection._async_dispatch_forever(
                         connection, asyncio.get_running_loop()))
                 return connection
-            except websockets.exceptions.InvalidStatusCode as status_code_exception:  # type: ignore[attr-defined]
-                if status_code_exception.status_code == 401:
+            except websockets.exceptions.InvalidStatus as status_exception:
+                if status_exception.response.status_code == 401:
                     if have_fresh_cookie:
                         raise
                     # Force request a cookie and try one more time.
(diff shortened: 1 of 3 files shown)


gnachman:
Unfortunately it looks like we still have to support older versions of Python, while websockets 14 needs Python 3.9. For compatibility reasons, scripts are generally locked at the major version they were originally implemented in. That could be as old as Python 3.7. I don't have a good plan to get people off of ancient Pythons. Do you think we could maintain a backward compatibility codepath?

PR #609, opened 2026-03-09
Hoist semaphore ownership out of TokenExecutor

chall37:
Move creation and wait-site of the backpressure semaphore from TokenExecutor into its callers: **TaskNotifier** (PTY read path) and **PTYSession** (channel/mux path). 

TokenExecutor no longer owns any semaphore; it receives one via the semaphore: parameter and passes it through to TokenArray, which signals it on consumption exactly as before.

TaskNotifier creates a per-task semaphore on registration and looks it up (under tasksLock) before each read. PTYSession creates one per channel session. No semantic change intended.

As a reminder, this is the first step in flow control abstraction -- no behavioral changes are intended.  Next step would be the flow control interface.

### DIFF
--- a/sources/TaskNotifier.m
+++ b/sources/TaskNotifier.m
@@ -9,6 +9,7 @@
 #import "TaskNotifier.h"
 #import "Coprocess.h"
 #import "DebugLogging.h"
+#import "iTermAdvancedSettingsModel.h"
 
 #include <sys/time.h>
 #include <sys/select.h>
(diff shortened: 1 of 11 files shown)


chall37:
Looks like d518a5930 broke the libsixel submodule ref (from e2c8546 to d300081).  Don't see the d300081 commit in libsixel fork or upstream.  Maybe local changes?

sources/TaskNotifier.m
```diff
@@ -30,6 +31,10 @@ @implementation TaskNotifier
 
     // A set of NSNumber*s holding pids of tasks that need to be wait()ed on
     NSMutableSet* deadpool;
+
+    // Per-task backpressure semaphores, keyed by task identity.
+    // Created on register, removed on deregister.
+    NSMapTable<id<iTermTask>, dispatch_semaphore_t> *_backpressureSemaphores;
```
gnachman, line 37: What if PTYTask owned its semaphore? Then we wouldn't need `_backpressureSemaphores`: in fact, TaskNotifier could remain unchanged. `-[PTYTask processRead]` could wait on and signal its semaphore. Another benefit of doing the wait inside processRead is that it could read before waiting, reducing latency (probably very slightly) when it is eventually unblocked.

gnachman (changes requested):

PR #632, opened 2026-03-26
style(about): follow the app theme; plain window on macOS 26, frosted glass before

kud:
### 📄 Description

Makes the About window follow iTerm2's theme rather than only the system appearance, updating live, and gives the backer list a proper container.

- **macOS 26:** the window is a plain window-background window, the way the system's own About panel is. The backer list sits in a rounded container that shares the sponsor cards' fill and radius (12pt on 26); cards and container use fills rather than hairline borders. Light and dark share one path since the colours are dynamic.
- **Before macOS 26:** dark themes get a frosted-glass look (`NSVisualEffectMaterialUnderWindowBackground`) with a translucent scrim behind the backers so the small text stays readable over the blur. Light themes are untouched.

Built and checked on macOS 26.6 under the Dark theme.

Before

<img width="1118" height="737" alt="screenshot-2026-09-10-18-33-27" src="https://github.com/user-attachments/assets/7628dfcb-8fbc-4444-8406-9c5d2a9576d0" />

After

<img width="1126" height="1088" alt="screenshot-2026-09-10-18-30-00" src="https://github.com/user-attachments/assets/4162e2bf-cb77-4f67-9fb8-a2f2325484d0" />


### ✅ Review feedback

**Second review (8 Apr)**

| #   | Asked for                                                                    | Status                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                 |
| --- | ---------------------------------------------------------------------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| 1   | Use the existing theme API instead of tab-style checks and `currentTerminal` | Done. The window inherits `NSApp.appearance`, which `iTermApplication -updateAppearance` already derives from the theme, and the content view reacts in `viewDidChangeEffectiveAppearance` using `effectiveAppearance.it_isDark`. The `iTermController`, `PseudoTerminal`, `PTYWindow`, `PreferencePanel` and `iTermPreferences` imports are all gone. Inheriting rather than pinning means a system light/dark switch under Automatic re-themes too, with no notification observer. Happy to switch to `it_appearanceForCurrentTheme` + `kRefreshTerminalNotification` if you'd rather it match Open Quickly exactly. |
| 2   | Unconditional HUD material                                                   | Done. On 26 the material is `windowBackground`; before 26, dark uses `underWindowBackground` and light restores the material the nib gave the view.                                                                                                                                                                                                                                                                                                                                                                                                                                                                    |
| 3   | Hardcoded magic numbers                                                      | Done. File-scope `static const` values, each with a comment on intent.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                 |
| 4   | Duplicated `kWrapperInset` / `kWrapperGap`                                   | Gone with the panel they sized (see below).                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                            |
| 5   | Description vs diff                                                          | Done. This description matches the diff.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                               |
| 6   | `_sponsorWrapper` lifecycle                                                  | Removed. With the per-logo cards master gained since, a panel behind the row was double-grounding and read as a band across the window. The container is now behind the backer list instead, and it is hidden where it does not apply.                                                                                                                                                                                                                                                                                                                                                                                 |
| 7   | XIB font and colour changes                                                  | Removed. They were outside the PR's title and you had deferred on them, so the xib is untouched.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                       |

**First review (1 Apr):** all seven points remain addressed. Source only, no submodule change, single commit, theme handled outside `-init`, no height or sizing changes, `linkColor` kept, sponsor heading kept.

### 🧹 Also in this revision

- Rebased onto current master.
- The backer container is drawn from `updateLayer` (the same idiom as `iTermSponsorBoxView`) rather than by setting properties on the scroll view's layer, which AppKit resets on the next display pass. Earlier revisions of this PR were silently not painting the scrim for that reason.
- The backer text view gets a small `textContainerInset` so names do not touch the container's rounded corners. Nothing in the nib moves.

### ✅ How to Validate

1. About iTerm2 on macOS 26, Dark theme: plain dark window, backers in a rounded container matching the sponsor cards, no borders.
2. Same under a Light theme: plain light window, same container.
3. On macOS 15 or earlier, Dark theme: dark frosted glass with a translucent scrim behind the backers. Light theme: identical to master.
4. With the window open, switch Appearance → Theme between light and dark: it re-themes in place. Under Automatic, switch the system appearance: same.
5. Resize the window: the container tracks the backer list.
6. Heading and all five sponsor logos present and clickable.

### 🛠️ Developer Checklist

- [x] Single atomic commit
- [ ] Tests included and passing (if applicable): no coverage exists for window theming here
- [x] Built and checked in the running app on macOS 26.6, Dark theme
- [ ] Light theme and live theme switching on 26: not yet checked
- [ ] Checked on a pre-26 macOS: not available here; the pre-26 path is the earlier revision's, with the paint fix above

### DIFF
--- a/Interfaces/AboutWindow.xib
+++ b/Interfaces/AboutWindow.xib
@@ -28,7 +28,7 @@
                         <rect key="frame" x="181" y="361" width="389" height="27"/>
                         <autoresizingMask key="autoresizingMask" widthSizable="YES" flexibleMinY="YES"/>
                         <textFieldCell key="cell" sendsActionOnEndEditing="YES" alignment="center" title="iTerm2" id="U4n-GV-8aZ">
-                            <font key="font" metaFont="system" size="22"/>
+                            <font key="font" metaFont="systemBold" size="26"/>
                             <color key="textColor" name="controlTextColor" catalog="System" colorSpace="catalog"/>
                             <color key="backgroundColor" name="controlColor" catalog="System" colorSpace="catalog"/>
                         </textFieldCell>
@@ -97,7 +97,7 @@
                         <autoresizingMask key="autoresizingMask" widthSizable="YES" flexibleMinY="YES"/>
                         <textFieldCell key="cell" lineBreakMode="clipping" alignment="center" title="By George Nachman and Contributors" id="XAW-F5-RXa">
                             <font key="font" metaFont="system"/>
-                            <color key="textColor" name="labelColor" catalog="System" colorSpace="catalog"/>
+                            <color key="textColor" name="secondaryLabelColor" catalog="System" colorSpace="catalog"/>
                             <color key="backgroundColor" name="textBackgroundColor" catalog="System" colorSpace="catalog"/>
                         </textFieldCell>
                     </textField>
--- a/docs/notes-3.7.txt
+++ b/docs/notes-3.7.txt
@@ -2960,3 +2960,9 @@ Bug Fixes:
 - Fixed the git ahead/behind arrows in the
   session toolbar staying black in dark mode
   instead of matching the text color.
+- The About window now follows the current
+  theme, updating live when you switch between
+  light and dark. On macOS 26 the backer list
+  sits in a rounded container matching the
+  sponsor cards; before that, dark themes get a
+  frosted-glass look.
--- a/sources/AboutWindow/iTermAboutWindowController.m
+++ b/sources/AboutWindow/iTermAboutWindowController.m
@@ -10,6 +10,7 @@
 
 #import "iTerm2SharedARC-Swift.h"
 #import "iTermLaunchExperienceController.h"
+#import "NSAppearance+iTerm.h"
 #import "NSArray+iTerm.h"
 #import "NSColor+iTerm.h"
 #import "NSMutableAttributedString+iTerm.h"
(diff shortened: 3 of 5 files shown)


kud:
Once I have properly tested it and added a screenshot, I will move it out of draft.

gnachman:
The core idea (frosted glass for dark themes) is reasonable, but this PR needs significant rework:

  1. Remove all binary/framework artifacts — only ship source changes
  2. Remove the submodule change
  3. Squash WIP commits
  4. Move theme detection out of -init to handle theme changes
  5. Reconsider the 75% height and removal of dynamic sizing — these are significant behavioral changes
  6. Keep link color as linkColor for accessibility
  7. Don't remove the sponsor label

kud:
@gnachman Thanks for the feedback; I will adjust the PR.

Also, for now, I publish draft PRs to ensure people are aware of the work I'm doing. Perhaps opening an issue and stating that I will work on it would be better?

Regarding the commits, I will certainly polish the PR at the end.

gnachman (commented):
Thanks for the contribution! The idea of making the About panel respect dark themes is welcome. However, there are several issues that need to be addressed before this can be merged.

### Use the existing theme API

The PR manually checks tab styles and queries `currentTerminal` to determine dark/light mode, but iTerm2 already has a centralized API for this in `NSAppearance+iTerm.h`:

```objc
self.window.appearance = [NSAppearance it_appearanceForCurrentTheme];
```

This is what `iTermOpenQuicklyWindowController` and other non-terminal windows use. It correctly handles all tab styles (including Automatic and Compact, which this PR misses) and doesn't require a terminal window to exist. The current approach would silently fall through to light mode for Automatic/Compact users in dark mode, and has undefined behavior when no terminal is open under Minimal.

Using `it_appearanceForCurrentTheme` would also eliminate most of the new `#import`s (`iTermController.h`, `PseudoTerminal.h`, `PTYWindow.h`, `PreferencePanel.h`) which create unnecessary coupling between the About window and the terminal infrastructure.

To determine the `BOOL dark` value for `configureForDark:`, you can use:

```objc
BOOL isDark = [NSAppearance it_appearanceForCurrentTheme].it_isDark;
```

### Unconditional HUD material

`awakeFromNib` unconditionally sets `NSVisualEffectMaterialHUDWindow` on the content view. This material is designed for dark floating panels and will affect the appearance even in light mode. The material/blending configuration should be conditional, or use a material that works well in both modes (or skip the vibrancy for light themes entirely).

### Hardcoded magic numbers

The alpha values (`0.18`, `0.22`, `0.07`, `0.05`), corner radii (`10`, `8`), insets (`16`, `12`) are scattered without explanation. Please define these as named `static const` values at file scope with comments explaining the intent.

### Duplicated constants

`kWrapperInset` and `kWrapperGap` are defined identically in both `updateSponsorWrapperLayout` and `configureForDark:`. Hoist them to file-scope `static const` values.

### PR description vs. actual diff

The description mentions two changes that aren't in the diff:
- "switch version-string attributes to `secondaryLabelColor` + `NSFontWeightMedium`" — the version text still uses `controlTextColor`
- "tighten line spacing from 3 → 2 pt" — `setLineSpacing:3` is unchanged

Please either include these changes or update the description.

### `_sponsorWrapper` lifecycle

Once created, the sponsor wrapper is never hidden or removed. If the theme changes from dark to light at runtime, it remains visible with a slightly different background. Consider hiding it for light themes or making it truly transparent.

### Minor

The XIB changes (22→26pt bold title, `labelColor` → `secondaryLabelColor` subtitle) are cosmetic choices I'll defer to @gnachman on.

kud:
@gnachman haha the last comment :D I see you using AI :P (as I do)

kud:
Tadaaaaaaaaa 
screenshot-2026-09-10-18-30-00

kud:
@gnachman Round two addressed, and the PR has changed shape enough that I've retitled it: the frosted glass is now the pre-26 path only. On macOS 26 the window is a plain window-background window like the system About panel, with the backer list in a container that matches the sponsor cards. Still one source file plus a release note, and nothing moves in the nib.

The one place I went past the letter of the review: rather than pinning `it_appearanceForCurrentTheme` on the window, it inherits `NSApp.appearance` (which `updateAppearance` already sets from the theme) and reacts in `viewDidChangeEffectiveAppearance` via `it_isDark`. Same category, no `PreferencePanel.h`, and a system light/dark switch under Automatic re-themes too. If you'd rather it mirror Open Quickly exactly, that's a small change.

Two things dropped on the way: the panel behind the sponsor row (the per-logo cards you added since make it redundant, and it read as a band) and the two xib cosmetic edits, to keep this to its title. Details in the description.

PR #633, opened 2026-03-26
New Feature: you can choose a folder of photos as your background picture,it will change by time.

Josh050608:
## Summary

  This change adds a new background image source mode for profiles: `Folder Rotation`.

  In `Settings > Profiles > Window > Background`, profiles can now choose between:
  - `Single Image`
  - `Folder Rotation`

  When `Folder Rotation` is selected, the profile can store:
  - a folder path
  - a rotation interval in seconds

  At runtime, sessions using the same profile share the same rotated background image.

  ## What Changed

  - Added new profile keys for:
    - background image source mode
    - background image folder location
    - background image folder interval
  - Extended the Profiles > Window > Background UI with:
    - `Source`
    - `Folder`
    - `Interval (seconds)`
  - Added a profile-level background image rotation manager
  - Updated `PTYSession` to resolve background images through the rotation manager when folder rotation is enabled
  - Kept existing single-image behavior unchanged

  ## Behavior

  - Rotates images from the configured folder at the configured interval
  - Uses a shared runtime state per profile, so windows/panes with the same profile stay in sync
  - Re-scans the folder on each tick so added/removed images are picked up automatically
  - Continues polling even if the folder is temporarily empty, so rotation resumes automatically when images appear again

  ## Manual Testing

  Tested manually with a Development build:
  - Verified `Single Image` still works as before
  - Verified `Folder Rotation` settings save and reload correctly
  - Verified folder preview appears in the settings UI
  - Verified images rotate at the configured interval
  - Verified multiple windows/panes using the same profile stay in sync
  - Verified newly opened windows using the same profile inherit the current rotated image
  - Verified changing the folder or interval updates existing sessions
  - Verified an empty folder does not crash and resumes when images are added later

## Demo video
https://github.com/user-attachments/assets/74ad56a0-8a25-4cd9-b9d7-2ec960fbc48f


gnachman:
Thanks for the PR and the detailed writeup — folder rotation is a cool feature idea!

Before going further with the implementation, I'd like to step back and align on the architecture. There are a few things I'd want to think through together:

**New code should be Swift.** `iTermBackgroundImageRotationManager` should be a Swift class rather than Objective-C. New files in the project are written in Swift unless there's a specific reason not to.

**Profile change handling.** The PR adds a new `iTermProfileDidChange` observer in `PTYSession` that duplicates logic already handled by the existing `reloadProfile` path. Adding a parallel notification-driven path for profile changes risks double-updates and subtle ordering bugs with the existing machinery. I think the rotation integration should hook into the existing profile reload flow rather than creating a second one.

**Thread safety and complexity.** The GCD serial queue plus `dispatch_sync`/`dispatch_async` pattern, the deck-based shuffling logic, and the per-tick full directory scans with `NSImage` allocations add up to a lot of moving parts. I'd want to simplify this — for example, whether a simple timer on the main thread (with cached file lists) would be sufficient.

**Memory leak.** `[[NSImage alloc] initWithContentsOfFile:]` is used to test whether a file is a valid image, but the allocated object is never released. Under manual retain counting this leaks on every timer tick for every file in the folder.

Could you put together a short design proposal (in the PR description or a comment) covering:
1. Where the rotation state lives and how it interacts with the existing profile/session lifecycle
2. How the timer and file scanning work without blocking the main thread or leaking
3. How sessions pick up rotated images without duplicating the profile-change notification flow

Happy to discuss the design before you invest more time in code. Thanks again for the contribution!

Josh050608:
> Thanks for the PR and the detailed writeup — folder rotation is a cool feature idea!
> 
> Before going further with the implementation, I'd like to step back and align on the architecture. There are a few things I'd want to think through together:
> 
> **New code should be Swift.** `iTermBackgroundImageRotationManager` should be a Swift class rather than Objective-C. New files in the project are written in Swift unless there's a specific reason not to.
> 
> **Profile change handling.** The PR adds a new `iTermProfileDidChange` observer in `PTYSession` that duplicates logic already handled by the existing `reloadProfile` path. Adding a parallel notification-driven path for profile changes risks double-updates and subtle ordering bugs with the existing machinery. I think the rotation integration should hook into the existing profile reload flow rather than creating a second one.
> 
> **Thread safety and complexity.** The GCD serial queue plus `dispatch_sync`/`dispatch_async` pattern, the deck-based shuffling logic, and the per-tick full directory scans with `NSImage` allocations add up to a lot of moving parts. I'd want to simplify this — for example, whether a simple timer on the main thread (with cached file lists) would be sufficient.
> 
> **Memory leak.** `[[NSImage alloc] initWithContentsOfFile:]` is used to test whether a file is a valid image, but the allocated object is never released. Under manual retain counting this leaks on every timer tick for every file in the folder.
> 
> Could you put together a short design proposal (in the PR description or a comment) covering:
> 
> 1. Where the rotation state lives and how it interacts with the existing profile/session lifecycle
> 2. How the timer and file scanning work without blocking the main thread or leaking
> 3. How sessions pick up rotated images without duplicating the profile-change notification flow
> 
> Happy to discuss the design before you invest more time in code. Thanks again for the contribution!



Thank you for the detailed review — your points are well taken. We agree that the original ObjC manager was unnecessarily complex and that the parallel observer in `PTYSession` was redundant. The revised plan rewrites the manager in Swift, removes the duplicate notification path, and deliberately keeps the core logic simple: a plain timer plus a cached file list, with file scanning dispatched to a background queue to ensure the main thread is never blocked and no `NSImage` allocations are made that could cause memory leaks.

---

**Background Image Rotation — Design Plan**

**State management**
A new Swift singleton `BackgroundImageRotationManager` (exposed to ObjC via `@objc(iTermBackgroundImageRotationManager)`) holds per-profile state keyed by GUID: configured folder, interval, current image path, a shuffled image deck, a main-thread timer, a cached file list, and a generation counter for async invalidation.

**Timer and file scanning**
One `Timer.scheduledTimer` per active profile runs on the main thread. On each tick, directory enumeration is dispatched to a background `DispatchQueue(qos: .utility)`, files are validated via `UTType(filenameExtension:).conforms(to: .image)` (no `NSImage` allocation, no leak), and results are delivered back to the main thread before updating `currentImage` and posting a notification.

**Integration with existing lifecycle**
Profile config changes flow through the existing path only (`kReloadAllProfiles` → `reloadBookmarks` → `reloadProfile` → `setPreferencesFromAddressBookEntry:` → manager). For timer-driven rotation, the manager posts `iTermBackgroundImageRotationDidChangeNotification`; `PTYSession` already observes this and calls `backgroundImagePath(forProfile:)` to update the background. No new observers or notification paths are added to `PTYSession`.

**Files changed**
- Add: `iTermBackgroundImageRotationManager.swift`
- Delete: `iTermBackgroundImageRotationManager.h/.m`
- Edit: `PTYSession.m` (remove redundant observer), `iTerm2SharedARC-Bridging-Header.h`

gnachman (commented):
## Code Review

Nice feature — well-scoped and the PR is clean. Here are the issues I found:

### Bugs / Correctness

1. **Memory leak / perf — `[[NSImage alloc] initWithContentsOfFile:]` used as a boolean test (two locations)**

   In `iTermBackgroundImageRotationManager.m` (`sortedImagePathsInFolder:`) and `ProfilesWindowPreferencesViewController.m` (`firstImageFilenameInFolder:`):
   ```objc
   if ([[NSImage alloc] initWithContentsOfFile:url.path]) {
   ```
   This allocates an `NSImage` purely to test if the file is a valid image, then discards it. This is expensive I/O on every file in the folder on every timer tick. Consider checking file extensions (`.png`, `.jpg`, `.jpeg`, `.gif`, `.tiff`, `.heic`, `.webp`) as a first-pass filter, or using `CGImageSourceCreateWithURL` to cheaply validate.

2. **`sortedImagePathsInFolder:` does synchronous file I/O on every timer tick**

   Directory enumeration + NSImage loading for every file on every rotation interval. For folders with many files this is expensive. The re-scan design is fine, but the NSImage allocation per file makes it O(n)-heavy per tick.

3. **`profileDidChange:` notification handler in PTYSession duplicates `reloadProfile` logic**

   The new `profileDidChange:` method calls `sharedProfileDidChange`, copies `_originalProfile`, calls `sessionProfileDidChange`, and calls `profileNameDidChangeTo:` — all of which `reloadProfile` already does. Subscribing to `iTermProfileDidChange` this way risks double-processing profile changes (once via the existing `reloadProfile` call path, once via this new notification handler). I'd recommend investigating whether the existing `reloadProfile` path can be leveraged instead of adding a parallel handler.

4. **`dispatch_sync` from main thread can stall the UI**

   `backgroundImagePathForProfile:` uses `dispatch_sync(self.queue, ...)` and is called from the main thread (in `reallySetPreferencesFromAddressBookEntry:` and `backgroundImageRotationDidChange:`). If the serial queue is busy doing file I/O (scanning a large folder), the main thread blocks. Consider caching `currentImage` atomically so it can be read without blocking.

### Design Concerns

5. **No `NSNotificationCenter` removal for `iTermProfileDidChange` in PTYSession**

   The new selector-based observer for `iTermProfileDidChange` in PTYSession's init is never explicitly removed. Selector-based observers added with `addObserver:selector:name:object:` are NOT auto-removed on dealloc. Verify that PTYSession's `dealloc` does a blanket `removeObserver:self` — if it does, this is fine.

6. **Rotation state is never cleaned up when profiles are deleted**

   If a profile with folder rotation is deleted, `iTermBackgroundImageRotationManager` keeps its state and timer running indefinitely. Consider observing profile deletion notifications and cleaning up stale entries.

7. **Minimum interval of 1 second may be too aggressive**

   The interval is clamped to >= 1 second but a 1-second interval means a full directory scan + NSImage load per file every second. Consider a higher minimum (e.g., 5–10 seconds).

### Minor / Style

8. **Label changed from "Background image:" to "Image:"** — less descriptive in the single-image case, though the "Background Image" checkbox above provides context. Worth considering whether the old label was clearer.

9. **`firstImageFilenameInFolder:` duplicates `sortedImagePathsInFolder:`** — the preferences controller has its own folder-scanning method that does essentially the same thing as the rotation manager's `sortedImagePathsInFolder:`. Consider sharing the implementation.

Josh050608:
> ## Code Review
> Nice feature — well-scoped and the PR is clean. Here are the issues I found:
> 
> ### Bugs / Correctness
> 1. **Memory leak / perf — `[[NSImage alloc] initWithContentsOfFile:]` used as a boolean test (two locations)**
>    In `iTermBackgroundImageRotationManager.m` (`sortedImagePathsInFolder:`) and `ProfilesWindowPreferencesViewController.m` (`firstImageFilenameInFolder:`):
>    ```objc
>    if ([[NSImage alloc] initWithContentsOfFile:url.path]) {
>    ```
>    
>    
>        
>          
>        
>    
>          
>        
>    
>        
>      
>    This allocates an `NSImage` purely to test if the file is a valid image, then discards it. This is expensive I/O on every file in the folder on every timer tick. Consider checking file extensions (`.png`, `.jpg`, `.jpeg`, `.gif`, `.tiff`, `.heic`, `.webp`) as a first-pass filter, or using `CGImageSourceCreateWithURL` to cheaply validate.
> 2. **`sortedImagePathsInFolder:` does synchronous file I/O on every timer tick**
>    Directory enumeration + NSImage loading for every file on every rotation interval. For folders with many files this is expensive. The re-scan design is fine, but the NSImage allocation per file makes it O(n)-heavy per tick.
> 3. **`profileDidChange:` notification handler in PTYSession duplicates `reloadProfile` logic**
>    The new `profileDidChange:` method calls `sharedProfileDidChange`, copies `_originalProfile`, calls `sessionProfileDidChange`, and calls `profileNameDidChangeTo:` — all of which `reloadProfile` already does. Subscribing to `iTermProfileDidChange` this way risks double-processing profile changes (once via the existing `reloadProfile` call path, once via this new notification handler). I'd recommend investigating whether the existing `reloadProfile` path can be leveraged instead of adding a parallel handler.
> 4. **`dispatch_sync` from main thread can stall the UI**
>    `backgroundImagePathForProfile:` uses `dispatch_sync(self.queue, ...)` and is called from the main thread (in `reallySetPreferencesFromAddressBookEntry:` and `backgroundImageRotationDidChange:`). If the serial queue is busy doing file I/O (scanning a large folder), the main thread blocks. Consider caching `currentImage` atomically so it can be read without blocking.
> 
> ### Design Concerns
> 5. **No `NSNotificationCenter` removal for `iTermProfileDidChange` in PTYSession**
>    The new selector-based observer for `iTermProfileDidChange` in PTYSession's init is never explicitly removed. Selector-based observers added with `addObserver:selector:name:object:` are NOT auto-removed on dealloc. Verify that PTYSession's `dealloc` does a blanket `removeObserver:self` — if it does, this is fine.
> 6. **Rotation state is never cleaned up when profiles are deleted**
>    If a profile with folder rotation is deleted, `iTermBackgroundImageRotationManager` keeps its state and timer running indefinitely. Consider observing profile deletion notifications and cleaning up stale entries.
> 7. **Minimum interval of 1 second may be too aggressive**
>    The interval is clamped to >= 1 second but a 1-second interval means a full directory scan + NSImage load per file every second. Consider a higher minimum (e.g., 5–10 seconds).
> 
> ### Minor / Style
> 8. **Label changed from "Background image:" to "Image:"** — less descriptive in the single-image case, though the "Background Image" checkbox above provides context. Worth considering whether the old label was clearer.
> 9. **`firstImageFilenameInFolder:` duplicates `sortedImagePathsInFolder:`** — the preferences controller has its own folder-scanning method that does essentially the same thing as the rotation manager's `sortedImagePathsInFolder:`. Consider sharing the implementation.



 Thanks for the thorough review. I've addressed all of the issues point by point.

  ## Bugs / Correctness

  ### Memory leak / perf: `NSImage` used as a boolean test

  Fixed. I replaced the `NSImage` allocation check with
  `UTType(filenameExtension:)?.conforms(to: .image)`, so image validation is now just a UTI lookup with no
  image decoding or object allocation. This applies in both the rotation manager and the preferences view
  controller.

  ### `sortedImagePathsInFolder:` does synchronous file I/O on every timer tick

  The directory scan is still necessary so added/removed files are picked up, but the expensive per-file
  `NSImage` allocation is gone. Validation now uses the cheaper `UTType` conformance check, and timer-
  triggered scans run on a background `DispatchQueue(.utility)` so they do not block the main thread.

  ### `profileDidChange:` handler in `PTYSession` duplicates existing reload logic

  Removed. I deleted the parallel `iTermProfileDidChange` observer and its handler from `PTYSession`. The
  code now relies on the existing `kReloadAllProfiles -> reloadProfile ->
  setPreferencesFromAddressBookEntry:` path, which already calls `profileDidChange:` on the rotation
  manager.

  ### `dispatch_sync` from the main thread can stall the UI

  Fixed. The manager is now main-thread owned and returns a cached `currentImage` immediately from
  `backgroundImagePath(forProfile:)`, with no `dispatch_sync` or blocking. The initial scan is synchronous
  so callers get an image immediately, while subsequent timer-driven rescans happen on a background queue
  and deliver results back to the main thread with `DispatchQueue.main.async`.

  ## Design Concerns

  ### No `NSNotificationCenter` removal for `iTermProfileDidChange` in `PTYSession`

  No longer applicable, since that observer and handler were removed entirely.

  ### Rotation state is never cleaned up when profiles are deleted

  Fixed. `profileDidChange:` now tears down the state whenever the mode is no longer `folderRotation`,
  which invalidates the timer and removes the state from the dictionary. Because
  `setPreferencesFromAddressBookEntry:` calls `profileDidChange:` on every profile reload, switching modes
  or deleting the profile cleans things up immediately.

  ### Minimum interval of 1 second may be too aggressive

  Changed to 5 seconds. The Swift manager now clamps the interval with `max(interval, 5)`, and the
  preferences UI enforces the same minimum by snapping lower values up to 5.

  ## Minor / Style

  ### Label change

  Adjusted the layout by adding a `Background Image:` label next to the checkbox, and renamed the mode
  selector label from `Source:` to `Type:` for clarity.

  ### `firstImageFilenameInFolder:` duplicates `sortedImagePathsInFolder:`

  Consolidated. The preferences controller now calls `+[iTermBackgroundImageRotationManager
  firstImagePathInFolder:]`, which shares the same UTType-based scanning logic as the rotation manager.

  ## Additional Changes

  - Full Swift rewrite: the manager is now a single Swift file,
  `iTermBackgroundImageRotationManager.swift`, replacing the previous Objective-C `.h/.m` pair and exposed
  to Objective-C via `@objc(iTermBackgroundImageRotationManager)`.
  - Divorced session support: the manager now uses `KEY_ORIGINAL_GUID` so divorced sessions share rotation
  state with their parent profile.

gnachman:
Thanks for the PR! A few things to address before this can land.

On the localized strings file: iTerm2 doesn't actually use localization — Interfaces/PreferencePanel.strings is vestigial and not consumed at runtime. Please drop all the changes to that file
from this PR; the XIB is the source of truth for UI labels.

XIB label: The source-mode label (cell id c5V-L0-e1R) has title="Type:". If you meant "Source:", please update it in the XIB.

Other issues to fix:

- Double outlet binding on XIB id 6324. _backgroundImageLabel (new) and _blendAmountLabel (pre-existing) both target the same text field. Please untangle — either repurpose one outlet or add a
separate view — so mutating hidden/enabled on one doesn't stomp the other.
- RotationState is never reclaimed. Entries in states are only removed when profileDidChange is called with a non-rotation profile. If a profile is deleted while in rotation mode, the timer +
cached paths leak for the life of the process. Please observe profile deletion (or GC stale states) and invalidate.
- Interval validation inconsistent with the minimum. iTermProfilePreferences.m validates KEY_BACKGROUND_IMAGE_FOLDER_INTERVAL with integerValue > 0, but the UI sanitizer and the manager both
clamp to a minimum of 5. Please align — tighten validation to >= 5.
- Interval text field needs an NSNumberFormatter. Users can type non-numeric text today; integerValue returns 0 and gets clamped to 5, but the displayed text can drift from the stored value. Add
an integer formatter (min 5) on LPm-4S-0v9.
- backgroundImagePath(forProfile:) has hidden side effects. The name reads like a pure accessor, but the lazy branch creates state, kicks off a synchronous folder scan on the main thread, and
starts a repeating timer. Either drop the lazy branch (require callers to go through profileDidChange first) or rename to something like ensureRotationStarted(for:).
- Synchronous folder scans on the main thread in profileDidChange (folder-changed branch) and the lazy branch of backgroundImagePath. For a folder with many entries this can stall session
creation and profile edits. The async machinery already exists — please use it for the first scan too, posting a placeholder (currentImage = nil) and filling in on the next main-thread hop.
- profile as! [AnyHashable: Any] force-casts in iTermBackgroundImageRotationManager.swift (lines 28, 52, 58, 60, 130). This is a trap-capable cast for a free bridge. Either declare the parameter
as [AnyHashable: Any] and let Swift bridge automatically, or use as (non-trapping). as! is misleading here.
- Drop-on-image-well UX in folder mode. imageWellDidPerformDropOperation: calls checkFolder: on the dropped filename; dropping an image shows a "not a directory" warning. Please either hide the
image well in folder mode, accept the dropped image's parent directory, or branch on isDirectory with a clearer message.
- Trailing blank lines removed in unrelated files (iTermProfilePreferences.m, and the strings file which is going away anyway). Please drop those hunks.
- Menu wording. "Folder Rotation" reads a bit stilted — consider "Rotate images from folder" or similar. Minor, up to you.
- Tests. The rotation logic (deck shuffling, scan generation counter, folder-change reset) is self-contained and easy to cover in ModernTests. Please add some coverage for selectNextImage and the
 scanCompleted generation check — those are the subtle parts.

gnachman:
Merged as ac09a6595bc1d0e1b3a6f685466c5cf34ec1fba2. Thanks!

PR #641, opened 2026-04-05
🐛 fix(appearance): derive popup and panel appearance from the session background

kud:
> [!NOTE]
> Rebased onto current master and squashed to a single commit. Still draft: not rebuilt since the rebase. I'll build and verify before marking it ready.

### 📄 Description

Secondary windows and popups hardcoded `NSAppearanceNameVibrantDark` or used `it_appearanceForCurrentTheme`, which only reads the macOS system appearance. In the Minimal theme with a dark session background, the system appearance can still be light — so those windows rendered light while the terminal looked dark.

They now derive appearance from the session's background brightness, via one helper, falling back to the theme when there's no session context.

### ✅ Review feedback

All four points addressed.

| Asked for                                                                                                             | Done                                                                                                                                                                                                                                   |
| --------------------------------------------------------------------------------------------------------------------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| **Compact should not use session background colour** — only Minimal should; everything else respects the global theme | `TAB_STYLE_COMPACT` moved to the fallthrough group, exactly as written in the review. Only `TAB_STYLE_MINIMAL` is session-aware now.                                                                                                   |
| **Remove the redundant `popupWindowBackgroundColor`** in `iTermStatusBarLargeComposerViewController`                  | Removed. The protocol method is `@optional` and the caller guards with `respondsToSelector:` (`iTermPopupWindowController.m:166`), so not implementing it is identical to returning nil. That file now drops out of the diff entirely. |
| **Method naming** — it takes a colour, not a session                                                                  | Renamed `it_appearanceForCurrentSessionWithBackgroundColor:` → **`it_appearanceForSessionBackgroundColor:`**, with all five call sites updated.                                                                                        |
| The import removal in `iTermOpenQuicklyWindowController.m` is fine                                                    | Left as-is.                                                                                                                                                                                                                            |

The helper now reads:

```objc
+ (instancetype)it_appearanceForSessionBackgroundColor:(nullable NSColor *)backgroundColor {
    iTermPreferencesTabStyle preferredStyle = [iTermPreferences intForKey:kPreferenceKeyTabStyle];
    switch (preferredStyle) {
        case TAB_STYLE_MINIMAL:
            if (backgroundColor) {
                BOOL isDark = backgroundColor.perceivedBrightness < 0.5;
                return [NSAppearance appearanceNamed:isDark ? NSAppearanceNameDarkAqua : NSAppearanceNameAqua];
            }
            return NSApp.effectiveAppearance;
        case TAB_STYLE_COMPACT:
        case TAB_STYLE_AUTOMATIC:
        case TAB_STYLE_LIGHT:
        case TAB_STYLE_LIGHT_HIGH_CONTRAST:
        case TAB_STYLE_DARK:
        case TAB_STYLE_DARK_HIGH_CONTRAST:
            return [self it_appearanceForCurrentTheme];
    }
}
```

To be explicit about the earlier exchange: the separate **Minimal Dark / Minimal Light theme** idea I floated in the thread is dropped. Nothing here changes Minimal into a straight light/dark, and nothing here changes Minimal's appearance at all — this only affects which appearance the popups adopt.

### 🧹 Since the last review

- Rebased across 1058 commits, no conflicts
- Squashed the merge commits away; single commit now

### ✅ How to Validate

1. Minimal theme, **dark** session background, macOS in **light** system appearance
2. Open each of: Toast (e.g. a copy-mode notification), Global Search, Instant Replay, the autocomplete/paste-history popups, and the large composer — all should render dark, matching the terminal
3. Flip the session background to light and confirm they follow, including that toast text stays readable (this was the first review's point about white-on-light)
4. **Switch to Compact** and confirm popups now follow the _global theme_, not the session background — this is the behaviour change in this round
5. Regular / Automatic / explicit Light / explicit Dark — unchanged from master
6. Trigger a popup with no current session and confirm the theme fallback applies rather than a hardcoded appearance

### 🛠️ Developer Checklist

- [x] Single atomic commit
- [ ] Tests included and passing (if applicable) — no test coverage for appearance selection in this area
- [ ] Rebuilt and manually verified after the rebase — pending


kud:
https://github.com/gnachman/iTerm2/pull/640#issuecomment-4189336281

gnachman (changes requested):
Thanks for the PR — this addresses a real issue with hardcoded dark appearance on popups.

A few things to address:

### Toast text will be invisible in light mode

`ToastWindowController.m:62` hardcodes `[textField setTextColor:[NSColor whiteColor]]` with a black drop shadow. With your change, light terminals now get `NSVisualEffectMaterialSheet` + `NSAppearanceNameVibrantLight`, which produces a light translucent background. White text on that will be unreadable. The text color and shadow need to adapt along with the appearance.

### Toast fallback should match GlobalSearch/InstantReplay

The toast falls back to hardcoded `YES` when there's no session:

```objc
BOOL isDark = bgColor ? bgColor.perceivedBrightness < 0.5 : YES;
```

But GlobalSearch and InstantReplay fall back to the theme preference:

```objc
BOOL isDark = bgColor ? bgColor.perceivedBrightness < 0.5 : [NSAppearance it_appearanceForCurrentTheme].it_isDark;
```

Please make the toast use the same fallback for consistency.

### Please revert the Open Quickly tint changes

The `iTermOpenQuicklyView.m` tint value changes are unrelated to the dark mode popup fix and modify values I just tweaked in `2bf7bbe27`. Please drop those from this PR.

kud:
Shall be good

gnachman (commented):
Thanks for working on this — the core idea of using session background color to determine darkness is the right approach for the minimal theme case, and matches what `PseudoTerminal` already does for its own alert sheets.

A few things I'd like to see addressed:

## Extract the repeated detection pattern

The same 4-line pattern is copy-pasted in 6 places:

```objc
PTYSession *session = [[iTermController sharedInstance] currentTerminal].currentSession;
NSColor *bgColor = session.effectiveUnprocessedBackgroundColor;
BOOL isDark = bgColor ? bgColor.perceivedBrightness < 0.5 : [NSAppearance it_appearanceForCurrentTheme].it_isDark;
self.window.appearance = [NSAppearance appearanceNamed:isDark ? NSAppearanceNameDarkAqua : NSAppearanceNameAqua];
```

This should be extracted into a helper — e.g., a class method on `NSAppearance+iTerm` like `+it_appearanceForCurrentSession` — so there's a single source of truth.

## Drop the Sparkle theming

Sparkle windows are system-level dialogs with no meaningful relationship to a terminal session. Grabbing whatever session happens to be current is arbitrary, and the flag-based notification dance is fragile (any unrelated window becoming key between the delegate callback and the Sparkle alert appearing would get its appearance overridden instead). The system appearance / `it_appearanceForCurrentTheme` is the right choice for standalone panels like these — it's what users expect from a standard macOS dialog.

## Open Quickly is similar

Open Quickly is a global navigation panel (like Spotlight) — not tied to any particular session. The "current" session is arbitrary since you might be invoking it precisely to switch away. I'd use `it_appearanceForCurrentTheme` here too, same as before this PR.

## Make `popupWindowBackgroundColor` @optional

The method is added to `PopupDelegate` without `@optional`. Since the popup controller already nil-checks the result, making it `@optional` with a `respondsToSelector:` guard would be more defensive and consistent with Cocoa conventions.

## Minor: inconsistent isDark check in popup controller

`iTermPopupWindowController.m` uses `bgColor.isDark` while everywhere else uses `bgColor.perceivedBrightness < 0.5`. They're equivalent, but the fallback also differs (`owningWindow.effectiveAppearance.it_isDark` vs `it_appearanceForCurrentTheme`). Extracting the helper (point 1) would resolve this naturally.

---

The popup/toast/instant replay/global search changes all look good conceptually — just needs the cleanup above.

kud:
@gnachman I start wondering if we should have a Minimal Dark and Minimal Light theme to ease the conditions instead of checking the background to understand the Minimal theme. Because on the theme, there is also a proper Dark and Light theme.

gnachman:
I'm not that tempted to make this any more complicated than it already is, and the current minimal looks pretty cool IMO so I don't want to change it to a straight light/dark.

That said, light vs dark could be decoupled from theme. That would benefit Compact, which does not have light and dark variants as Regular does, and would let us have an app-wide theme separate from the system light/dark theme for Minimal.

I haven't pursued that becuase this is already so complex and the benefit is small.

kud:
@gnachman In this case, a helper to determine the theme thanks to the background for Compact and Minimal would work, but I do admit that even for the user, it's confusing to understand how you choose the dark theme when you select Compact and Minimal. Maybe a note in the app might be helpful.

kud:
@gnachman

All addressed in 72a4b26d4:

- Extracted the repeated detection into `+it_appearanceForCurrentSessionWithBackgroundColor:` on `NSAppearance+iTerm`. Made it theme-aware while at it: Minimal and Compact use session background brightness (they have no dedicated light/dark chrome variants); Regular/Automatic/explicit themes delegate to `+it_appearanceForCurrentTheme`.
- Dropped the Sparkle theming entirely — reverted all three delegate methods and the `_waitingForSparkleWindow` flag.
- Reverted Open Quickly to `+it_appearanceForCurrentTheme`.
- Made `popupWindowBackgroundColor` `@optional` with a `respondsToSelector:` guard.
- The `isDark` inconsistency in the popup controller is gone — it now calls the helper directly.

kud:
@gnachman

The new helper actually does a small version of this — `+it_appearanceForCurrentSessionWithBackgroundColor:` routes Minimal and Compact through session background brightness, while Regular/Automatic stay on `+it_appearanceForCurrentTheme`. Keeps the scope minimal but gets the right behaviour for the themes that need it.

gnachman (changes requested):
Thanks for tackling this — the visual inconsistency in Minimal theme with a dark session + light system appearance is a real problem and the approach of consolidating into a helper is the right one. A few things to address:

## Compact theme should not use session background color

The new helper groups `TAB_STYLE_COMPACT` with `TAB_STYLE_MINIMAL`, making both session-aware. Only Minimal theme should derive appearance from the session background color. Compact (and all other themes) should respect the global light/dark theme, matching the existing behavior in `it_decorationsAreDarkWithTerminalBackgroundColorIsDark:`.

Move `TAB_STYLE_COMPACT` to the fallthrough group:

```objc
+ (instancetype)it_appearanceForCurrentSessionWithBackgroundColor:(nullable NSColor *)backgroundColor {
    iTermPreferencesTabStyle preferredStyle = [iTermPreferences intForKey:kPreferenceKeyTabStyle];
    switch (preferredStyle) {
        case TAB_STYLE_MINIMAL:
            if (backgroundColor) {
                BOOL isDark = backgroundColor.perceivedBrightness < 0.5;
                return [NSAppearance appearanceNamed:isDark ? NSAppearanceNameDarkAqua : NSAppearanceNameAqua];
            }
            return NSApp.effectiveAppearance;
        case TAB_STYLE_COMPACT:
        case TAB_STYLE_AUTOMATIC:
        case TAB_STYLE_LIGHT:
        case TAB_STYLE_LIGHT_HIGH_CONTRAST:
        case TAB_STYLE_DARK:
        case TAB_STYLE_DARK_HIGH_CONTRAST:
            return [self it_appearanceForCurrentTheme];
    }
}
```

## Remove redundant `popupWindowBackgroundColor` in `iTermStatusBarLargeComposerViewController`

Since the method is `@optional` in the protocol and the caller guards with `respondsToSelector:`, not implementing the method produces the identical result (bgColor = nil → global theme fallback). The explicit `return nil` implementation is unnecessary — just remove it and let the `@optional` mechanism handle it.

## Minor nits

- **Method naming**: `it_appearanceForCurrentSessionWithBackgroundColor:` takes a color, not a session. Something like `it_appearanceForSessionBackgroundColor:` would be more precise.
- The import removal in `iTermOpenQuicklyWindowController.m` is fine — confirmed no category methods are used.

PR #644, opened 2026-04-06
🐛 fix(jobs-tool): remove orphaned quotes from process names containing shell-special characters

kud:
## 🎟️ Ticket

No ticket — bug fix surfaced during daily use.

### 📄 Description

The Jobs tool outline view displayed stray `"` characters when a process name contained shell-special characters such as `@` (common in npm/npx package names, e.g. `mcp-github-copilot@latest`).

**Root cause:** `iTermLSOF.commandLineArgumentsForProcess:` passes every argv token through `escapedArgument:`, which wraps `@`-containing strings in double quotes to make them shell-safe. That same escaped string was also used as the display value in the jobs outline view. When the column was narrow enough to clip the text, the orphaned closing `"` became visible at the truncation point — e.g. `mcp-github-copilot@latest"` or `node "/Users/kud/…`.

**Fix:** Separated the raw (display) and escaped (shell-safe) representations:

- Extracted a private `rawCommandLineArgumentsForProcess:execName:` that returns unescaped argv tokens.
- Added a new public `displayCommandForProcess:execName:` that joins raw tokens — used for UI display only.
- `commandLineArgumentsForProcess:` now delegates to the raw helper and then escapes each token, preserving the existing behaviour for all callers that need a shell-safe string.
- `iTermJobTreeViewController` uses `displayCommandForProcess:` for the outline cell value (no quotes).
- The "Copy Job Name" action continues to call `commandForProcess:` directly, so copy-pasted commands remain shell-safe.

### 📽️ Screencast

No visual changes beyond the removal of the stray quote characters.

### ✅ How to Validate

1. Run an npm/npx command whose package name contains `@`, for example:
   ```
   npx mcp-github-copilot@latest
   ```
2. Open **Window → Job Manager** (or the Toolbelt → Jobs tool).
3. Locate the `node` process running the package.
4. Resize the Jobs column to be narrow enough to clip the command — confirm no orphaned `"` appears.
5. Right-click the row and choose **Copy Job Name** — confirm the copied string is still correctly shell-quoted (e.g. `node "/Users/…/mcp-github-copilot@latest"`).

### 🛠️ Developer Checklist

- [ ] Tests included and passing (if applicable)
- [x] PR is atomic and focused on a single feature or bug


gnachman (changes requested):
Thanks for the fix — the separation of raw vs escaped representations is the right approach. Two things to address:

1. **Restore the removed comments in `rawCommandLineArgumentsForProcess:execName:`** — The comments explaining the procargs binary format (`// Consume argc`, `// Skip exec_path`, `// Skip trailing nulls`, `// Pull out null terminated argv components`) are helpful for understanding a non-obvious data structure. Please put them back.

2. **Tab/window titles should also use unquoted strings** — `iTermProcessInfo.swift` (around line 221) still uses `commandLineArguments(forProcess:execName:)` which returns escaped arguments, and joins them for `commandLineValue`. This is used for tab and window titles, so the same orphaned-quote problem can appear there. Please update that path to use unquoted arguments as well, so display strings are consistent across the UI.

kud:
@gnachman Should be fixed now, thanks

gnachman (changes requested):
## Issue 1: Unintended behavioral change in `iTermProcessInfo.swift`

The change at `sources/iTermProcessInfo.swift:221` from `commandLineArguments` to `displayCommandLineArguments` affects the `commandLine` property, which is consumed far beyond the Jobs tool:

- **Automatic Profile Switching** (`PTYSession.m:6753`) — users with APS rules matching command-line patterns will now see unescaped values. Anyone who wrote rules matching against quoted arguments (e.g., `node "/Users/…"`) will silently break.
- **Session variable** `commandLine` (`PTYSession.m:6706`) — exposed to scripts and the session title built-in function. This changes the format users and scripts may depend on.
- **Session title display** (`iTermSessionTitleBuiltInFunction.m:47`) — titles will change from e.g., `node "/Users/foo/pkg@latest"` to `node /Users/foo/pkg@latest`.

The fix for the Jobs tool should be scoped to `iTermJobTreeViewController` only — the `fullName` change there is correct, but the `iTermProcessInfo.swift` change has unintended blast radius. Please revert the `iTermProcessInfo.swift` change.

## Issue 2: `copyJobName` guard semantic change

In `sources/iTermJobTreeViewController.m:558`, the guard changed from `if (!info.fullName)` to `if (!info.pid)`. Since `pid` is `pid_t` (an int), `!info.pid` is true when pid == 0. The original tested whether a display name was available; the new version tests whether a PID is non-zero. These aren't equivalent. The subsequent `if (!shellCommand)` nil check covers the failure case so this is safe in practice, but the intent of the original guard was different — worth a second look.

kud:
@gnachman

📝 Both points addressed in 9a5e35bfe. The `iTermProcessInfo.swift:221` change is reverted back to `commandLineArguments`, and the now-dead `displayCommandLineArguments` method has been removed from the `ProcessDataSource` protocol and all conformances. On the `!info.pid` guard: the guard is intentional — we need a valid PID to call `iTermLSOF commandForProcess:`, so checking `pid` is the right sentinel here. Added an inline comment to `iTermJobTreeViewController.m:558` explaining that the escaped form is used so the user can paste directly into a shell, with `fullName` as the fallback.

kud:
@gnachman Is there a reason why I can't ask you to review it again? The status is still "changes requested".

kud:
Just realised why I couldn't request a review — I don't have write access to the repo since I'm contributing via a fork. All makes sense now.

PR #648, opened 2026-04-09
✨ feat(status-bar): add a Session Status component

kud:
> [!NOTE]
> Rewritten from scratch against the review below, and rebased onto current master. Still draft: not yet compiled. `swiftc -parse` is clean on all four Swift files, but that is syntax only — no type checking, no symbol resolution. I'll build and verify before marking this ready.

### 📄 What changed

This is a rewrite, not a revision. The previous version hardcoded three Claude Code status strings and built a parallel system around them. That's gone. This component is generic: it shows whatever any program has set via OSC 21337 or a Set Tab Status trigger, exactly as the toolbelt tool does.

Point by point against the review:

| Asked for                                                   | Done                                                                                                                                                                             |
| ----------------------------------------------------------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| Don't hardcode status strings                               | Membership is `hasActiveStatus`. There is no string comparison anywhere in the diff.                                                                                             |
| Use `SessionStatusController` for all active statuses       | Subscribes via `addObserver`, retaining the returned `NotifyingDictionaryObserverToken` (its `deinit` is what unsubscribes).                                                     |
| Sort by `StatusPrioritySettings`, then age, then session ID | Uses `iTermSessionTabStatus.priority`, which is already wired to `StatusPrioritySettings.shared`. Tiebreak matches `ToolStatus`: priority → last-changed ascending → session ID. |
| Scope to the current window                                 | Yes — see the note below, the suggested method wasn't reachable.                                                                                                                 |
| Popover lists sessions like the toolbelt tool               | Reuses `ToolStatusCellView` directly rather than reimplementing rows.                                                                                                            |
| Name it "Session Status"                                    | Done.                                                                                                                                                                            |
| Model it on `iTermStatusBarTriggersComponent`               | Done — see the note at the end.                                                                                                                                                  |

### 🔍 Two things I couldn't follow literally

**`toolbeltWindowContainsSession` isn't reachable from a status bar component.** It's on the toolbelt delegate protocol in `iTermToolWrapper.h`, implemented at `PseudoTerminal.m:15158`. `iTermStatusBarComponentDelegate` vends no session, window, or GUID — a component's only route to its own identity is `scope` → `iTermVariableKeySessionID` → `iTermController`, and `scope` is nil in the setup UI, which every path has to survive.

So I reimplemented the containment test from the same two public calls `PseudoTerminal` uses — `allSessions()` plus `peerPort?.containsPeer(guid:)`, so sessions buried in a workgroup still count — with a comment naming the original. Happy to move it somewhere shared if you'd prefer one implementation.

**The shortcut helper was fileprivate.** `shortcutString(for:)` lived in a `private extension ToolStatus`, so the popover couldn't reach it. Since the rows are supposed to look like the toolbelt tool's, and duplicating eighty lines to achieve that seemed worse, I extracted it to `sources/Toolbelt/SessionStatusShortcut.swift` and pointed `ToolStatus` at it. That's the `−102` in your file, and it's the one part of this diff that goes beyond adding a component.

The two callers differ only in which session counts as active, so that became the single parameter. If you'd rather I didn't touch `ToolStatus.swift` at all, say so and I'll drop the shortcut column from the popover rows instead — the rest of the PR is add-only.

### 📦 Files

| File                                                                      |                                 |
| ------------------------------------------------------------------------- | ------------------------------- |
| `sources/StatusBar/Components/iTermStatusBarSessionStatusComponent.swift` | new — the component             |
| `sources/StatusBar/Components/SessionStatusPopoverViewController.swift`   | new — the popover list          |
| `sources/Toolbelt/SessionStatusShortcut.swift`                            | new — shortcut logic, extracted |
| `sources/Toolbelt/ToolStatus.swift`                                       | −102, delegates to the above    |
| `sources/StatusBar/Setup/iTermStatusBarSetupViewController.m`             | +1 registry line                |
| `iTerm2.xcodeproj/project.pbxproj`                                        | +12                             |

The label summarises rather than enumerating — it offers the bar several width variants (`Compiling +2` / `Compiling` / `3`), leading with the highest-priority session's own text whatever that text happens to be, and falling back to a bare count when the winning status carries only an indicator dot and no text. You left this open in the review ("maybe a count, or the highest-priority status text"), so it's the part I'd most expect you to have an opinion on.

### ✅ How to Validate

1. Add "Session Status" to a status bar layout — it should appear in the setup UI picker
2. From a session, set a status: `printf "\033]1337;SetTabStatus=Compiling\a"` (or via a trigger). The component should show it
3. Set different statuses in several sessions in the same window — the label should lead with the highest-priority one and indicate the others
4. Set a status in a **different window** — the component should ignore it
5. Click the component — the popover should list the sessions with dot, name, shortcut and status text, matching the toolbelt tool
6. Click a row — it should navigate to that session
7. Reorder priorities in Status Priority settings and confirm the label's choice follows
8. Clear all statuses — the component should empty out
9. Split panes, and a session inside a workgroup peer, to exercise the containment test
10. Open the status bar setup UI with no session — nothing should crash on the nil scope

### 📝 One observation while I was in there

`iTermStatusBarTriggersComponent.statusBarComponentIsEmpty()` returns `true` when there **are** triggers:

```swift
return (dataSource?.numberOfTriggers ?? 0) > 0
```

Every sibling reads the other way — `iTermStatusBarActionComponent` uses `count == 0`, `iTermStatusBarBatteryComponent` uses `!machineHasBattery`. I didn't copy the polarity and I haven't touched the file; flagging it in case it's unintentional.

Also, since it changes the framing of the original PR: `sources/ClaudeCode/` has landed on master in the meantime, and `iTermSessionTabStatus` now carries a `backgroundTasks` field referencing `cc-status`. The argument for a generic component is stronger for it, not weaker — but I didn't want to write a description that pretended not to notice.

### 🛠️ Developer Checklist

- [x] Single atomic commit
- [ ] Tests included and passing (if applicable) — none added; happy to add coverage for the sort and containment logic if you'd like it
- [ ] Built and manually verified — pending


kud:
## Progress update

### ✅ Done

- **Status bar component** (`iTermStatusBarClaudeCodeComponent`) — registered, shows summary label ("1 waiting", "2 waiting, 1 working", "No sessions") with brain SF symbol icon
- **Session popover** — single-click opens a popover listing each Claude session with its name and a coloured status pill (teal = Waiting, orange = Working, green = Idle)
- **Hover effect** — subtle rounded highlight on row hover via custom `SessionRowView`
- **Pointing hand cursor** — only active over actual session rows (not empty space), via `NSTrackingArea` + `cursorUpdate`
- **Navigation** — clicking a row in the popover navigates to that session via `revealSession(withGUID:)` and closes the popover; double-click navigates directly when one session is active
- **No popup on empty state** — clicking "No sessions" does nothing
- **Icon spacing** — SF symbol constrained to 11pt to leave breathing room before the text
- **Unit tests** — 21 tests covering `ClaudeCodeSummaryBuilder` (status classification, pluralisation, edge cases)
- **Xcode project** — both files added via `add_file_to_xcodeproj.rb`

### 🔲 Not yet done / known gaps

- **Allow / Deny buttons in popover** — responding to Claude's permission prompts still requires switching to the tab manually. Implementing this requires detecting which prompt type is currently shown and injecting the correct keystrokes into Claude's TUI without ambiguity. Deferred to a follow-up.
- **Stale status after Claude exits** — when `claude` quits within a tab, its last status ("Idle", "Waiting") lingers in `SessionStatusController` until the tab itself closes. The trigger's `job: "claude"` gate prevents the shell's `❯` prompt from writing new statuses, but the old one persists. Needs a mechanism to detect process exit and clear the status (e.g. subscribing to `GlobalJobMonitor.didChangeNotification`).
- **Onboarding not updated** — the "Install Claude Code Integration" wizard still only mentions the Toolbelt Session Status tool; it doesn't mention the new status bar component.
- **Session filtering by provenance** — currently filters by status text values ("Waiting", "Working…", "Idle"). A more robust approach would propagate the trigger's `provenance: "claude-v1"` field through to `iTermSessionTabStatus` so non-Claude sessions that happen to use the same status strings are excluded.

gnachman:
Thanks for taking this on — a status bar version of the Session Status tool is a good idea and I'd like to see it happen. However, this PR needs a different approach. Let me explain why and how to get there.

## The core issue

The Session Status system in iTerm2 is generic. Any program can set status text, indicator dots, and colors on any session via OSC 21337 or triggers. The toolbelt's "Session Status" tool (`ToolStatus.swift`) shows all of these, regardless of what set them. Your PR hardcodes three specific strings ("Waiting", "Working…", "Idle") and builds everything around Claude Code specifically. That's too narrow — this component should work for any status, not just Claude Code's.

## What to study before starting over

Read these files carefully and understand how they fit together:

1. **`ToolStatus.swift`** — This is the toolbelt version of what you're building. Your status bar component should show the same data in a compact form. Pay attention to:
   - How it uses `SessionStatusController` to get *all* active statuses (no filtering by specific strings)
   - How it scopes to the current window via `toolbeltWindowContainsSession`
   - How it sorts by `StatusPrioritySettings` (user-configurable), then age, then session ID
   - How it does incremental table updates (insert/remove/move) instead of full reloads
   - How it shows dot indicators, status text with custom colors, session names, and keyboard shortcuts

2. **`ToolStatusCellView.swift`** — The cell layout: dot, name, shortcut, status text, detail. Your popover rows should show similar information.

3. **`iTermSessionTabStatus.swift`** and **`SessionStatusController`** — The data model. Note that `hasActiveStatus` is the right filter, not checking against hardcoded strings.

4. **`StatusPrioritySettings.swift`** — The user-configurable priority system. Use this for sorting instead of a hardcoded sort order.

5. **`iTermStatusBarTriggersComponent.swift`** — A good example of a status bar text component to model yours after.

## Suggested approach

1. Start by understanding the full data flow: OSC 21337 → `VT100TabStatusUpdate` → `iTermSessionTabStatus` → `SessionStatusController` → `ToolStatus`. Your component taps into the same pipeline at `SessionStatusController`.

2. The status bar label should summarize all active statuses in the current window — maybe a count, or the highest-priority status text. Don't hardcode specific strings.

3. The popover (or menu) should list sessions the same way the toolbelt tool does: session name, dot, status text, shortcut. Clicking navigates to the session.

4. Use `StatusPrioritySettings` for sort order so it matches the toolbelt tool's behavior and respects user configuration.

5. Scope to the current window, not all windows.

6. Name it "Session Status" to match the toolbelt tool.

## General advice

When contributing to an unfamiliar codebase, spend time reading the existing code before writing new code. The patterns are already there — your job is to extend them, not reinvent them. The toolbelt tool is your reference implementation; your status bar component should be a compact version of it, not a parallel system.

Give it another shot with this framing and I think it'll come together well.

kud:
@gnachman Okay, I see! I based this PR on the new Claude Integration actually. Maybe it was too Claude-centric, and you're right—we can make it more generic. I will have another round then. Thanks for the explanation.

gnachman:
Thanks for the rewrite. I rebased this onto current master and built it (the draft note said it wasn't compiled yet), then pushed my changes to a branch:

https://github.com/gnachman/iTerm2/tree/session-status-review-fixes

While testing, the popover clipped the last line of every row (the status text under the session name was cut off). I fixed three things:

1. `tableView(_:heightOfRow:)` needed `@objc(tableView:heightOfRow:)`. It is an optional `NSTableViewDelegate` method, so without the explicit selector the table never called it and fell back to its default 24pt row height, clipping the multi-line rows.
2. The popover was sized from a hand-summed row total, which omits the styled `NSTableView`'s own vertical inset (about 20pt). I now size `preferredContentSize` from the table's tiled frame height in `loadView`, so nothing is clipped.
3. Forced the view controller's view to load before the component reads `preferredContentSize`, so it uses the measured height rather than the provisional set in `init`.

My main design feedback is on the summary label. Right now it leads with the highest priority session's status text plus a count of the rest ("Compiling +2"). I think a count broken down by status would be more useful at a glance, for example "1 working, 2 waiting, 3 idle" (group sessions by their status text and show a count per group). That reads as a real summary of the window rather than one session's text with a number tacked on. The grouping stays generic since it keys off whatever status strings programs set. Happy to discuss the exact format.

PR #652, opened 2026-04-22
Reduce energy cost of short-interval timers and move recurring DNS off main thread

DatanoiseTV:
Small, self-contained follow-up to #649 — independent of it and touches unrelated files, so I opened it separately. Each fix is a verified energy or latency issue found while re-reading the codebase.

## Changes

**1. `iTermGlobalScopeController.m`** — the 60-second hostname refresh timer was calling `+[NSHost fullyQualifiedDomainName]` on the main thread, which can block on slow reverse-DNS resolvers (captive portals, VPN handoff, etc.). Moved the recurring lookup to a utility queue and publish the result back on main. The initial synchronous fetch at init is preserved so the variable is never briefly unset. Also stored the timer in an ivar and added `.tolerance = 6` to enable coalescing.

**2. Search-field animation timers** — four controllers schedule a 60 FPS redraw timer with no `.tolerance`, forcing precise 16.67 ms wakeups:
- `MiniFilterViewController.swift:183`
- `iTermDropDownFindViewController.m:232`
- `iTermGlobalSearchWindowController.m:102`
- `iTermMiniSearchFieldViewController.m:231`

Added `.tolerance = 1.0 / 120.0` (half the period) to each. All four still self-invalidate when the animation ends, so duration is unchanged — only wakeup coalescing is improved.

**3. Short-interval polling timers** — three more sites scheduled tight polling without tolerance:
- `TailFindController.swift:136` — 100 Hz → `.tolerance = 0.005`
- `ChatSearchResultsViewController.swift:48` — 50 Hz → `.tolerance = 0.01`
- `Autocomplete.m:750` — 100 Hz one-shot → `.tolerance = 0.002`

**4. `iTermTipController.m`** — `tryToShowTip` is scheduled recursively via `performSelector:withObject:afterDelay:` (~24h delay), but `dealloc` only removes `NSNotificationCenter` observers, not pending performSelector invocations. Added `cancelPreviousPerformRequestsWithTarget:` at the top of dealloc so any pending retry chain is torn down with the controller.

## Test plan

- [ ] Verify hostname variable updates normally on network change (it still refreshes every ~60s).
- [ ] Verify search field animations still complete visually at the same rate.
- [ ] Verify tail-find and chat search still find results promptly.
- [ ] Check in Energy Instruments that CPU wakes per second drop when a search field is animating on battery.

## Aside on issue tracking

Noticed the repository has GitHub issues disabled — I had a couple of small findings I wanted to file as separate issues rather than inflate this PR, and couldn't. If you'd like, enabling issues on the repo would make it easier for contributors to report small quality-of-life bugs like these without opening speculative PRs. Happy to defer to your existing workflow if there's a preferred tracker I should be pointing at instead.

### DIFF
--- a/docs/notes-3.7.txt
+++ b/docs/notes-3.7.txt
@@ -270,15 +270,11 @@ Settings Improvements:
   don't get reset by control sequence.
 
 Bug Fixes:
-- Move main-thread reverse-DNS lookup in
-  iTermGlobalScopeController to a background
-  queue so the UI cannot stall on slow
-  resolvers.
-- Add tolerance to short-interval timers in
-  search and autocomplete controllers so the
+- Add tolerance to search-field animation
+  timers and the tail-find poll timer so the
   OS can coalesce wake-ups on battery.
-- Cancel pending performSelector calls in
-  iTermTipController dealloc.
+- Add tolerance to the hostname refresh timer
+  in iTermGlobalScopeController.
 - Fix various crashes.
 - Fix wrong terminal size after exiting
   fullscreen with multiple tabs.
--- a/docs/notes-3.7.txt
+++ b/docs/notes-3.7.txt
@@ -270,6 +270,15 @@ Settings Improvements:
   don't get reset by control sequence.
 
 Bug Fixes:
+- Move main-thread reverse-DNS lookup in
+  iTermGlobalScopeController to a background
+  queue so the UI cannot stall on slow
+  resolvers.
+- Add tolerance to short-interval timers in
+  search and autocomplete controllers so the
+  OS can coalesce wake-ups on battery.
+- Cancel pending performSelector calls in
+  iTermTipController dealloc.
 - Fix various crashes.
 - Fix wrong terminal size after exiting
   fullscreen with multiple tabs.
--- a/sources/Autocomplete.m
+++ b/sources/Autocomplete.m
@@ -752,6 +752,7 @@ precededByWhitespace:(BOOL)precededByWhitespace
                                                              selector:@selector(_populateMore:)
                                                              userInfo:nil
                                                               repeats:NO];
+        _searchEngine.timer.tolerance = 0.002;
     }
 
     AcLog(@"Nothing more to do.");
--- a/sources/Autocomplete.m
+++ b/sources/Autocomplete.m
@@ -752,7 +752,6 @@ precededByWhitespace:(BOOL)precededByWhitespace
                                                              selector:@selector(_populateMore:)
                                                              userInfo:nil
                                                               repeats:NO];
-        _searchEngine.timer.tolerance = 0.002;
     }
 
     AcLog(@"Nothing more to do.");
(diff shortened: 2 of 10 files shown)


gnachman:
Thanks for the careful follow-up.

iTermGlobalScopeController — background DNS: +[NSHost fullyQualifiedDomainName] isn't actually doing reverse DNS here. Under the hood it resolves via gethostname(3), which on Darwin is just a sysctl read of kern.hostname — see Libc/gen/FreeBSD/gethostname.c. No network traffic, no resolver involved, so the main-thread stall you were protecting against can't occur. I'd prefer to drop the
  queue hop and the _hostnameTimer/updateLocalhostNameAsync scaffolding, but keep .tolerance = 6 on the existing timer — that part is a clear win.

Tolerance on Autocomplete.m (and similar rarely-used timers): I'm not sure this is worth it. Autocomplete only runs while the popup is visible during an active user interaction, which is infrequent and short-lived. The OS coalescing opportunity is tiny compared to the always-on timers, and it's one more thing to remember. Happy to take the tolerance on the repeating animation timers, as those can run for noticeable durations and are the ones that show up in Energy Instruments. I'd drop the one in Autocomplete and ChatSearchResultsViewController to keep the scope tight.  Keep: .tolerance on the four 60 FPS animation timers (MiniFilterViewController, iTermDropDownFindViewController, iTermGlobalSearchWindowController, iTermMiniSearchFieldViewController) and the TailFindController 100 Hz poll. Those are the real wins.

iTermTipController dealloc: performSelector:withObject:afterDelay: retains the target, so dealloc can't run while there's a pending invocation; and iTermTipController is a dispatch_once singleton anyway, so it never deallocates. Please drop that hunk and the corresponding release-notes line.

On the aside about GitHub Issues being disabled: thanks, that's intentional. Bugs and feature requests go to the iTerm2 GitLab at https://gitlab.com/gnachman/iterm2/-/issues. Github's issues are too limited in the attachments they accept.

DatanoiseTV:
Thanks — all addressed in the latest push.

- **Hostname timer** — pushed the correction. Kept only `.tolerance = 6` on the existing inline timer; reverted the async scaffolding and removed the `_hostnameTimer` ivar / `updateLocalhostNameAsync` method. Thanks for the pointer on `gethostname(3)` being a sysctl read — I had assumed reverse DNS without checking.
- **Autocomplete and ChatSearchResultsViewController tolerances** — reverted. Fair point on scope.
- **iTermTipController dealloc** — reverted the hunk and the release-notes line. Makes sense given the singleton + target-retain combination.

Kept the `.tolerance` additions on the four 60 FPS animation timers (`MiniFilterViewController`, `iTermDropDownFindViewController`, `iTermGlobalSearchWindowController`, `iTermMiniSearchFieldViewController`) and the `TailFindController` 100 Hz poll per your guidance.

Noted on GitLab as the right place for bugs — I'll use that for the small follow-ups I mentioned earlier. Thanks again.

gnachman:
Merged as 5abb241e689d3a6bcb61d2752c54ffbd9f0bd85f. Thanks!

gnachman:
A few tweaks in 5dd476236415e248150491bd69df51c9e49a36a8

PR #653, opened 2026-04-22
Add common Unicode range presets to Special Exceptions

Visio-Vanitas:
Summary

This adds an “Add Preset…” button to the Special Exceptions window for a few common Unicode ranges/scripts:

- Arabic
- Cyrillic
- Greek (Greek and Coptic)
- Han (CJK Unified Ideographs)
- Hangul Syllables
- Hiragana/Katakana
- Private Use Area

Special Exceptions already supports per-range font overrides, but configuring it currently requires manually entering Unicode code point ranges. This change does not introduce a new font fallback mechanism; it improves discoverability and usability on top of the existing per-range override system.

The existing Non-ASCII Font behavior remains unchanged:
- explicit Special Exceptions entries still take precedence
- unmatched characters still fall back to the legacy Non-ASCII Font

Details

- adds preset entries for common ranges/scripts in the Special Exceptions UI
- sorts preset menu entries alphabetically by title
- keeps the existing manual range-based workflow intact
- treats an empty font selection as unset so invalid entries are not persisted

Tests

Added focused tests covering:
- preset range definitions
- alphabetical preset menu order
- explicit range overrides taking precedence over the legacy Non-ASCII font
- unchanged fallback behavior when no preset matches
- no code point remapping for preset-only entries

Validation

- built the app locally
- ran `SpecialExceptionsPresetTests` successfully

Related

- GitLab feature request: https://gitlab.com/gnachman/iterm2/-/work_items/12823

### DIFF
--- a/ModernTests/SpecialExceptionsPresetTests.swift
+++ b/ModernTests/SpecialExceptionsPresetTests.swift
@@ -1,0 +1,111 @@
+import XCTest
+@testable import iTerm2SharedARC
+
+final class SpecialExceptionsPresetTests: XCTestCase {
+    private func makeDistinctFonts() -> [NSFont] {
+        let candidates = [
+            "Menlo",
+            "Monaco",
+            "Courier",
+            "Courier New",
+            "Helvetica",
+            "Times New Roman"
+        ]
+        var fonts: [NSFont] = []
+        for candidate in candidates {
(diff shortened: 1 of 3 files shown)


gnachman:
Merged as 9be6cc9e95a2933f83bc1935fb345d0c456c373c. Thanks!

PR #655, opened 2026-04-28
Add advanced setting to tint background of past commands

ctrlok:
Adds an opt-in advanced setting Tint background of every executed command line that paints a subtle background tint on the rows where each past command was typed. 

I added it because I think it looks cool 😄 

CleanShot 2026-04-28 at 20 34 04@2x

I used claude to implement this, still I read code, and it looks pretty normal to me 😅 and without any bullshit, but I'm not really macos developer, so please read carefully. 

Tint implemented by opt-in advanced setting, so I don't think it will affect a lot of users. 

Also I added 8px padding to the bottom to make it prettier. 

Also, I found that sb file for local building via `make paranoid-deps` is not correct, and I don't know - it is better submit it as another PR? 

Anyway, thanks for your work!


ctrlok:
found a bug with wrong allignment of the tint for custom fonts and custom margings on metal

ctrlok:
so it is all working now, still I'll take a few days to check all potential issues

gnachman:
Thanks very much for the PR and for taking the trouble to read through the code yourself.

I'm going to pass on this one, and I want to be upfront about the reason since it isn't really about your work. I've started leaning away from accepting changes that are purely gated behind an advanced setting. Now that AI tooling has made it nearly effortless to produce these, I'm seeing a lot more of them, and each one still carries a real ongoing cost for me: code to maintain, edge cases to chase down (transparency interactions, Metal/CG parity, selection overlays, folds, tmux integration, etc.), and a settings surface that keeps growing. The benefit of an opt-in cosmetic toggle that only the author and a handful of users will ever turn on doesn't quite balance that out anymore.

PR #657, opened 2026-05-01
Add bindable key action to toggle last tab

allewun:
"Toggle last tab" is a useful action for quickly jumping back and forth between tabs. Expose it as a bindable key action.

https://github.com/user-attachments/assets/7904cfd2-c86c-4792-b9fc-2a07fda1419e


gnachman:
Thanks for the PR! Before I merge this, I want to make sure we're not adding a redundant action. The existing Cycle Tabs Forward in MRU Order action (KEY_ACTION_NEXT_MRU_TAB) already produces  this exact toggle behavior on a single press: the currently-selected tab is at MRU index 0, so one press selects index 1 (your previous tab) and promotes it, and the next press toggles back.  It's the same Cmd-Tab-style behavior you've shown in the video.

Did you try binding to that action and find it didn't do what you wanted? If so, I'd love to understand the difference. Otherwise I'd lean toward not adding a near-duplicate action, and instead maybe improving the discoverability of the existing one (e.g. clearer label/help text).

allewun:
@gnachman I wasn't aware of the Cycle Tab actions (speaking of discoverability), but it doesn't appear to work the way you're describing.

Given `MRU[0] == current tab` and `MRU[1] == previous tab`, then the Cycle Tabs action goes from `MRU[0] -> MRU[1] -> MRU[2] -> ...` until it wraps back to 0.

Whereas the toggle tab goes: `MRU[0] -> MRU[1] -> MRU[0] -> ...` (though I think [under the hood](https://github.com/gnachman/iTerm2/pull/657/changes#diff-6ef65b53f2a9053ddab20a52043dbc549734b78b685cf56021fadf9c1d8b01adR232) MRU is reset when you visit MRU[1], so it's really just `MRU[1] -> MRU[1] -> MRU[1] -> ...`)

gnachman:
Cycle Tabs has a little hack. It can do either one, depending on whether a modifier flag (e.g., command, control, shift, option) is held down continuously. So control-tab-tab-tab will take you to mru[2] while control-tab (release) control-tab (release) control-tab (release) just alternates between mru[0] and mru[1]

allewun:
Nice that works!

I understand the motivation to avoid introducing a near-duplicate action, but given how hidden this was, I'm struggling to come up with a concise and descriptive enough title to explain this nuance.

A separate action feels like the right balance of discoverability and duplication (I've never wanted to cycle beyond 1 tab, maybe there are others like me too).

allewun:
Thanks for clarifying in 7734036dff900faf737c4827ded73c4e6fedee54! Where does the help text show up btw?

PR #658, opened 2026-05-01
Update DeepSeek models to v4-flash/v4-pro and disable thinking mode

Diluka:
## Summary

Update DeepSeek AI model configurations to the latest available models (`deepseek-v4-flash` and `deepseek-v4-pro`) and explicitly disable thinking mode in API requests to prevent 400 errors.

Closes https://gitlab.com/gnachman/iterm2/-/work_items/12707

### Commit 1: Update DeepSeek model configurations ([`AIMetadata.swift`](sources/AITerm/AIMetadata.swift))

- Add `deepseek-v4-flash` (recommended) and `deepseek-v4-pro` models
- Remove deprecated models: `deepseek-chat`, `deepseek-coder`, `deepseek-reasoner`
- Update API endpoint from `/v1/chat/completions` to `/chat/completions`
- Update context window to 1M tokens, max output to 384K tokens

**Why:** The old `deepseek-chat` and `deepseek-reasoner` model names are being deprecated on 2026/07/24. The new v4 models offer significantly larger context windows (1M vs 128K) and output limits (384K vs 8K).

**Reference:** [DeepSeek Models & Pricing](https://api-docs.deepseek.com/quick_start/pricing)

### Commit 2: Disable thinking mode for DeepSeek v4 ([`DeepSeek.swift`](sources/AITerm/DeepSeek.swift))

- Add `thinking: {"type": "disabled"}` to all DeepSeek API request bodies

**Why:** DeepSeek v4 models enable thinking mode by default. In thinking mode, the API returns a `reasoning_content` field alongside `content`. According to [DeepSeek's thinking mode documentation](https://api-docs.deepseek.com/guides/thinking_mode), when tool calls are involved, `reasoning_content` **must** be passed back in all subsequent requests — otherwise the API returns a 400 error:

> *"The content[].thinking in the thinking mode must be passed back to the API."*

The current codebase does not store or propagate `reasoning_content` in the message history (`LLM.Message` has no field for it), so enabling thinking mode would cause failures in multi-round conversations with tool calls. Explicitly disabling it avoids this issue entirely.

**To properly support thinking mode in the future**, the following would be needed:
1. Add a `reasoning_content` field to `LLM.Message`
2. Parse `reasoning_content` from both streaming and non-streaming responses in `DeepSeekResponseParser` / `DeepSeekStreamingResponseParser`
3. Serialize `reasoning_content` back into assistant messages in `DeepSeekRequestBuilder.Message`
4. Handle the distinction: with tool calls → must pass back; without tool calls → can omit

### DIFF
--- a/docs/notes-3.7.txt
+++ b/docs/notes-3.7.txt
@@ -181,6 +181,7 @@ Triggers Improvements:
 AI Improvements:
 - AI chat can connect to LAN-hosted models
   over HTTP.
+- Add DeepSeek v4-flash and v4-pro models.
 
 Python API Improvements:
 - Move sessions to new tabs or windows via the
--- a/sources/AITerm/AIMetadata.swift
+++ b/sources/AITerm/AIMetadata.swift
@@ -343,6 +343,16 @@ class AIMetadata: NSObject {
         features: [.functionCalling, .streaming],
         vendor: .deepSeek
     )
+    // Deprecated by DeepSeek on 2026-07-24; remove after that date
+    private static let deepseek_chat = Model(
+        name: "deepseek-chat",
+        contextWindowTokens: 128_000,
+        maxResponseTokens: 8_000,
+        url: "https://api.deepseek.com/v1/chat/completions",
+        api: .deepSeek,
+        features: [.functionCalling, .streaming],
+        vendor: .deepSeek
+    )
     private static let gemini_2_0_flash = Model(
         name: "gemini-2.0-flash",
         contextWindowTokens: 1_048_576,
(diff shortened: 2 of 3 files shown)


Diluka:
@gnachman Could you take a look at this PR?

gnachman (commented):
Thanks for tackling this. The rationale for disabling thinking mode is well-explained and updating to v4 is the right move. A few requests before this lands.

**1. Please don't remove the legacy `deepseek-chat`, `deepseek-coder`, and `deepseek-reasoner` entries yet.**

DeepSeek's stated deprecation date is 2026-07-24, so the old aliases still work server-side for ~2.5 months. Removing them now means anyone who had one of these selected loses their model metadata immediately. `objc_url(modelName:)`, `contextWindowTokens(modelName:)`, etc. all return `nil` for unknown names, so their DeepSeek profile silently breaks.

Also, `deepseek-coder` and `deepseek-reasoner` aren't 1:1 replaced by `v4-flash`/`v4-pro` with thinking disabled. `reasoner` was specifically the chain-of-thought model, and `coder` was a code-specialized fine-tune. Users who picked those intentionally are losing real functionality, not just a name change.

Suggested: keep all three legacy entries in `models[]`, add a `// Deprecated by DeepSeek on 2026-07-24; remove after that date` comment above them, and let them age out naturally. The new `v4-flash` and `v4-pro` entries plus the `recommendedDeepSeekModel` switch can stay as-is.

**2. The `thinking: "disabled"` change is the right stopgap, but please add a code comment explaining *why*.**

A future reader looking at `DeepSeek.swift:114` will have no idea this isn't a quality or performance choice. Something like:

```swift
// DeepSeek v4 enables thinking mode by default, which adds a reasoning_content
// field to assistant messages. Their API requires that field to be echoed back
// in subsequent requests whenever tool calls are involved, or it returns 400.
// LLM.Message has no place to store reasoning_content yet, so we disable
// thinking to avoid breaking multi-turn tool-call conversations.
thinking: Thinking(type: .disabled))
```

**3. Make `Thinking.type` a typed enum.**

The `// "enabled" or "disabled"` comment is the kind of thing that drifts. Suggest:

```swift
private struct Thinking: Codable {
    enum Mode: String, Codable {
        case enabled
        case disabled
    }
    var type: Mode
}
```

**4. Release notes.**

Please add an entry to `docs/notes-3.7.txt` (max 50 cols). Something like "Added DeepSeek v4-flash and v4-pro models", since this is a user-visible default change.

Once these are in I'm happy to approve.

PR #661, opened 2026-05-06
fix: VT100ScreenMark dealloc crash with unresolved return-code

tjasko:
This change fixes a crash caused by `VT100ScreenMark` deallocation when `returnCodePromise` was requested but `setCode:` was never called. In that state, `_codeSeal` was left unresolved and iTermPromiseSeal asserted in dealloc with `_promise` still set, terminating the app.

The fix resolves the promise lifecycle in `VT100ScreenMark` dealloc. If `_codeSeal` still exists at teardown time, it is rejected with the default error and cleared before normal cleanup. This preserves the existing semantics for successful command completion while safely handling session teardown or similar early-exit paths where a running command never reports a final code.

A regression test was added in `VT100ScreenTests`. The test creates a mark, accesses returnCodePromise, intentionally does not call `setCode:`, and verifies deallocation does not crash. This reproduces the previously failing lifecycle pattern and locks in the behavior.

### DIFF
--- a/ModernTests/VT100ScreenTests.swift
+++ b/ModernTests/VT100ScreenTests.swift
@@ -414,6 +414,21 @@ class VT100ScreenTests: XCTestCase {
         XCTAssertTrue(VT100GridCoordRangeEqualsCoordRange(rangeAfterResize, expected))
     }
 
+    // MARK: - VT100ScreenMark dealloc tests
+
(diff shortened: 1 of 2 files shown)


gnachman:
Merged as 18ddc62f153558f6e21de12cf9f911066973aca6

PR #663, opened 2026-05-12
Add a preference to hide session status text from tab subtitles

ivnnv:
The colored dot on a tab already shows whether Claude (or anything writing tab status) is working, waiting, or idle. The text subtitle duplicates that signal and clashes with user-set tab colors. This adds an opt-out for the subtitle text without touching the dot, the toolbelt entry, or anything that writes status.

The preference defaults to ON, so existing installs see no change. It gates both `PTYTab`'s subtitle weaving and `psmTabStatusSubtitleColor`. Triggers, `cc-status`, OSC 21337, and `iterm2.set_status` keep writing the same payload, the gate is at the render layer only.

Two surfaces flip it: a checkbox on the "Using Session Status" installer step, and one in the Session Status toolbelt tool's gear popover. Both bind to the same key and post a change notification, so open tabs refresh immediately from either.
Screenshot 2026-05-12 at 19 26 06

Tested by deploying locally and toggling from both surfaces. Subtitle text and its color tint disappear and reappear without restart, dot stays, custom tab colors unaffected.
Screenshot 2026-05-12 at 19 37 05

### DIFF
--- a/docs/notes-3.7.txt
+++ b/docs/notes-3.7.txt
@@ -282,6 +282,11 @@ Expressions and Variables:
   interpolated strings.
 
 UI Improvements:
+- A new preference hides the session status
+  text from tab subtitles while keeping the
+  colored dot. Toggle it from the Session
+  Status toolbelt tool or during Claude Code
+  integration setup.
 - Contextual alternates (calt) are enabled by
   default for ligature-capable fonts.
 - Prettier, animated toasts.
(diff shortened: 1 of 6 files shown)


gnachman:
Merged as 2e5e9a1634813544029a99121ec589c3c5e6eb97. Thanks!

ivnnv:
hey @gnachman, thanks for landing this. small ask for next time: if the PR is being used mostly as-is, could you please land it (with your commit on top or a master commit later) in a way GitHub marks it as `Merged` instead of `Closed`? authorship is fine on the commit, this is mainly for the PR status / merged filters :)

gnachman:
I'm not sure how to do a squash and merge that marks it as merged. I don't use merge commits, which is unfortunately what Github wants to push us towards :(

PR #665, opened 2026-05-13
Fix GlobalKeyMap matching for keyboards with layout-dependent charactersIgnoringModifiers

TankTechnology:
## Problem

On keyboards where the same physical key reports different `charactersIgnoringModifiers` values depending on input method state (e.g., Chinese/Pinyin layouts where the backtick key returns 0xb7 in one mode and 0x60 in another), **GlobalKeyMap key bindings become unrecognized after text input**.

### Reproduction

1. Configure a GlobalKeyMap binding for Cmd+` (recorded when the keyboard returns char=0xb7)
2. Open iTerm2, create two split panes
3. Type any character in one pane (changes input method state)
4. Press Cmd+` — the binding is not recognized

The stored binding key is `0xb7-0x100000-0x32` but the keystroke from the event serializes as `0x60-0x100000-0x32` because `charactersIgnoringModifiers` now returns 0x60.

### Root Cause

In `iTermKeystroke.keyInBindingDictionary:`, the normal-path fallbacks all depend on the `character` field (from `charactersIgnoringModifiers`):

1. `serialized` — `0x<char>-0x<mods>-0x<keycode>` ✗ (char differs)
2. `legacySerialized` — `0x<char>-0x<mods>` ✗ (char differs)
3. `modifiedSerialized` — `:0x<modChar>:0x<mods>` ✗ (char differs)
4. Slow legacy iteration — same problem

None of these match when the character field varies.

### Fix

Add a final fallback using `portableSerialized` (`*-0x<mods>-0x<keycode>`), which matches on virtual key code and modifiers only — the same strategy already used when `LanguageAgnosticKeyBindings` is enabled. This is a 15-line addition that only activates when all existing matching strategies have failed, so it's fully backward-compatible.

## Commits

1. **Test reproduction** — Python script that 1:1 mirrors the `keyInBindingDictionary:` matching logic, plus a Swift XCTest
2. **Fix** — Add portableSerialized fallback in the normal matching path

## Testing

Run `tests/reproduce_keystroke_bug.py` to verify the fix:

```bash
$ python3 tests/reproduce_keystroke_bug.py
BEFORE fix: NO MATCH — BUG REPRODUCED
AFTER  fix: MATCH
```

The Swift XCTest in `ModernTests/iTermKeystrokeBindingTests.swift` covers the same scenarios and can be run with Xcode.

### DIFF
--- a/ModernTests/iTermKeystrokeBindingTests.swift
+++ b/ModernTests/iTermKeystrokeBindingTests.swift
@@ -1,0 +1,114 @@
+//
+//  iTermKeystrokeBindingTests.swift
+//  iTerm2
+//
+//  Test for key binding matching when charactersIgnoringModifiers differs
+//  from the recorded binding due to keyboard layout / input method state.
+//
+
+import XCTest
+@testable import iTerm2SharedARC
+
+final class iTermKeystrokeBindingTests: XCTestCase {
+
+    /// kVK_ANSI_Grave = 0x32 (50), the physical backtick/tilde key on US keyboards.
+    /// On Chinese/Pinyin layouts the same physical key reports a different
+    /// charactersIgnoringModifiers value (0xb7 = middle dot ·) in some input
+    /// modes and 0x60 (backtick `) in others.
+    private let graveKeyCode: Int32 = 0x32
+    private let cmdModifier: Int64 = 0x100000  // NSEventModifierFlagCommand
+
+    // MARK: - Binding created with Chinese-layout character (0xb7)
+
+    func testBindingRecordedWithChineseCharMatchesUSCharPress() {
+        // Simulate binding created when keyboard reports 0xb7 for the backtick key.
+        let bindingKey = "0xb7-0x100000-0x32"
+        let bindingValue: NSDictionary = [
+            "Version": 2,
+            "Apply Mode": 0,
+            "Action": 30,
+            "Text": "",
+            "Escaping": 0
+        ]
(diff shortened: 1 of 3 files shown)


gnachman:
Thanks for digging into this and for the thorough writeup, @TankTechnology. The diagnosis is right: on keyboards where `charactersIgnoringModifiers` changes for a physical key (a Chinese/Pinyin input method, or a layout switch), a stored binding stops matching.

We looked at the layout mechanics closely, including live testing with Pinyin, and decided not to take the unconditional keycode-only fallback. It runs in the default matching path, so it would silently change behavior for users who deliberately want character-based bindings that follow the letter across layouts (for example a QWERTY/AZERTY switcher who expects Cmd+` to require the key that actually produces that character, not the same physical key). At the keystroke level that case and the input-method case are indistinguishable from the character alone, so the choice really belongs to the user.

iTerm2 already has an opt-in for exactly this: "Interpret key bindings based on physical key" (Settings > Keys). What was missing was discoverability. So instead we landed a change that detects precisely your scenario (a keystroke that matches no binding by character but would match the same physical key + modifiers) and offers, once, to turn that preference on, with a Learn More explanation and a Don't Ask Again. That fixes the Pinyin/layout case without overriding anyone who relies on character-based matching.

Landed in 3541ec9c3. Closing in favor of that approach, but thank you again: the report and reproduction were genuinely helpful in getting to the right fix.

PR #667, opened 2026-05-16
Add user-configurable custom HTTP headers for AI backend requests

benedikt90:
Adds a key/value table in the "Configure AI Model Manually…" popover where users can define extra headers (e.g. proxy auth, routing tags) that are appended to every AI request when the master toggle is on. Custom headers override built-in headers on collision (DLog warning is emitted). Covers chat/completion, vector store, and file upload paths.


benedikt90:
accidental creation, closing

PR #668, opened 2026-05-19
Add "Confirm closing a tab" preference

nimitbhargava:
A new opt-in checkbox under Settings > General > Closing prompts the user before any tab is closed (Cmd+W on the last pane, Cmd+Opt+W, middle-click, or the close button). Default is off to preserve existing behavior; the existing "Confirm closing multiple sessions" still applies independently.

### DIFF
--- a/docs/notes-3.7.txt
+++ b/docs/notes-3.7.txt
@@ -417,6 +417,9 @@ Settings Improvements:
   that works while another app is in
   front, without activating iTerm2 until
   you commit a selection.
+- A new “Confirm closing a tab” setting in
+  Settings > General > Closing prompts
+  before any tab is closed (off by default).
 
 Bug Fixes:
 - Pressing Enter to commit an IME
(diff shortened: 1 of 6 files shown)


gnachman:
What is the motivation for this? You can request to confirm closing a session (Settings > Profiles > Session > Prompt before closing) which is awfully similar.

nimitbhargava:
You're right that they overlap. The gap I was trying to fill:

- The existing "Prompt Before Closing" is per-profile, so a user with several profiles has to set it on each one and new profiles default to off.
- It lives under Profiles > Session, which is harder to find than a single checkbox under General > Closing for someone who just wants "ask before closing any tab."

If you'd rather solve it a different way, e.g. an "apply to all profiles" affordance on the existing setting, or just leaving things as they are; happy to close this. Let me know what you'd prefer.

gnachman:
New profiles are a copy of the default profile, so as long as it's set there they will pick it up. I think the change that is really needed is to make it easier to modify multiple profiles at at the same time, but that is a much bigger change than this.

PR #669, opened 2026-05-19
Bind ⌘⇧T to Shell > Undo Close

nimitbhargava:
Matches the reopen-tab shortcut used by Chrome, Safari, Firefox, and other browsers, so the muscle memory developed in browsers also restores the last closed session/tab/window here. The existing Undo Close behavior in iTermApplicationDelegate already restores into the original window when it still exists, mirroring Chrome.

Frees ⌘⇧T from View > Show Tabs in Fullscreen, which is a rarely-used set-and-forget toggle that remains reachable via the menu and can be re-bound in Settings > Keys.

### DIFF
--- a/docs/notes-3.7.txt
+++ b/docs/notes-3.7.txt
@@ -210,6 +210,11 @@ Tabs, Windows, and UI:
   folder of images at a configurable interval.
   Pick a folder instead of a file in
   Settings > Profiles > Window > Background.
+- ⌘⇧T now triggers Shell > Undo Close,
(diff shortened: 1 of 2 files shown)


gnachman:
I implemented a different approach for this in commit 070d60df521d3dcc50f760b08c7d1ee874f02154

nimitbhargava:
Thanks for picking this up, and nice touch with the migration warning + moving the old binding to ⌘⇧U, I hadn't thought about easing the transition for existing users. Glad to see this in.

PR #670, opened 2026-05-23
Menu bar status item for busy AI agents

ivnnv:
Opt-in menu bar status item showing how many AI agents are currently working. Off by default, gated on the new ShowMenuBarItem advanced setting. The existing UIElement plus statusBarIcon combination still installs the same item for dock-less mode.

The icon recolors the existing StatusItem template asset orange and replaces the dollar glyph with the count (1 to 9, or + at 10 or more). Idle keeps isTemplate true so light/dark menu bars auto-tint.

Count is the union of two sources, deduped by namespaced id: terminal sessions whose OSC 21337 status text starts with "Working" (case-insensitive — so Working, Working…, Working… 24s all count, but Waiting/Idle/custom statuses don't), and ChatBroker chats whose agent participant is currently typing (the built-in AI chat).

Screenshot 2026-05-23 at 23 59 02
Screenshot 2026-05-23 at 23 58 57
Screenshot 2026-05-23 at 23 58 51
Screenshot 2026-05-23 at 23 58 40

## Test plan:
- [ ] Toggle Settings > Advanced > ShowMenuBarItem on; idle dollar icon appears.
- [ ] In any pane, run `printf '\e]21337;status=Working;indicator=#ffa500\a'` and confirm orange icon with count 1; repeat in more panes to climb, + at 10 or more.
- [ ] Clear with `printf '\e]21337;status=;indicator=\a'` and confirm idle returns.
- [ ] Set a non-working status with `printf '\e]21337;status=Waiting;indicator=#5f87ff\a'` and confirm the badge does not tick up. Same with `status=Idle`.
- [ ] Open the built-in AI chat, send a prompt, confirm the badge ticks up while the agent is typing.


gnachman:
Interesting idea but I think the scope is too narrow. I can see the value of a global status indicator but it needs to be useful more broadly to justify the added complexity. When clicked it should show something useful, like a list of sessions sorted by status, maybe integrate with triggers, which sessions have had bells. I'll keep mulling this over.

ivnnv:
sure @gnachman, I wanted to keep this one atomic but yeah this could be expanded on click to show the status of each of the tabs/agents right there and clicking to take you directly to the upfront app, at the given tab/agent you select, I could implement it in this very PR if you want 👍🏽

gnachman (changes requested):
Thanks for this! Nice feature. A few things I'd like addressed before merging.

**1. Busy-session count should key off the "Working" status text, not `hasIndicator`.**

In `busyCount()` you filter with `.filter { $0.hasIndicator }`, but the PR description and test plan say only sessions whose status text starts with "Working" should count. `hasIndicator` is set purely by the OSC 21337 `indicator=` field, independent of the status text, so:

- The test-plan case `printf '\e]21337;status=Waiting;indicator=#5f87ff\a'` sets `hasIndicator = true` and will tick the badge up, failing the step that expects it not to.
- More importantly, `WorkgroupIntrospection.swift` already documents this exact pitfall: "cc-status always sets a dot color even at idle, so hasIndicator alone is misleading... would make every Claude Code session look perpetually working." So every idle Claude Code session inflates the count.

The correct classification already exists. Please reuse `WorkgroupIntrospection.state(forTabStatus:)` (which switches on `statusText.lowercased()`), or at minimum filter on `statusText?.lowercased().hasPrefix("working")` to match the stated semantics.

**2. Remove the now-dead `statusBarItem` property on `iTermApplication`.**

This PR deletes the only writes to `self.statusBarItem` (the old install/remove block in `setIsUIElement:`), and there are no other readers anywhere. Please drop the `@property` declaration in both `iTermApplication.h` and `iTermApplication.m` so nobody assumes it still tracks a live status item.

**3. Avoid refreshing on every defaults write.**

`defaultsChanged` is wired to `UserDefaults.didChangeNotification`, which fires on any defaults write (we write defaults frequently). When the item is visible, each one runs `updateImage()` and re-renders a fresh `NSImage` even when the count is unchanged. Please observe only the relevant key(s), or short-circuit when the computed count and pref state haven't changed.

(1) is the important one; (2) and (3) are cleanup.

ivnnv:
hey @gnachman

All three done, plus rebased on master (was 610 behind).

Found one more while in there: refreshOnMain read the raw HideFromDockAndAppSwitcher pref instead of the effective isUIElement, so with UIElementRequiresHotkeys on the legacy item stayed after updateProcessType turned it off.

You were right about point 1 and I could confirm it live: cc-status sets lowercase working / idle but keeps hasIndicator true in BOTH, so the old filter counted idle sessions forever. Badge does 0 -> 1 -> 0 now.

PR #673, opened 2026-05-28
Show Codex CLI activity in tab status

ivnnv:
Codex doesnt emit OSC 21337, so iTerm2 currently cant show the usual AI-agent tab status for it. It does update the terminal title with a braille spinner while its working, so this PR uses that as a compatibility path for Codex only.

The shim watches title changes, verifies the foreground process is actually `codex`, and synthesizes the same Working/Idle status the existing OSC 21337 path produces. Real OSC 21337 status still wins, so this wont overwrite agents that already report status directly.

The synthesized status is tagged as shim-owned, so it only clears state it created, and its excluded from saved arrangements so restored windows dont come back with stale fake status. It re-derives from the next title or foreground-process update.

Matching is on the foreground process whose last path component is `codex`, so it works for both install layouts (brew native binary and the npm/npx Rust binary) and ignores lookalikes like `codex-ish`. Prerequisite for #670 (menu bar item counting working AI agents), which picks up Codex support for free once this lands.

## Test plan
- [ ] `codex` shows Working while the title spinner is active, Idle when it stops
- [ ] exiting Codex clears the synthesized status
- [ ] non-Codex commands with similar titles dont trigger it
- [ ] real OSC 21337 status is preserved, including inside a Codex tab
- [ ] restored windows dont persist synthesized status


ivnnv:
@gnachman friendly ping, did you get a chance to look at this? 🙇🏽 (I have been using this myself for a while now)
Its useful when working with the typical Claude/Codex combo, and opens the door to more CLI adapters later (like OpenCode)

gnachman:
This could be done but it's probably not a great idea because it is so tightly bound to the current behavior of Codex. The best way is for codex itself to send OSC 21337. If that is off the table, the second-best way is to add a new category of triggers that run on title change and make it part of a formal Codex integration. That way they can be updated if Codex changes its behavior in the future.

ivnnv:
thanks @gnachman! makes sense.

Reworked it the way you suggested: added a Title Changed event-trigger category, and Codex rides it as a built-in trigger now instead of the hardcoded hook in screenSetWindowTitle. the detection itself (spinner decode + Working/Idle mapping) stays in CodexTitleStatusDecoder/Adaptor, just reused from the trigger action, so same behavior but title handling is part of the trigger system now and updatable if Codex changes it.

Synced with master and resolved the conflicts.

gnachman:
The PRs I sent OpenAI has been making some progress and that is a much cleaner solution. Let's hold off on this and see if gets landed. You can find them from the issue: https://github.com/openai/codex/issues/25879

ivnnv:
hey @gnachman

just synced this with master and CI is green.

would you mind having a second look at here, i know the ideal is codex emitting the standard status codes and then this title parsing wouldnt be needed at all, but they said theyd do that months ago and nothing moved, so from now i think we better land the title based detection. if openai ever changes it ill happily follow up and swap this over

PR #674, opened 2026-05-29
fix: quit Vim keep full scrollback buffer with iTerm2's tmux integration

zeroleo12345:
fix the problem when using iTerm2’s integrated tmux, full-screen interactive applications like Vim can be configured not to enter the alternate screen. 
In this mode, the application repaints the main screen directly. When it clears from the home position, iTerm2 may discard part of the visible content instead of first preserving it in scrollback. This makes it “eats” part of the terminal history after quit Vim.

### DIFF
--- a/docs/notes-3.7.txt
+++ b/docs/notes-3.7.txt
@@ -293,12 +293,6 @@ Tmux Improvements:
 - Bracketed paste state is restored on attach
   (prep for tmux’s upcoming bracket_paste_flag
   format).
-- Integrated tmux now preserves local scrollback
-  outside tmux’s captured history when unpausing
-  a pane.
-- Integrated tmux now saves the visible primary
-  screen to scrollback before a non-alternate-screen
-  full-screen app repaints from the home position.
 - The per-pane modifyOtherKeys state is now
   restored on attach (using tmux 3.5+’s
   pane_key_mode format), so reattaching while
--- a/docs/notes-3.7.txt
+++ b/docs/notes-3.7.txt
@@ -293,6 +293,12 @@ Tmux Improvements:
 - Bracketed paste state is restored on attach
   (prep for tmux’s upcoming bracket_paste_flag
   format).
+- Integrated tmux now preserves local scrollback
+  outside tmux’s captured history when unpausing
+  a pane.
+- Integrated tmux now saves the visible primary
+  screen to scrollback before a non-alternate-screen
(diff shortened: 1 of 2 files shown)


gnachman:
Thanks for the PR, and for the clear writeup of the problem. Full-screen apps that repaint the primary screen and eat your content is a real annoyance, and I appreciate you digging into it.

I'm going to decline this one, though, and I want to explain why rather than just close it.

The core issue is how the change decides when to preserve the screen. tmuxClientIsRunningInteractiveAppWithoutAltScreen infers "an interactive full-screen app is running" from the DEC private modes (cursorMode, keypadMode, bracketedPasteMode, mouseMode). Because this is a heuristic, it is vulnerable to false positives. For example, zsh's zle enables bracketed paste (DECSET 2004) by default at the prompt. So at a normal shell prompt inside tmux integration, the new condition returns true, and ^L (which uses ED 0 from home in the screen/tmux case) would scroll the whole screen into scrollback and skip the prompt-mark cleanup in the existing branch just below it. That's essentially the behavior issue 8822 was filed about, and it's exactly the kind of regression that's hard to discover: it only shows up for some shells, in tmux integration, on a clear, so it could ship and quietly annoy people for a long time before anyone connects it back to this change.

The bigger reason, though, is that iTerm2 already has two supported mechanisms for this, and I think they serve your use case better:

1. Profiles > Terminal > "Disable alternate screen mode switching." With this on, iTerm2 ignores smcup/rmcup, so you can leave Vim configured normally (no need to suppress the alternate screen in your .vimrc) and it'll run on the primary screen. This is a deterministic, explicit choice rather than something inferred from terminal state.
2. The advanced setting "Save the screen to scrollback when an app moves the cursor above the prompt" (on by default). It's designed for exactly this: full-screen programs that don't use the alternate screen and repaint from the top, destroying visible content. It preserves the screen into history first, and because it keys off cursor-moves-above-prompt plus shell integration rather than sniffing DEC modes, it doesn't have the false-positive problem above.

If you try those and there's still an issue (for instance, if your Vim clears via ED from the home position rather than by moving the cursor up over existing content, so option 2 doesn't fire) I'd love to hear the specific escape sequences your setup emits. The right fix in that case would be to extend that existing, well-gated mechanism to cover the ED-from-home path, rather than add a parallel detector that can misfire on shells.

Thanks again for the contribution and for caring about getting this right.

PR #675, opened 2026-06-01
feat(api): add select_tab to CreateTabRequest for background tab creation

localparty:
## Summary

The Python API's `create_tab` (and the underlying `CreateTabRequest`) always
selects the new tab and orders its window to the front, so any automation that
creates tabs steals the user's keyboard focus on every create. This is asymmetric
with split panes — `SplitPaneRequest` doesn't raise or steal focus — and there is
no per-creation equivalent of `ActivateRequest`'s existing `select_tab` /
`order_window_front` controls. (See GitLab #8223.)

This adds an opt-in field so a client can create a tab in the background:

- One new attribute: `optional bool select_tab = 6` on `CreateTabRequest`.
- **Default-preserving:** unset (or `true`) reproduces today's behavior exactly.
- **Position is unchanged** — that's the existing `CreateTabRequest.tab_index`
  field (a background client just sets `tab_index = 0`); this change governs only
  selection/focus.
- **API-only, no UI** — no new preference or menu item; a user-facing toggle can
  be layered on later.
- Python: `Window.async_create_tab` / `iterm2.rpc.async_create_tab` gain
  `select: bool = True`.

## How it works

Tab selection in `-[PseudoTerminal insertTab:atIndex:]` is already gated by the
existing `automaticallySelectNewTabs`. The window/focus motion came from three
spots: `insertTab:atIndex:`'s `makeKeyAndOrderFront:` and its `setCurrentTerminal:`,
plus the launcher's `-[iTermSessionLauncher makeKeyAndActivateIfNeeded:]` (note
`canActivate = NO` only suppresses cross-app activation, not the in-app raise).
All three are now gated behind a new `automaticallyOrderFrontNewTabs` property
(default `YES`), mirroring `automaticallySelectNewTabs`, keeping the existing
`self.tabs.count == 1` clause so a brand-new window still appears.
`apiServerCreateTab:` derives both flags (and `launcher.makeKey`) from `select_tab`
and restores them after the create.

**Backward compatibility:** `select_tab` is `optional` (field 6, previously
unused); unset falls back to `YES` and the new property defaults to `YES`, so an
old client — or one that doesn't set it — gets exactly today's behavior. The
Python `select` parameter defaults to `True`.

## Test plan

Verified against a local build (run as an isolated `-suite` instance alongside a
normal iTerm2). A tab created with `select=False` does not move focus; the default
path is unchanged:

- [x] `select=False` into a **background** window — window not raised, tab not selected
- [x] `select=False` into the **active** window — focus stays put
- [x] `select=False` in a **single-window** setup
- [x] `select=False` with `tab_index=0` (leftmost insert)
- [x] `select=False` for the **1-tab → 2-tab** case (tab bar first appears — a real relayout)
- [x] `select=True` / unset — still raises + selects (no regression); new window still appears
- [x] New session is functional (real tty); proto field round-trips (`HasField`/value)
- [x] Existing suites `iTerm2XCTests` + `ModernTests` (this branch vs baseline
  `5956b2164`, 2026-05-31, macOS 26.3.1 / arm64 / Xcode 26.5): failing set
  identical to baseline → **0 new failures** (remaining failures are pre-existing
  or flaky env-dependent tests, unrelated to this change)

## Files changed

- `proto/api.proto` — new `select_tab` field.
- `sources/TerminalView/PseudoTerminal.{h,m}` — `automaticallyOrderFrontNewTabs`
  property; gates the window raise and `setCurrentTerminal:`.
- `sources/API/iTermAPIHelper.m` — honors `select_tab` (sets the flags +
  `launcher.makeKey`).
- `api/library/python/iterm2/iterm2/{window,rpc}.py` — `select` parameter.
- Regenerated: `sources/proto/Api.pbobjc.{h,m}`, `api/library/python/iterm2/iterm2/api_pb2.py`/`.pyi`.

> Generated files were regenerated with protoc 3.21.12, which emits the same ObjC
> runtime gen-version (30004) as the repo's pinned 3.19; maintainers preferring a
> minimal `api_pb2.py` diff can re-run `tools/build_proto.sh` with protoc 3.19.
> Wire format is unchanged.


gnachman:
Merged as ab4c5bdca5608fe191f33a4e01dcb69a571570cf

PR #676, opened 2026-06-04
Add a Lock Layout advanced setting

mtichikawa:
## Summary

Adds a “Lock window layout?” advanced setting (Windows section, default off). When enabled, it freezes a window’s layout so an accidental click or drag can’t rearrange or destroy it.

**Prevented while locked:**
- Splitting panes
- Adding tabs
- Closing panes and tabs (including “Close Other Tabs” / “Close Tabs to the Right”)
- Reordering tabs
- Dragging panes within a tab
- Moving panes or tabs to other windows

**Still allowed:**
- Resizing panes by dragging the divider
- Opening new windows (a new window still gets its first tab — you just can’t add a second while locked)
- Closing the whole window

## Implementation

Enforcement is centralized at a small number of choke points rather than spread across every action, so the lock holds regardless of how an operation is triggered — menu, keyboard shortcut, context menu, drag, or the scripting/Python API. `validateMenuItem:` additionally greys out the corresponding menu commands for clear feedback.

`createTabWithProfile:` only blocks adding a tab when the window already has one (`objectType == iTermTabObject`), so opening a new window and its first tab is unaffected.

## Notes / open questions

- Implemented as an advanced setting for low friction, but it could just as easily be a Window-menu toggle if you’d prefer it more discoverable.
- The lock intentionally applies to the scripting API too (“locked means locked”). Happy to scope it to UI-only if you’d rather scripts bypass it.
- Default is off; no behavior change unless enabled.

## Testing

- Builds clean (Development).
- Manually verified in a dev build: each blocked operation is inert and its menu item greyed; resizing panes, opening new windows, and closing the window still work; toggling the setting off restores normal behavior.

### DIFF
--- a/docs/notes-3.7.txt
+++ b/docs/notes-3.7.txt
@@ -206,6 +206,14 @@ Tabs, Windows, and UI:
   highlight the focused pane.
 - New Lock Size menu item and per-profile
   preference to prevent window resizing.
+- New Lock Layout advanced setting that
+  freezes a window’s layout: it prevents
+  splitting, adding or closing tabs,
+  reordering tabs, closing or dragging
+  panes, and moving panes or tabs to
+  other windows. Resizing panes, opening
+  new windows, and closing the window
+  still work.
 - You can now choose to show a Proxy icon in
   Compact and Minimal themes.
 - The tab color menu now has a color picker,
--- a/sources/AppKit/iTermApplicationDelegate.m
+++ b/sources/AppKit/iTermApplicationDelegate.m
@@ -420,6 +420,15 @@ static NSModalResponse iTermCompareRenderingRunModal(id self, SEL _cmd) {
 }
 
 - (BOOL)validateMenuItem:(NSMenuItem *)menuItem {
+    if ([iTermAdvancedSettingsModel lockLayout] &&
+        [menuItem action] == @selector(newSessionInTabAtIndex:)) {
+        // Frozen layout: refuse to add a tab to a window that already has one.
+        // Opening a new window (no current terminal, or one with no tabs) is fine.
+        PseudoTerminal *term = [[iTermController sharedInstance] currentTerminal];
+        if (term && term.tabs.count > 0) {
+            return NO;
+        }
+    }
     if ([menuItem action] == @selector(toggleUseBackgroundPatternIndicator:)) {
       [menuItem setState:[self useBackgroundPatternIndicator]];
       return YES;
--- a/sources/Settings/iTermAdvancedSettingsModel.h
+++ b/sources/Settings/iTermAdvancedSettingsModel.h
@@ -250,6 +250,7 @@ extern NSString *const iTermAdvancedSettingsDidChange;
 + (BOOL)laxNilPolicyInInterpolatedStrings;
 + (BOOL)leftAlignTitleBarMinimalTahoe;
 + (double)lightModeInactiveTabDarkness;
++ (BOOL)lockLayout;
 + (NSString *)llmPlatform;
 + (BOOL)logDrawingPerformance;
 + (BOOL)logRestorableStateSize;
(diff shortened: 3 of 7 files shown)


gnachman:
Thanks for this, Mike, and for the thorough write-up of what is and isn't blocked.

I'm going to close this one, but not because the idea isn't wanted. master already grew a very similar feature in the meantime: a per-window **Lock Layout** toggle (`-[PseudoTerminal toggleLayoutLocked:]`, backed by the `_layoutLocked` ivar) that:

- Is a real Window-menu toggle with on/off state feedback (addresses your "could be a menu toggle if you'd prefer it more discoverable" open question), rather than a global advanced setting.
- Is per-window rather than all-windows-at-once.
- Persists in window arrangements (`TERMINAL_ARRANGEMENT_LAYOUT_LOCKED`), so a locked layout stays locked across restart/restore.
- Guards a superset of the choke points in this PR (it additionally covers close-other-tabs, close-tabs-to-the-right, move-tab-to-new-window, and tab reordering at the same spots).

Because it touches the same methods (`createTabWithProfile:`, `asyncSplitVertically:`, `closeSessionWithConfirmation:`, `validateMenuItem:`, etc.), this PR would conflict rather than apply cleanly, and it would end up double-gating the same lines.

One genuinely useful thing your PR surfaced: both implementations only guard the *async/profile* construction methods (`asyncSplitVertically:`, `createTabWithProfile:`), so a couple of lower-level launcher paths slip past the lock:

1. **Split bypass** - "Open URL in Vertical/Horizontal Split Pane" (and semantic-history / file-open with a split style) routes through `iTermSessionLauncher` -> the synchronous `-splitVertically:before:addingSession:targetSession:performSetup:`, never touching `asyncSplitVertically:`.
2. **Tab bypass** - opening a URL/file as a new tab routes through `addSessionInNewTab:` -> `insertSession:atIndex:`, never touching `createTabWithProfile:`.

I'm fixing those two gaps directly in master so the lock is airtight regardless of trigger. Appreciate the contribution - it directly improved the shipped feature.

gnachman:
Following up now that the work this PR inspired has landed in master (commit 053dc0612).

Your PR surfaced that Lock Layout was only enforced at the async/profile construction methods (`asyncSplitVertically:`, `createTabWithProfile:`), which let a few lower-level paths slip past the lock. Digging into it turned up a bit more, so the fix went in two directions:

**Redirect content opens instead of letting them fail.** Opening a URL or file that would split or add a tab to a locked window now opens in a new window: the lock is honored and the user's action still succeeds (rather than silently doing nothing). This covers cmd-click URLs, OSC 8 shift-clicks, "Open URL in Split Pane", semantic history, registered-scheme URLs, and Finder/`open` file opens. It's done at the two choke points all of those flow through (`-[iTermController openURL:...]` and `-[iTermSessionLauncher possiblyNewWindowControllerForProfile:...]`). Scripted launches are deliberately excluded so they still respect the lock strictly ("locked means locked" for scripts).

**Disable every menu item that triggers a blocked operation.** A menu item that can't or shouldn't work needs to be greyed out, not left enabled to no-op. The audit found gaps in both the main and context menus: the context-menu Split Pane / Close / Move / Swap items, the tab-bar Close Tab and Duplicate Tab items, Move Session to Window / Tab, and the main-menu Move Pane item. Several of those were gated only on the per-pane lock, not the window layout lock, so a new delegate hook was added to expose the window's layout-lock state to the text-view context menu.

**Plug three holes with no choke point at all.** Move Session to Window, Move Session to Tab, and Duplicate Tab were mutating a locked layout directly, bypassing the lock entirely regardless of menu state. Those now have guards of their own.

Net result: the lock holds regardless of how an operation is triggered, which was exactly the goal you stated in this PR. Thanks again - this was a genuinely useful contribution.

mtichikawa:
Thanks, George — this is a great outcome from my end. The per-window toggle is clearly the better design than a global advanced setting: per-window scope, real menu state-feedback, and persisting in arrangements address exactly the open questions I'd flagged, so closing this in its favor makes total sense.

Really glad the PR was useful for surfacing the bypass paths. The two-direction fix — redirecting content-opens into a new window so the user's action still succeeds, and greying out the menu items rather than letting them silently no-op — is a much more thorough answer than just bolting on more guards, and the three direct-mutation holes (Move to Window/Tab, Duplicate Tab) are good catches. "Locked means locked for scripts" is the right call too.

I'll switch to a nightly to start using it. Thanks for the thoughtful close and the write-ups.

PR #677, opened 2026-06-06
Improve tab lifecycle performance

theanotherwise:
**Tuning for:**
- improved loading tabs from a file as iTerm2 tabs
- optimized closing behavior when many tabs are open
- improved tab switching/navigation with large numbers of open tabs
- improved rapid tab creation when holding the new-tab shortcut
- reduced redundant window-wide updates during bulk tab operations

### DIFF
--- a/ThirdParty/PSMTabBarControl/source/PSMMinimalTabStyle.m
+++ b/ThirdParty/PSMTabBarControl/source/PSMMinimalTabStyle.m
@@ -460,8 +460,8 @@ static CGFloat PSMWeightedAverage(CGFloat l, CGFloat u, CGFloat w) {
 
 - (PSMTabBarCell *)selectedVisibleCell {
     PSMTabBarControl *bar = self.tabBar;
-    for (PSMTabBarCell *cell in bar.cells.reverseObjectEnumerator) {
-        if (!cell.isInOverflowMenu && cell.state == NSControlStateValueOn) {
+    for (PSMTabBarCell *cell in bar.visibleCells.reverseObjectEnumerator) {
+        if (cell.state == NSControlStateValueOn) {
             return cell;
         }
     }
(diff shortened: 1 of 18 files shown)


theanotherwise:
@gnachman

gnachman:
Thanks for this, the performance wins here are real and I want them. My main concern is scope: the PR bundles several independent efforts (vertical tab virtualization, the bulk-operation deferral in `PseudoTerminal`, lazy progress indicators, title-update debouncing) together with some small, obviously-safe refactors. That makes the risky parts hard to review with confidence because they're interleaved with the trivial ones.

To shrink the diff, I've pulled three of the safe, self-contained changes straight to `master`:

- `393f7dc42` - fork child processes off the main thread (the legacy/monoserver `dispatch_sync` -> `dispatch_async`; also fixes the monoserver completion passing an unassigned `savedStatus`/`nil` instead of the real status + error code)
- `e917e2a92` - hoist the repeated `[aTab sessions]` calls in the tab close-confirm path
- `1a0e706da` - early-return in `-[PTYTab setObjectCount:]` when the count is unchanged

If you rebase on latest `master` these should drop out of your diff.

Could you split the rest into separate PRs? Roughly: (1) the vertical-tab virtualization in `PSMTabBarControl`/styles/drag assistant, and (2) the bulk-operation deferral in `PseudoTerminal`/`PTYTab`/`PTYTabView`. Those are each big and independent enough to want their own before/after profiling and their own review. The lazy-indicator and title-debounce changes are also separable if you'd rather land them incrementally.

One heads-up: `-[PTYTab isProcessing]` and `-setLabelAttributesForActiveTab:` drop the `![self isForegroundTab]` guard, which is a user-visible behavior change (activity spinner now animates on the foreground tab), not a cleanup. Let's discuss that one on its own rather than fold it in.

theanotherwise:
@gnachman i will prepare soon

PR #678, opened 2026-06-09
Improve AI chat provider and model UX

theanotherwise:


sources/AITerm/AITermControllerRegistrationHelper.swift
```diff
@@ -149,9 +180,22 @@ protocol AIRegistrationProvider: AnyObject {
         _ completion: @escaping (AITermController.Registration?) -> ())
 }
 
+extension AIRegistrationProvider {
+    func registrationProviderRequestRegistration(for vendor: iTermAIVendor,
```
gnachman, line 184: HIGH: the per-chat vendor is silently dropped here, which defeats the main feature of the PR. `registrationProviderRequestRegistration(for:_:)` is only a protocol-extension default, not an `AIRegistrationProvider` requirement. Every call site holds the existential type (AIConversation.registrationProvider, ChatBroker sub.registrationProvider), so `provider.registrationProviderRequestRegistration(for: vendor, ...)` binds to this default, which discards `vendor` and calls the no-arg requirement; the concrete conformers then re-inject LLMMetadata.effectiveVendor. Result: a chat on a non-default-vendor model (e.g. Anthropic while the global vendor is OpenAI) authenticates with the wrong vendor's key and shows the wrong signup URLs, and AITermController.registration's new isValid(for:) gate then rejects the mismatched registration and re-requests. Fix: declare the `for vendor:` overload as a requirement in the AIRegistrationProvider protocol so it dispatches through the witness table.

gnachman (commented):
Automated review (medium effort, single-vote verify) of the AI chat provider/model UX work. The two HIGH items below (vendor routing in registration, and the per-vendor key fallback) look blocking to me; the rest are medium or polish. One process note: per the repo guidance, the ChatAgent transcript change needs tools/run_ai_live.sh (anthropic + openai) before merge, since unit tests cannot cover it.

sources/AITerm/AITermControllerObjC.swift
```diff
@@ -18,44 +18,144 @@ class AITermControllerObjC: NSObject, AITermControllerDelegate, iTermObject {
...
+
+        let account = keychainAccount(for: vendor)
+        var value = try? SSKeychain.password(forService: keychainService,
+                                             account: account)
+        if let stored = value, !key(stored, matches: vendor) {
+            value = nil
+        }
+        if keyIsEmpty(value),
+           account != legacyKeychainAccount,
+           let legacy = try? SSKeychain.password(forService: keychainService,
```
gnachman, line 91: HIGH: cross-vendor key leak. When a vendor's keychain slot is empty this falls back to the legacy OpenAI account and adopts it whenever key(legacy, matches: vendor) is true, then persists it into that vendor's account. But key(_:matches:) (line 144) is prefix-only: .deepSeek accepts 'anything not sk-ant- and not AIza', and .llama/.apple return true unconditionally, so a real OpenAI key matches. A user with only an OpenAI key who switches a chat to DeepSeek will have their OpenAI secret returned, written into the DeepSeek keychain account, and sent to DeepSeek's endpoint. Suggest gating the legacy-account fallback to .openAI only (whose account is the legacy account anyway).

sources/AITerm/ChatViewController.swift
```diff
@@ -478,8 +475,144 @@ extension Message {
...
+        }
+        for message in Array(messages).reversed() {
+            if let result = model(named: message.configuration?.model) {
+                return result
+            }
+        }
+        return nil
+    }
+
+    private var effectiveChatModel: AIMetadata.Model? {
```
gnachman, line 516: MEDIUM: a chat whose saved modelName was removed in this PR (claude-opus-4-0/4-1, claude-sonnet-4-0, gemini-3-pro-preview, gemini-2.0-flash[-lite]) resolves model(named:) to nil and silently falls through to the global default model here, which may be a different vendor. ensureCurrentChatHasModel() only writes when modelName == nil, so the dead name is never repaired and the chat silently changes provider with no UI indication. Consider a successor mapping, or repairing the stored name to effectiveChatModel on load.

sources/AITerm/ChatViewController.swift
```diff
@@ -1958,6 +2149,336 @@ fileprivate enum PickSessionButtonIdentifier: String {
...
+        } else {
+            return nil
+        }
+    default:
+        return nil
+    }
+    return CommandTranscriptMarkdown(raw)
+}
+
+fileprivate func CommandTranscriptMarkdown(_ raw: String) -> String? {
```
gnachman, line 2188: MEDIUM: this reparses display text back into structure by sniffing a 'Ran ' prefix plus a newline. Any normal .markdown/.plainText agent message that begins with 'Ran ...' followed by a blank line (e.g. 'Ran the full suite' then a blank line then 'All tests passed') gets re-rendered as a fake command + code block. The structured form already exists on .terminalCommand; suggest applying this only to content that is actually a command transcript rather than pattern-matching arbitrary prose.

sources/Settings/GeneralPreferencesViewController.m
```diff
@@ -1070,11 +1315,283 @@ - (BOOL)canCustomizeAPI {
...
+    if ([modelName containsString:@"llama"] || [host containsString:@"localhost"]) {
+        return iTermAIVendorLlama;
+    }
+    return iTermAIVendorOpenAI;
+}
+
+- (NSString *)defaultAIModelTitleForManualConfiguration:(NSDictionary *)configuration {
+    NSString *name = configuration[kAIManualModelNameKey] ?: @"Untitled model";
+    iTermAIVendor provider = [self providerForManualAIModelConfiguration:configuration];
+    return [NSString stringWithFormat:@"Manual: %@ — %@",
```
gnachman, line 1415: Convention (CLAUDE.md): 'Do not use em dashes ... anywhere that might be shared with third parties.' This user-visible model title uses U+2014. Same on lines 2467 and 2471 of this file. Please replace with a hyphen.

sources/AITerm/ChatSplitViewController.swift
```diff
@@ -73,24 +106,73 @@ class ChatSplitViewController: NSSplitViewController {
...
+            self?.updateSidebarToggleButton()
+            self?.scheduleSidebarToggleButtonRelayouts()
         })
     }
+
+    @objc private func chatContentViewFrameDidChange(_ notification: Notification) {
+        layoutSidebarToggleButton()
+    }
+
+    private func scheduleSidebarToggleButtonRelayouts() {
```
gnachman, line 139: Polish/altitude: this positions the floating toggle button via a four-deadline dispatch-after cascade [0, 0.05, 0.15, 0.35] guarded by a generation counter, on top of an autoresizing mask, viewDidLayout, and frame/bounds observers - four redundant relayout paths, and the delay list is a 'window has settled' guess with no real signal (a slow full-screen transition past 0.35s lands the button at the wrong Y). Suggest pinning the button to chatViewController.view.safeAreaLayoutGuide (or using a toolbar item) and deleting the timer cascade.

sources/AITerm/RegularMessageCellView.swift
```diff
@@ -11,23 +11,175 @@
 class MessageTextView: NSTextView {}
 class MessageTimestamp: NSTextField {}
 
+final class ChatCodeBlockView: NSView {
```
gnachman, line 14: Reuse: ChatCodeBlockView re-implements the code-block view MultipartMessageCellView already has (makeCodeBlockHeader / makeCodeBlockContainer / updateCodeBlock*Colors), including a near-identical copy-button + SFSymbol fallback construction, and measureCodeBlock is duplicated again in TerminalCommandMessageCellView. The two already diverge (corner radius 16 vs 6, padding 16/12 vs 8) and will drift. Consider promoting one shared code-block view + measurement helper. Minor: the fallbackImageName: 'document.on.document' literal is an SF Symbol name; CLAUDE.md asks for SFSymbol.documentOnDocument.rawValue instead of the literal.

sources/AITerm/ChatAgent.swift
```diff
@@ -338,11 +339,7 @@ class ChatAgent {
                 break
             }
         }
-        // Pair any orphaned tool_result with a synthesized tool_use so the
-        // vendor doesn't reject the rebuilt prompt. Heals conversations that
-        // were serialized before this fix as well as new ones.
-        conversation.messages = AIChatToolCallRepair.repairingOrphanedToolResults(
-            translate(messages: messages))
+        conversation.messages = translate(messages: messages)
```
gnachman, line 342: Risk to verify before merge: this removes AIChatToolCallRepair.repairingOrphanedToolResults (which existed to stop Anthropic 400-ing on orphaned tool_result after a reload) and replaces structured function_call/function_output replay with free-text 'iTerm2 tool transcript' prose for every vendor. Anthropic/Gemini/DeepSeek replay tool_use/tool_result fine, so the prose flattening is (a) paid by all vendors for an OpenAI-reasoning-only constraint and (b) can be re-read by the model as its own prior words. Per CLAUDE.md this is exactly the change that needs tools/run_ai_live.sh (anthropic + openai), since unit tests cannot cover it. Suggest branching the down-conversion to the OpenAI/Responses path only, or persisting the reasoning item so structured replay stays valid.

gnachman:
Landed on master as bad0e75f8 (cherry-picked with your authorship credited via a Co-authored-by trailer). The changes were integrated on top of the new data-driven model catalog: the per-model reasoning-effort and service-tier metadata now flows through ai-models.json / AIModelCatalog. Thanks!

theanotherwise:
<3

PR #681, opened 2026-06-09
Use directional labels for split actions

szymeo:
## Problem

The current split menu labels describe the orientation of the divider rather than the direction where the new pane appears:

 - “Split Vertically” creates a new pane to the right
 - “Split Horizontally” creates a new pane below

This is technically consistent with split-view terminology, but it’s easy to misread in day-to-day use. As a user, I think of the action spatially: “put a terminal to my right” or “put one below this pane.” The current labels make me pause and mentally translate “vertical split” into “right” and “horizontal split” into “down.”

Other terminal/editor UIs often use directional wording for this reason: Split Right, Split Down, etc. This PR makes the menu labels match the user’s mental model and also exposes the existing “before” split behavior as Split Left and Split Up, giving the Shell menu a complete directional set:

 - Split Right
 - Split Left
 - Split Down
 - Split Up

## Summary
- Rename visible split menu labels from orientation-based wording to directional wording: Split Right, Split Left, Split Down, and Split Up.
- Wire new left/up menu actions to existing `before:` split support while preserving existing horizontal/vertical selectors as compatibility aliases.
- Update related split labels in key binding/profile UI and Python menu ID docs, keeping existing identifiers where possible for script compatibility.

## Validation
- `git diff --check`
- `xmllint --noout sources/MainMenu/MainMenu.xib sources/Settings/ProfilesWindow.xib sources/SplitPanel/SplitPanel.xib`
- source-level static check for split menu labels/actions/icon map/API constants
- `python3 -m py_compile api/library/python/iterm2/iterm2/mainmenu.py`

Could not run `xcodebuild` or `ibtool` locally because the active developer directory is Command Line Tools, not a full Xcode install:

```
xcode-select: error: tool 'xcodebuild' requires Xcode, but active developer directory '/Library/Developer/CommandLineTools' is a command line tools instance\n```


ablaszkiewicz:
Bump

gnachman:
Thank you for taking the time to put this together, and for the thoughtful writeup of the motivation. I understand the appeal of directional wording, but I'm going to pass on this one.

The horizontal/vertical split vocabulary is used in a lot of places beyond these menu items: key binding action names, the Python API menu identifiers and constants, AppleScript, documentation, and longstanding user muscle memory. Renaming the visible labels while keeping the old identifiers underneath (as this PR does for compatibility) creates a lasting mismatch where the label says one thing and the identifier/API says another, which I think causes more long-term confusion than it removes.

The menu item icons already disambiguate the direction visually (the square.split glyphs show where the new pane lands), which addresses the core readability concern without splitting the terminology.

I really do appreciate the contribution and the care you put into preserving script compatibility. Please don't let this discourage you from sending other PRs.

PR #682, opened 2026-06-11
Keeper: Nested Shared Folder support and add-account toggles

pkamble-ks:
## Summary

Extends the Keeper Security password manager integration to support **Nested Shared Folders (NSF)** alongside the classic vault.

- **Listing:** Merges classic vault records (`list --format=json`) with NSF records (`nsf-list --records --format=json`). Entries are labeled **(Classic)** or **(Nested)**.
- **Mutations:** Add, edit password, and delete route to the correct Commander verbs based on record type (`record-*` / `rm` vs `nsf-record-*` / `nsf-rm`). Account IDs use internal `classic:` / `nested:` prefixes so routing survives the host protocol; Commander still receives plain UIDs.
- **Add account:** New records default to Nested Shared Folders. A **Use classic permission model** checkbox (declared in the adapter handshake) adds to the classic vault when checked.
- **Protocol:** Adapters can declare optional `addAccountToggles` in the handshake; the Password Manager Add Account panel renders them as checkboxes and passes `flags` on add.
- **Reliability:** Login validates with a lightweight list call instead of blocking on `sync-down`. Connectivity and HTML proxy errors surface actionable messages.

Commander Service Mode must allow: `list`, `get`, `sync-down`, `record-add`, `record-update`, `rm`, `nsf-list`, `nsf-record-add`, `nsf-record-update`, `nsf-rm`.

## Test plan

- [x] `cd pwmplugin && swift test --filter KeeperIntegrationTests` — 21/21 passed
- [ ] Manual: list classic and nested records; confirm **(Classic)** / **(Nested)** labels
- [ ] Manual: fill password and enter username for both record types
- [ ] Manual: edit password on classic and nested records
- [ ] Manual: add record with toggle off (nested, default) and on (classic)
- [ ] Manual: delete classic and nested records; confirm list refreshes after Sync Down
- [ ] Manual: login with unreachable URL or bad API key — confirm actionable error message


sources/PasswordManager/iTermPasswordManager.xib
```diff
@@ -414,14 +416,14 @@ Gw
         <window title="New Account" allowsToolTipsWhenApplicationIsInactive="NO" autorecalculatesKeyViewLoop="NO" restorable="NO" releasedWhenClosed="NO" visibleAtLaunch="NO" frameAutosaveName="" animationBehavior="default" id="5nJ-MG-kpS" customClass="NSPanel">
             <windowStyleMask key="styleMask" titled="YES"/>
             <windowPositionMask key="initialPositionMask" leftStrut="YES" rightStrut="YES" topStrut="YES" bottomStrut="YES"/>
-            <rect key="contentRect" x="344" y="373" width="330" height="136"/>
+            <rect key="contentRect" x="344" y="373" width="330" height="196"/>
```
gnachman, line 423: The New Account panel height is bumped from 136 to 196 for **all** password managers, but `configureFirstAddAccountToggle` only toggles `hidden` on the checkbox/label — nothing shrinks the panel or repositions the OK/Cancel buttons when there's no toggle. Every adapter that doesn't declare an add-account toggle (Keychain, 1Password, Bitwarden, LastPass, KeePassXC) will now show ~60px of empty space between the Password field and the buttons. Could the panel resize (or the controls collapse) when `firstAddAccountToggleDescription` is nil so the non-toggle case looks the same as before?

sources/PasswordManager/AdapterPasswordDataSource.swift
```diff
@@ -337,18 +339,53 @@ class AdapterPasswordDataSource: CommandLinePasswordDataSource {
...
+        keychainSettingsCache.removeAll()
+        keychainSettingsCacheCheckedKeys.removeAll()
+    }
+
     private func storedSettingsValue(forKey key: String) -> String? {
         guard let fields = handshakeInfo?.settingsFields else { return nil }
         guard let field = fields.first(where: { $0.key == key }) else { return nil }
         if field.persistInKeychain {
-            return try? SSKeychain.password(forService: keychainCredentialServiceName, account: key)
+            if keychainSettingsCacheCheckedKeys.contains(key) {
```
gnachman: Two concerns with this keychain cache:

1. It's never invalidated across instances. `PasswordManagerDataSourceProvider.forTerminal` and `.forBrowser` each hold their own process-lifetime `_keeper`, but they share one keychain entry. `invalidateKeychainSettingsCache` only runs on the instance that did the write (via `storeSettingsValue`), so if the API key is updated on one side, the other keeps serving the stale value for the rest of the session.
2. Misses are cached too (the key goes into `keychainSettingsCacheCheckedKeys` even when the value is nil), so a key read before it's populated is pinned to nil on that instance.

Net effect is a stale/empty credential that survives until the app restarts. Was there a measured cost to the uncached `SSKeychain.password` reads that motivated the cache? If so it may be worth keying invalidation off the actual keychain rather than only local writes.

sources/PasswordManager/AdapterPasswordDataSource.swift
```diff
@@ -337,18 +339,53 @@ class AdapterPasswordDataSource: CommandLinePasswordDataSource {
                 pathToDatabase = u
             }
         }
-        if masterPassword == nil {
-            masterPassword = loadPersistedCredentials()
+        guard masterPassword == nil else { return }
+
+        if let secretFromSettings = persistedSecretSettingsValue(),
```
gnachman: This changes credential precedence: `persistedSecretSettingsValue()` now wins over `loadPersistedCredentials()`. They read different keychain entries — the secret settings field is stored under `account: <fieldKey>` (e.g. `apiKey`), whereas a successful login persists the working credential under `account: identifier`. So if the value typed into the settings field ever diverges from the credential that actually logged in, the (possibly stale) settings value now shadows the known-good one and auto-login can fail where it previously succeeded. Is the precedence intentional? Falling back to `loadPersistedCredentials()` first (or only using the settings value when there's no persisted credential) would preserve the old behavior.

pwmplugin/Sources/iterm2-keeper-adapter/KeeperCommanderClient.swift
```diff
@@ -450,27 +643,76 @@ private func extractPasswordFromRawGet(data: Data) -> String? {
...
+    var lastResponse: Data?
+    var lastNetworkError: Error?
+    for verb in verbs {
+        do {
+            let result = try attempt(verb)
+            if result.success {
+                KeeperAdapterLog.write("setPassword: success uid=\(uid) via \(verb)")
+                return
+            }
+            lastResponse = result.raw
```
gnachman: In the `.none` source case (any accountID without a `classic:`/`nested:` prefix — i.e. records already stored by existing users before this change), `setPassword`/`deleteRecord` try the first verb then the second, and `lastResponse` is overwritten unconditionally. So when the first verb fails for a real reason (e.g. a genuine permission error on a classic record), the code then runs the nsf verb on the same UID and the user is shown the *nsf* error ("not found in any nested shared folder") instead of the real one. Consider preserving the first meaningful error, or only falling through to the second verb when the first failure looks like a wrong-record-type/not-found signal.

pwmplugin/Sources/iterm2-keeper-adapter/KeeperCommanderClient.swift
```diff
@@ -483,36 +725,109 @@ private func escapeForKeeperDoubleQuotedCommandField(_ s: String) -> String {
...
     }
     struct RecordAddData: Decodable {
         let record_uid: String?
+        let uid: String?
+        var effectiveUid: String? { record_uid ?? uid }
     }
     let response = try JSONDecoder().decode(RecordAddResponse.self, from: data)
-    guard response.status == "success", let uid = response.data?.record_uid, !uid.isEmpty else {
+    let extractedUid: String? = response.data?.effectiveUid
+        ?? response.message?.trimmingCharacters(in: .whitespacesAndNewlines)
```
gnachman: `addRecord` falls back to using the response `message` field as the record UID, and unlike the get/set/delete paths it doesn't run the result through `validatedRecordUID`. If a Commander response ever puts something other than a bare UID in `message`, the returned identifier becomes `nested:<that text>` and later get/set/delete on it will fail validation (the add appears to succeed but the record can't be operated on). Could this parse the UID only from a structured field and validate it before returning?

pwmplugin/Sources/iterm2-keeper-adapter/KeeperCommanderClient.swift
```diff
@@ -450,27 +643,76 @@ private func extractPasswordFromRawGet(data: Data) -> String? {
...
-    let b64 = Data(newPassword.utf8).base64EncodedString()
-    let cmd = "record-update -r \(uid) password=$BASE64:\(b64)"
-    let data = try client.executeCommand(apiKey: apiKey, command: cmd)
-    if let response = try? JSONDecoder().decode(KeeperExecuteResponse.self, from: data), response.status == "success" {
-        return
+    let parsed = parseAccountIdentifier(recordUid)
+    let uid = try validatedRecordUID(parsed.uid)
+
+    func attempt(_ verb: String) throws -> (success: Bool, raw: Data) {
+        let cmd = "\(verb) --force -r \(uid) \(passwordCommandFragment(password: newPassword))"
```
gnachman: Just confirming `--force` is intentional/accepted on `record-update`/`nsf-record-update` (and `record-add`/`nsf-record-add` at line 791)? The test mock matches commands by prefix, so the flag isn't actually exercised — worth a note in the PR that it's been confirmed against a live Commander.

pwmplugin/Sources/iterm2-keeper-adapter/KeeperAdapterLog.swift
```diff
@@ -0,0 +1,44 @@
...
+            .urls(for: .libraryDirectory, in: .userDomainMask)
+            .first?
+            .appendingPathComponent("Logs")
+        let dir = logsDir ?? URL(fileURLWithPath: NSTemporaryDirectory())
+        try? FileManager.default.createDirectory(at: dir, withIntermediateDirectories: true)
+        return dir.appendingPathComponent("iterm2-keeper-adapter.log")
+    }()
+
+    static let isEnabled: Bool = {
+        ProcessInfo.processInfo.environment["ITERM2_KEEPER_ADAPTER_LOG_DISABLED"] == nil
```
gnachman: This logger is enabled by default and writes to a persistent file at `~/Library/Logs/iterm2-keeper-adapter.log`. Passwords are redacted, but account titles and usernames are written in plaintext. For a password-manager integration it'd be safer to default this off (opt-in via the env var) rather than opt-out.

docs/notes-3.7.txt
```diff
@@ -369,6 +369,18 @@ Python API Improvements:
...
+- The Keeper adapter supports Nested Shared
+  Folders as well as the classic vault.
+  Both appear in one list, labeled (Classic)
+  or (Nested). Add, edit, and delete work
+  for either type.
+- Password manager adapters can declare
+  optional toggles in their handshake that the
+  Add Account panel renders as checkboxes. The
+  Keeper adapter uses this to offer a “Use
+  classic permission model” choice — leaving
```
gnachman: Minor: this line uses an em dash, which the project style avoids in user-facing text. Could swap it for " - " or reword.

gnachman (commented):
Thanks for this — nested-folder support is a nice addition. A few things to look at, mostly on the iTerm2 host-side integration (the password-manager UI and credential plumbing shared with the other adapters). Inline comments below.

sources/PasswordManager/iTermPasswordManager.xib
pkamble-ks: Fixed. Reverted the xib default to 136px and only expand the panel to 196px in `configureFirstAddAccountToggle` when the adapter declares a toggle. Adapters without a toggle (Bitwarden, KeePassXC, etc.) should look the same as before.

sources/PasswordManager/AdapterPasswordDataSource.swift
pkamble-ks: Removed the keychain settings cache entirely. `storedSettingsValue` reads from the keychain directly again, so terminal/browser instances and keychain updates no longer go stale for the rest of the session.

sources/PasswordManager/AdapterPasswordDataSource.swift
pkamble-ks: Unintentional regression — fixed. 
`loadPersistedCredentials()` (login credential under `identifier`) is tried first; the settings keychain value is only used as a fallback when there is no persisted login credential.

pwmplugin/Sources/iterm2-keeper-adapter/KeeperCommanderClient.swift
pkamble-ks: Fixed. For the `.none` source case, the second verb is only tried when the first failure looks like not-found / wrong record type (`keeperCommandFailureLooksLikeWrongRecordType`). Otherwise the first error is surfaced to the user.

pwmplugin/Sources/iterm2-keeper-adapter/KeeperCommanderClient.swift
pkamble-ks: Fixed. The returned UID must pass `validatedRecordUID`. Structured `data.record_uid` / `data.uid` is preferred; `message` is only used as a fallback when it validates as a bare UID.

pwmplugin/Sources/iterm2-keeper-adapter/KeeperCommanderClient.swift
pkamble-ks: Yes, intentional. Confirmed against live Commander Service Mode: classic `record-add `failed without `--force` under password policy, and updates needed it as well. The integration mock matches commands by prefix so it does not assert the flag explicitly.

docs/notes-3.7.txt
pkamble-ks: Fixed — replaced the em dash with a period per project style.

pwmplugin/Sources/iterm2-keeper-adapter/KeeperAdapterLog.swift
pkamble-ks: Changed to opt-in: logging is off by default and enabled with `ITERM2_KEEPER_ADAPTER_LOG_ENABLED=1`.

pkamble-ks:
@gnachman Thanks for the review, pushed fixes addressing all inline comments (commit ae7e41d, now on this branch via fork merge).

Summary:
- New Account panel: default 136px, expands to 196px only when adapter declares a toggle
- Removed keychain settings cache; restored login-credential precedence
- Unprefixed account IDs: second verb only on not-found/wrong-type errors
- `addRecord` validates UID before returning
- Debug logging opt-in via `ITERM2_KEEPER_ADAPTER_LOG_ENABLED=1`
- Release notes punctuation fix
- `--force` on add/update confirmed against live Commander Service Mode

Tests still 21/21. Happy to adjust anything else.

gnachman:
## Review notes

Reviewed the diff plus the PR head blob (`pkamble-ks/iTerm2@63cda37`) to check the credential path. Findings below, most severe first. Several correctness items hinge on real Keeper Commander output shapes that the mock server encodes as assumptions, so they are worth confirming against a live Commander before merge.

### Correctness

1. **Editing the API key in Settings is silently ignored; the stale key is used forever.** `AdapterPasswordDataSource.setSettingsValue` nils `masterPassword`/`authToken` when a settings field changes but never calls `deletePersistedCredentials()`. The old key was persisted under keychain account = `identifier` at login, while the new key is stored under account = `apiKey`. On the next auth, `hydratePersistedCredentialsIfNeeded` sees `masterPassword == nil`, `loadPersistedCredentials()` returns the **old** key and returns early, so `persistedSecretSettingsValue()` (the new key) is never consulted. After a user updates a rejected or rotated key, the adapter keeps authenticating with the old one.

2. **An NSF record that also appears in `list` can become permanently uneditable.** In `listAccountsRecords`, the merge inserts the `list` copy first and skips the `nsf-list` copy for the same UID, so the record is labeled `classic:` (the mock has `list` returning `record_category:"Classic"` even for `shared:true` rows). For a known source, `setPassword`/`deleteRecord` use a **single** verb with no cross-type fallback (`.classic -> ["record-update"]`). If that record is only mutable via `nsf-record-update`/`nsf-rm`, every edit and delete fails with no recovery.

3. **NSF folders can leak into the account list as bogus credentials.** `taggedAsNested` rebuilds each record through an init that hardcodes `itemType = nil`, discarding the `"Item Type"` signal. `isFolder` uses `itemType == "folder"` as one of its two folder tests, so an `nsf-list --records` folder row identified only by `"Item Type":"Folder"` (no lowercase `"Type"`, so `effectiveType` is nil) flips to `isFolder == false` after the rebuild and is emitted as an account with an empty username.

4. **A successful nested add can report "Add failed".** The nested path reads the new UID from `response.message` and runs the entire string through `validatedRecordUID` (`^[A-Za-z0-9_-]{15,}$`). The mock returns a bare UID, but if real `nsf-record-add` returns a sentence (e.g. `"Record NSFUID... added"`), the whole-string regex rejects it and the `else` branch throws `"Add failed"` even though the record was created, orphaning it and producing a duplicate on retry.

5. **Legacy (unprefixed) records rely on a fragile English-substring heuristic to route.** For `source == .none` (pre-existing bare UIDs), the classic -> nested fallback only fires when the failure text matches a fixed list (`"not found"`, `"nested shared folder"`, ...). A nested record whose `record-update` rejection is phrased differently (e.g. `"permission denied"`, or a localized message) fails `keeperCommandFailureLooksLikeWrongRecordType`, breaks the loop, and never tries `nsf-record-update`, so the operation permanently fails.

### Cleanup / efficiency

6. **`setPassword` and `deleteRecord` duplicate ~55 lines of verb-fallback machinery** (the `attempt` closure, the per-source verb array built by an identical `switch parsed.source`, the `index == 0 && count > 1` wrong-type `continue`, and the `lastResponse`/`lastNetworkError` unwind). They already differ subtly (one surfaces `keeperUserFacingPasswordUpdateError`, the other the raw string). Extracting one shared fallback runner would also let items 2 and 5 be fixed in a single place.

7. **Three sequential blocking round-trips per list refresh.** Every refresh runs `sync-down`, then `list --format=json`, then `nsf-list --records --format=json` as separate POST-then-poll cycles (they cannot overlap), where the old code used a single `ls -R -l`. This triples worst-case latency and the HTTP-429 exposure that the deleted comment ("Do not call `get` per record ... Commander Service Mode often returns HTTP 429") specifically warned about.

### Conventions

8. Two `CLAUDE.md` violations:
   - `IntegrationTests.swift`: `var env = ProcessInfo.processInfo.environment; process.environment = env` is never mutated, which triggers the Swift "never mutated; consider `let`" warning ("treat warnings as errors").
   - Five new log strings use em dashes (`setPassword: FAILED uid=... — ...`, `deleteRecord: FAILED ...`, and three `no stdin input — aborting` lines), against the "no em dashes in strings ... shared with third parties" rule (debug logs are shareable).

### Cleared
Release-notes line widths (all under 50), no `fatalError`/`assert` added, the HTML-detection helper is string matching (not the inline-HTML rule), and `KeeperAdapterLog.swift` is in the SwiftPM target dir so no xcodeproj step is needed.

pkamble-ks:
@gnachman Thanks for the detailed second pass — pushed fixes in b149796 (merged to master via PR #4). Summary by item:

### Correctness

- API key in Settings: `setSettingsValue` now calls `deletePersistedCredentials()` when a `persistInKeychain` settings field changes, so the old login keychain entry cannot override the updated API key.
- Duplicate UID merge: `nsf-list` now wins when the same UID appears in both `list` and `nsf-list`, so records route as (Nested) / `nested:` instead of (Classic).
- NSF folders in list: `taggedAsNested` preserves `itemType` via `withRecordCategory()`, so `"Item Type":"Folder"` rows stay filtered out.
- NSF add “Add failed”: added `extractRecordUID(from:)` to pull the UID from sentence-style Commander `message` strings before validation.
- Unprefixed ID fallback: replaced the English-substring heuristic with a shared `runKeeperMutationAttempts` helper; for `source == .none`, both verbs are tried and the first failure is surfaced if both fail.

### Cleanup / efficiency

- Deduped mutation paths: `setPassword` and `deleteRecord` share the same fallback runner.
- List latency: removed automatic `sync-down` from the default list path (`syncFirst` defaults to `false`); Sync Down remains on the explicit custom command.

### Conventions

- `var env` → `let env` in `IntegrationTests.swift`.
- Em dashes removed from new log strings in `main.swift` and `KeeperCommanderClient.swift`.

### Add Account panel (from live testing)

- Fixed layout after round 1 sizing: fields were clipped and the checkbox overlapped User Name when the sheet opened. Panel now uses explicit compact (136) vs expanded (196) field positions with fixed frames re-applied after `beginSheet`.

### Tests & live validation

- Integration tests now 23/23 (added duplicate-UID merge, NSF folder exclusion, and sentence-style NSF add message cases).
- Validated against live Commander: list, add (classic + nested), set password, delete, Sync Down, and API key change.

Happy to adjust anything else.

gnachman:
Thanks for the substantial work here. I went through it in depth. Most of my feedback is architectural, with a batch of concrete bugs at the end. The recurring theme is Keeper-specific concepts leaking into shared password-manager infrastructure, and I would like to see those pulled back before this is ready for line-by-line review.

## 1. The API key is modeled as both a master password and a settings field

The handshake sets `requiresMasterPassword: true` with `masterPasswordLabel: "API key"` AND declares an `apiKey` settings field (`isSecret: true, persistInKeychain: true`). So the same secret has two entry points and two keychain items (account = adapter identifier vs account = "apiKey"). Almost all of the credential complexity in this PR is downstream cleanup for that one decision:

- `persistedSecretSettingsValue()` exists only to let the Settings copy seed `masterPassword` when the prompt copy is absent.
- `hydratePersistedCredentialsIfNeeded()` prefers `loadPersistedCredentials()` (account = identifier) before the settings copy, so once you authenticate once, editing the API key in Settings silently has no effect (the two items diverge and the identifier-keyed copy wins).
- The `standardHeader` change (`persistInKeychain && credentialAlreadyAvailable -> omit`) is another patch on the same overload.

`masterPassword` alone is sufficient: the prompt already collects the key, `persistsCredentials: true` stores it, and it re-hydrates on launch. Please pick one source of truth: keep the master-password prompt (labeled "API key") and drop the `apiKey` settings field, keeping only `serviceURL`. That deletes `persistedSecretSettingsValue()`, removes the divergence/staleness, and removes the need for the header omission. The only thing the settings field uniquely offered was editing the key later without an auth failure, which is better handled by a "clear credentials / re-enter" affordance than a duplicate secret.

If a future adapter genuinely needs a login-only settings secret, add an explicit `omitIfAuthenticated: Bool` (default false) to `SettingsField` and gate the header omission on that, rather than overloading `persistInKeychain`. Not needed for Keeper if the field is dropped.

## 2. Do not bake the (Classic)/(Nested) label into accountName

`AdapterPasswordDataSource` returns each list entry's `accountName` as `displayTitleWithSource` ("Example (Nested)"), but the add path returns the raw typed name. The host matches accounts by exact `[accountName isEqualToString:]`, so this regresses generic host behavior:

- After add, `didAddAccount` never finds the new row, so it is not selected.
- The pre-add duplicate guard in `reallyAdd` never matches, so duplicate accounts can be created.

Please model record source as a first-class field on the account and let the host format the label. Keep `accountName` a stable identity.

## 3. Keep classic:/nested: routing internal to the adapter

The `classic:`/`nested:` account-id prefixes are threaded through the host's shared `AccountIdentifier`, and the classic-vs-nested split is re-parsed from a colon-prefixed string in several places rather than modeled. Even if it is meant to be opaque, this is Keeper routing living in shared types. Prefer keeping the routing fully internal to the adapter.

## 4. Add-account toggles and flags-add belong on the base data source, not AdapterCapabilities

Right now the flags-carrying `add(...)` and `addAccountToggleDescriptions` live on `AdapterCapabilities`, so `iTermPasswordManagerWindowController` has to `conformsToProtocol:@protocol(iTermAdapterCapabilities)` and downcast to render the toggles and again to call the flags-add (via `respondsToSelector:`, which silently falls back to the flags-less add and drops the user's toggle choice). Adding an account is a generic operation, so the generic Add Account UI should not know what an adapter is.

Please move `addAccountToggleDescriptions` and the flags-carrying `add(...)` onto the base `PasswordManagerDataSource`, with trivial defaults for sources that have no toggles (toggle descriptions -> nil, flags-add -> forward to the plain add ignoring flags). The controller then only ever talks to `PasswordManagerDataSource`, all the `iTermAdapterCapabilities` downcasting and the `respondsToSelector:` fallback in the add path go away, and the silent-drop bug goes with them. The adapter Settings-sheet capabilities (`settingsValue`, `settingsFieldDescriptions`, `showAdapterSettingsSheet`) are genuinely adapter-specific and can stay on `AdapterCapabilities`.

## 5. Correctness bugs

1. Classic records from plain `list` have no category/source, so `sourceLabel == nil`, so they get a bare identifier and no "(Classic)" label. The same record added via the classic toggle gets `classic:UID`. The identifier churns across refreshes and the promised label never shows for classic records.
2. Login validation switched to `list` under a roughly 20s budget: a slow/cold Commander or a large vault will report a login failure with valid credentials.
3. The `tryAll` fallback does `catch { networkError = error; break }`, so it aborts on the first thrown error and never tries the alternate verb, defeating the classic/nested fallback for unprefixed ids.
4. `addRecord` decodes `data` as a non-optional object; if `nsf-record-add` returns `data` as a JSON string (as the get paths show Commander sometimes does) it throws typeMismatch and reports a created record as a failure.
5. The merge routes any UID present in both `list` and `nsf-list` to the nested verbs (nsf overwrites list). A genuinely classic record that also surfaces in `nsf-list` then gets `nsf-record-update`/`nsf-rm`, which Commander rejects.
6. `KeeperAdapterLog` writes raw, unredacted Commander response bodies to a file in `~/Library/Logs` (env-gated off by default, and command passwords are redacted). For a password manager, response bodies can contain secrets. Please reconsider logging response bodies at all.

## 6. Question: cache invalidation after set-password

`invalidateListAccountsCache()` is called after add, delete, and set-password. Add and delete clearly change the account set, but the cached list holds only `Account(identifier, userName, accountName, hasOTP)` (no password), and `SetPasswordRequest` carries only `accountIdentifier` and `newPassword`, so a password edit cannot change any cached field. Can you confirm the set-password invalidation is necessary and explain why? If not, dropping it avoids a full Commander round-trip (sync + list + nsf-list) after every password edit.

## 7. Cleanup

- `sourceLabel` duplicates two near-identical record_category-vs-source switch blocks, and `setPassword`/`deleteRecord` duplicate the same three-case routing switch. Parameterize into one helper each.
- `configureNewAccountPanelFieldLayoutShowingToggle:` is two parallel branches of about 18 absolute magic Y values kept in sync by hand. They are all derivable: each label = field + 3, the generate button = field - 3, each row steps by 27, and the two branches differ only by the starting Y (`panelHeight - 34`). It collapses to a single cascade (behavior-identical, verified against the current constants):

```objc
- (void)configureNewAccountPanelFieldLayoutShowingToggle:(BOOL)showingToggle {
    const CGFloat fieldHeight = 27;
    const CGFloat labelOffset = 3;
    const CGFloat topMargin = 34;
    const CGFloat toggleGap = 5;
    const CGFloat panelHeight = showingToggle ? kNewAccountPanelHeightWithToggle
                                              : kNewAccountPanelHeightWithoutToggle;

    CGFloat y = panelHeight - topMargin;
    iTermFixViewY(_newAccount, y);
    iTermFixViewY(_newAccountLabel, y + labelOffset);
    y -= fieldHeight;

    iTermFixViewY(_newUserName, y);
    iTermFixViewY(_newUserNameLabel, y + labelOffset);
    y -= fieldHeight;

    iTermFixViewY(_newAccountPassword, y);
    iTermFixViewY(_newPasswordLabel, y + labelOffset);
    iTermFixViewY(_generatePasswordButton, y - labelOffset);

    if (!showingToggle) {
        return;
    }
    y -= fieldHeight + toggleGap;
    iTermFixViewY(_addAccountToggleCheckbox, y);
    y -= fieldHeight;
    iTermFixViewY(_addAccountToggleLabel, y);
}
```

pkamble-ks:
@gnachman Thanks for the thorough architectural review — addressed in 40885a1 (on this PR via cec0504).

### 1. API key as master password only
Dropped the Settings `apiKey` field. Handshake keeps `requiresMasterPassword` with `masterPasswordLabel: "API key"` and Settings declares only `serviceURL`. Removed `persistedSecretSettingsValue()` and the related header-omission path. To rotate a key: **Reset Keeper Security Configuration**, then authenticate again via the prompt.

### 2. Stable `accountName` + first-class source
Adapter returns bare titles; added optional `sourceLabel` on the protocol/`PasswordManagerAccount`. The host formats `name (Classic)` / `name (Nested)` for display. Add selection and the duplicate-name guard match on the stable `accountName` again.

### 3. Routing stays adapter-internal
Wire account IDs are bare UIDs (legacy `classic:`/`nested:` prefixes are still stripped if present). Classic vs nested verb choice is internal; set/delete always try classic then nested.

### 4. Toggles / flags-add on `PasswordManagerDataSource`
Moved `addAccountToggleDescriptions` and the flags-carrying `add(...)` onto `PasswordManagerDataSource`. The Add Account UI no longer downcasts to `iTermAdapterCapabilities` for those. Settings-sheet APIs remain on `AdapterCapabilities`.

### 5. Correctness
1. Classic `list` rows get a Classic label via `taggedAsClassic` when category/source is missing.
2. Login validation timeout raised to 120s.
3. `tryAll` continues after thrown errors so the alternate verb still runs.
4. `addRecord` accepts `data` as an object or a JSON string.
5. Duplicate UIDs prefer the classic `list` row over `nsf-list`.
6. Adapter logs response **byte counts** only — no raw bodies.

### 6. Set-password cache invalidation
Confirmed unnecessary and **dropped**. The list cache has no password fields, and set-password does not change identity/username/title, so invalidating forced a needless Commander round-trip.

### 7. Cleanup
Shared `normalizedKeeperSourceLabel` and a single `runKeeperMutationAttempts` helper for set/delete. Panel layout uses the Y-cascade you suggested.

Happy to adjust anything else before line-by-line.

gnachman:
## Review notes

Generally well-structured. The load-bearing risk is the classic-vs-nested mutation routing: the adapter guesses the vault by trying both Commander verbs rather than using the vault it already knows, and the mock server in `IntegrationTests.swift` prefix-matches every verb and returns canned success, so several of the items below are not covered by tests. Findings ranked most-severe first; items that depend on real Keeper Commander CLI behavior are marked PLAUSIBLE.

### 1. `addRecord` default path uses `nsf-record-add` with no fallback
`pwmplugin/Sources/iterm2-keeper-adapter/KeeperCommanderClient.swift` (`addRecord`, ~L809-841)

The toggle defaults to `false` → `verb = "nsf-record-add"`, and unlike `setPassword`/`deleteRecord` (which try classic then nested via `runKeeperMutationAttempts`), `addRecord` issues one verb and never falls back. A user on a plan/Commander build without a Nested Shared Folder who clicks Add with the box unchecked (the default) gets a generic `"Add failed"` and no record, even though classic `record-add` would have worked. (PLAUSIBLE — depends on the vault having an NSF target.)

### 2. Known vault is discarded; nested edit/delete can silently no-op and report success
`KeeperCommanderClient.swift` (`setPassword` ~L740-768, `deleteRecord` ~L783-806)

`listAccountsRecords` emits `AccountIdentifier(accountID: uid)` (bare UID, no `classic:`/`nested:` prefix), so `parseAccountIdentifier(...).source` is always `nil` in production; both mutation paths ignore `source` regardless and unconditionally run `[record-update, nsf-record-update]` / `[rm -f, nsf-rm]`, returning on the first `status == "success"`. If Commander's classic `rm -f <uid>` / `record-update -r <uid>` returns a success envelope for a UID that actually lives in NSF (a no-op), the nested verb never runs and the UI reports success — the nested record is never changed/deleted and reappears on the next list. `testSetPasswordOnNestedRecordUsesNsfRecordUpdate` etc. only pass because they hand-craft `nested:` prefixes production never sends. (PLAUSIBLE — hinges on Commander's success semantics for a wrong-vault verb; the deeper fix is to route on the `sourceLabel` already known at list time.)

### 3. Newly added `--force` on `record-update`/`record-add` is unverified and untested
`KeeperCommanderClient.swift` (`record-update` ~L754, `record-add` ~L818)

Neither verb had `--force` before (only `rm` had `-f`). If Commander rejects `--force` as unknown, `addRecord` breaks for every add (no fallback, both classic and nested get it), and classic `setPassword` silently falls through to `nsf-record-update` against the wrong vault. The mock accepts any prefix, so tests won't catch it. (PLAUSIBLE — worth verifying against the real Commander CLI.)

### 4. Failing `nsf-list` is swallowed; Nested accounts silently vanish
`KeeperCommanderClient.swift` (`listAccountsRecords`, ~L665-670)

Classic `list` throws on failure (surfaced to the user), but `nsf-list` is wrapped in `try?` with only an off-by-default log line on failure. A momentary 429/timeout/HTML-error on `nsf-list` while `list` succeeds renders the panel with Classic records only; a user may conclude their Nested accounts were deleted. (CONFIRMED by code; consider surfacing a partial-failure signal.)

### 5. Mutation error precedence masks network failures
`KeeperCommanderClient.swift` (`runKeeperMutationAttempts`, ~L549-558)

`if let response = firstFailureResponse { throw }` is checked before `if let error = firstNetworkError`, and the loop continues past a first-verb network error. Rotating a classic password where `record-update` commits server-side but the result read times out (throws) → the loop runs `nsf-record-update`, whose `{"error":"record not found"}` is what the user sees, instead of the actionable connectivity message this PR otherwise adds — while the classic write may have actually landed. (CONFIRMED ordering bug; PLAUSIBLE real-world trigger.)

### 6. API key settings field removed with no migration
`pwmplugin/Sources/iterm2-keeper-adapter/main.swift` (handshake ~L83-92, `apiKey(fromHeader:)` ~L107-112)

The `apiKey` settings field and its `header.settings["apiKey"]` read path are removed. Existing users who configured Keeper by typing the key into the secret "API key:" settings field will find it ignored after upgrade (re-prompted as master password), and the old secret is orphaned at keychain `account:"apiKey"` — never read and never cleaned up, even by "Reset Configuration" (which only iterates current `settingsFields`). Worth confirming the migration is intended and whether the stale keychain entry should be purged. (CONFIRMED behavior change.)

### 7. Dead code + triplicated source-label formatting
`KeeperCommanderClient.swift` (`displayTitleWithSource`, ~L352-355)

`displayTitleWithSource` is never called (list emits `displayTitle` + `sourceLabel` separately), and the same `"name (source)"` formatting is duplicated in `CommandLineProvidedAccount.formattedAccountName` and `-accountNameForRow:`. A future maintainer wiring `displayTitleWithSource` into `accountName` would bake the label into the identity string the doc-comment warns must stay stable, producing `"Web (Classic) (Classic)"`. Delete it and hoist one shared formatter. (Lower priority: the parallel `taggedAsClassic`/`taggedAsNested` near-duplicate and the 16-field `withRecordCategory` copy are similar tidy-ups.)

---

The connectivity-error formatting, xib/outlet wiring, protocol-flags plumbing, and the login `validateApiKey` change all traced clean. I'd want findings 1, 2, and 3 confirmed against a real Commander instance before merge.

<sub>Reviewed with Claude Code.</sub>

pkamble-ks:
@gnachman Thanks for the notes — addressed in ad107045 (on this PR via the fork merge). Summary by item:

### 1. NSF default add with no classic fallback
Default add still prefers `nsf-record-add`. If that fails, we now fall back to classic `record-add`. The “Use classic permission model” toggle still forces classic-only.

### 2. Known vault discarded; nested edit/delete false success
Agreed this was the load-bearing risk. Live Commander confirmed classic `record-update` can return `status=success` as a no-op on an NSF UID, so try-both was unsafe.

The host now forwards list `sourceLabel` on set/delete. The adapter routes:
- Classic → `record-update` / `rm` only
- Nested → `nsf-record-update` / `nsf-rm` only
- Unknown set → nested first, then classic
- Unknown delete → classic then nested (`rm` errors on NSF rather than false-succeeding)

Wire IDs stay bare UIDs; legacy `classic:`/`nested:` prefixes are still accepted if present. Integration tests use bare UID + `sourceLabel` (production shape).

### 3. `--force` on update/add
Kept. Verified against live Commander Service Mode — required for password-policy paths. Without it, updates can fail even when the record is writable.

### 4. Failing `nsf-list` swallowed
`nsf-list` failures are no longer silent. Classic `list` still drives the panel; we set `ListAccountsResponse.warning` and the host shows an alert that Nested accounts may be missing until retry / Sync Down.

### 5. Mutation error precedence
`runKeeperMutationAttempts` now prefers the first network/timeout error over a later verb’s body failure, so a timed-out classic write is not masked by a nested “not found”.

### 6. Orphaned Settings `apiKey` keychain entry
On hydrate and Reset Configuration we purge the legacy keychain account `"apiKey"`. Current path remains master-password prompt labeled “API key” + Settings `serviceURL` only. To rotate a key: **Reset Keeper Security Configuration**, then re-authenticate.

### 7. Dead helper / duplicated formatting
Removed unused `displayTitleWithSource`. Host uses one shared formatter for `Account (Classic)` / `Account (Nested)` display; `accountName` stays the stable identity.

### Verification
- `swift test --filter KeeperIntegrationTests` — 26/26
- Manual UI against live Commander: list labels, get/fill, classic + nested add/edit/delete, Sync Down

Happy to adjust anything else before line-by-line.

sources/PasswordManager/iTermPasswordManagerWindowController.m
```diff
@@ -1696,7 +1858,7 @@ - (void)tableView:(NSTableView *)aTableView
         NSString *userName = entry.userName;
         NSString *accountName = entry.accountName;
         if (aTableColumn == _accountNameColumn) {
-            accountName = anObject;
+            accountName = [self accountNameByStrippingSourceSuffix:anObject fromAccount:entry];
```
gnachman, line 1861: Editing an account name or username inline can quietly move a Classic Keeper record into a Nested Shared Folder.

The inline edit in `setObjectValue` is implemented as fetch password, `delete` the old record, then re-add it. The delete carries `sourceLabel`, so it routes to the right vault, but the re-add goes through the plain `addUserName:accountName:password:context:` variant with no flags. Only `reallyAdd:` (the Add button) was updated to pass the toggle flags.

For the Keeper adapter, an add with no flags means `useClassicPermission` defaults to `false`, which takes the `nsf-record-add` path. So renaming a Classic record inline deletes it from the classic vault and recreates it in Nested Shared Folders, changing its permission model without the user asking for it. The edit path probably needs to preserve the record's vault (either by passing a flag derived from `sourceLabel`, or by using a real update rather than delete-then-add).

pwmplugin/Sources/iterm2-keeper-adapter/KeeperCommanderClient.swift
```diff
@@ -482,37 +795,170 @@ private func escapeForKeeperDoubleQuotedCommandField(_ s: String) -> String {
...
+}
+
+func addRecord(apiKey: String,
+               userName: String,
+               accountName: String,
+               password: String?,
+               useClassicPermission: Bool,
+               client: KeeperCommanderClient) throws -> String {
+    let verbs: [String]
+    if useClassicPermission {
```
gnachman, line 876: This fallback can land a credential in the wrong vault (or create a duplicate) against the user's explicit choice.

When the toggle is left unchecked the user is asking for a Nested Shared Folder record, but the loop catches *any* error from `nsf-record-add` and retries with `record-add`. That is fine when NSF is genuinely unavailable, but a transient 500, a timeout that fires after the record already committed server-side, or a policy error would silently create the credential in the classic vault instead, or leave two records for the same login. Either way the user sees success with no hint that the vault they picked was not used.

It would be safer to only fall back on the specific "NSF not available" condition, and to surface a warning when the fallback actually changes the target vault.

pwmplugin/Sources/iterm2-keeper-adapter/KeeperCommanderClient.swift
```diff
@@ -450,27 +735,55 @@ private func extractPasswordFromRawGet(data: Data) -> String? {
...
+
+    let classic = makeAttempt(verb: "record-update")
+    let nested = makeAttempt(verb: "nsf-record-update")
+    // Known vault: use only that family. Unknown: nested first — classic record-update
+    // returns status=success as a no-op for NSF UIDs on live Commander.
+    let attempts: [KeeperMutationAttempt]
+    switch source {
+    case .classic: attempts = [classic]
+    case .nested: attempts = [nested]
+    case nil: attempts = [nested, classic]
```
gnachman: The nested-first ordering here relies on an assumption that may not hold, and can silently drop a password change.

`runKeeperMutationAttempts` returns on the first response whose status is `success`. The comment notes that classic `record-update` returns a no-op success for NSF UIDs, which is why nested is tried first. But that reasoning is only safe if `nsf-record-update` does *not* also no-op-succeed on a classic UID. If it does (the symmetric case), then for an unlabeled classic record the nested attempt "succeeds" and the real `record-update` never runs, so the password is left unchanged while the UI reports success.

This is reachable in practice: legacy bare-UID identifiers from older sessions have no `sourceLabel`, and the add-account return path builds its `AccountIdentifier` with `sourceLabel == nil`, so editing a freshly added record before the list refresh also hits this heuristic. Worth confirming the nsf-record-update behavior against live Commander, or routing strictly by a known vault.

sources/PasswordManager/AdapterPasswordDataSource.swift
```diff
@@ -333,13 +333,18 @@ class AdapterPasswordDataSource: CommandLinePasswordDataSource {
 
     private func hydratePersistedCredentialsIfNeeded() {
         guard handshakeInfo?.persistsCredentials == true else { return }
+        // One-time cleanup of pre-migration Settings apiKey keychain entries.
```
gnachman: Dropping the settings-based API key with no migration may lock out existing users.

Before this change the adapter accepted the key from the `apiKey` settings field (`persistInKeychain: true`) as a fallback in `apiKey(fromHeader:)`. This PR removes that fallback and, here, also deletes the stored `apiKey` keychain entry on the very first hydrate, before anything reads it. A user who had entered their key in the API-key settings field rather than at the master-password prompt will lose it on upgrade and won't be able to authenticate until they re-enter it.

Consider migrating an existing `apiKey` keychain value into the master-password slot once before deleting it, so upgrades stay seamless.

sources/PasswordManager/AdapterPasswordDataSource.swift
```diff
@@ -583,11 +598,25 @@ class AdapterPasswordDataSource: CommandLinePasswordDataSource {
...
+                        sendOTP: entry.hasOTP,
+                        sourceLabel: entry.sourceLabel)
+                }
+
+                if let warning = response.warning?.trimmingCharacters(in: .whitespacesAndNewlines), !warning.isEmpty {
+                    DispatchQueue.main.async {
+                        iTermWarning.show(withTitle: warning,
+                                          actions: ["OK"],
+                                          accessory: nil,
+                                          identifier: "NoSyncKeeperListPartialFailure",
```
gnachman: A persistently silenceable warning here will permanently hide real failures.

This uses a fixed identifier with `silenceable: .kiTermWarningTypePersistent`, so it offers a "do not show again" that sticks. Once a user checks it, every future `nsf-list` failure is suppressed and their Nested accounts just quietly vanish from the list with no indication of why.

For a transient connectivity/partial-failure signal, a non-silenceable warning (or a less permanent surface like a status line) would fit better than one the user can permanently mute.

pwmplugin/Sources/iterm2-keeper-adapter/KeeperCommanderClient.swift
```diff
@@ -450,27 +735,55 @@ private func extractPasswordFromRawGet(data: Data) -> String? {
...
+            KeeperAdapterLog.write("setPassword: issuing verb=\(verb) uid=\(uid) \(redactedPasswordFragment(password: newPassword))")
+            return try client.executeCommand(apiKey: apiKey, command: cmd)
+        }
     }
-    let raw = keeperHumanReadableError(fromResponseData: data) ?? "Update failed"
-    throw KeeperClientError.message(keeperUserFacingPasswordUpdateError(apiDetail: raw))
+
+    let classic = makeAttempt(verb: "record-update")
+    let nested = makeAttempt(verb: "nsf-record-update")
+    // Known vault: use only that family. Unknown: nested first — classic record-update
```
gnachman: Small thing: this comment contains an em dash ("nested first — classic record-update"). The repo/global convention is to avoid em dashes in comments and other text that ships to third parties, so a comma or period would be preferable.

sources/PasswordManager/AdapterPasswordDataSource.swift
```diff
@@ -375,14 +380,24 @@ class AdapterPasswordDataSource: CommandLinePasswordDataSource {
...
             } else {
                 iTermUserDefaults.userDefaults().removeObject(forKey: "AdapterSetting_\(identifier)_\(field.key)")
             }
         }
+        deleteOrphanedLegacyApiKeySettings()
+    }
+
+    /// Pre-round-3 builds stored the API key under settings account "apiKey". That field is gone;
+    /// purge the orphan so Reset Configuration fully clears credentials.
+    private func deleteOrphanedLegacyApiKeySettings() {
```
gnachman, line 416: This one-time migration runs far more often than needed.

`hydratePersistedCredentialsIfNeeded` is called on every credential access, and it calls `deleteOrphanedLegacyApiKeySettings`, so a synchronous keychain delete happens on every hydrate just to clean up a legacy entry that is gone after the first run. On top of that, `eraseConfiguration` invokes it twice (once directly and once via `deleteAllSettingsFieldStorage`).

Since it is a strictly one-shot cleanup, it could be gated behind a NoSync "already migrated" flag and called once, and the duplicate call in the erase path removed.

pwmplugin/Sources/iterm2-keeper-adapter/KeeperCommanderClient.swift
```diff
@@ -482,37 +795,170 @@ private func escapeForKeeperDoubleQuotedCommandField(_ s: String) -> String {
...
+    KeeperAdapterLog.write("deleteRecord: source=\(source?.rawValue ?? "unknown") attempts=\(attempts.map(\.label))")
+
+    try runKeeperMutationAttempts(
+        logPrefix: "deleteRecord",
+        uid: uid,
+        attempts: attempts,
+        formatFailure: { keeperHumanReadableError(fromResponseData: $0) ?? "Delete failed" })
+}
+
+func addRecord(apiKey: String,
```
gnachman, line 870: There is some duplicated machinery worth consolidating.

`setPassword` and `deleteRecord` both route through the new `runKeeperMutationAttempts` helper, but `addRecord` hand-rolls its own verb-fallback loop with separate `firstError` bookkeeping, so the two implementations of "try each verb, prefer the first transport error, fall through on a body failure" can drift over time. In the same file, `taggedAsClassic` and `taggedAsNested` are identical except for the literal category string, and `KeeperRecordSource.fromLabel` just re-spells the compiler-synthesized `init(rawValue:)`. Collapsing each of these would remove parallel code paths that have to be kept in sync by hand.

gnachman (commented):
Automated review pass on the Nested Shared Folder work. Overall this is a solid, well-tested change; my main concerns are two spots where a record can end up in the wrong vault (the inline-edit re-add and the add-account fallback), plus a possible upgrade regression around the API key. Details are in the inline comments.

pkamble-ks:
@gnachman Thanks for the review — addressed in d1ae4cf77.

### Inline edit vault jump
Inline rename/username edit still does delete-then-re-add, but the re-add now passes `useClassicPermission` derived from the row’s `sourceLabel` (Classic → `YES`, Nested → `NO`), so a Classic record is no longer recreated via `nsf-record-add`.

### NSF add fallback
Default add still prefers `nsf-record-add`. Classic fallback now runs only when the error clearly indicates NSF is unavailable; other NSF failures fail closed instead of silently creating a Classic record.

### Unknown set-password routing
When vault is unknown (`sourceLabel` missing), set-password no longer guesses nested-first. It fails with a reload message so we don’t risk a no-op success on the wrong verb family.

### Legacy Settings `apiKey`
On hydrate we migrate an existing keychain `apiKey` into the master-password slot once (gated by a NoSync flag), then delete the orphan. Reset Configuration still clears it fully.

### `nsf-list` partial-failure warning
The host alert now uses `identifier: nil` so it can’t be permanently silenced. Nested-list failures stay visible.

### Cleanup
Em dash removed from the set-password comment; shared tag helper / `fromLabel` simplified.

### Verification
- `swift test --filter KeeperIntegrationTests`
- Manual UI: rename / add / set password / delete for both Classic and Nested

Happy to adjust anything else.

pwmplugin/Sources/iterm2-keeper-adapter/KeeperCommanderClient.swift
```diff
@@ -343,9 +444,111 @@ private func validatedRecordUID(_ recordUid: String) throws -> String {
...
+/// Prefers a network/timeout error over a later verb's body failure.
+private func runKeeperMutationAttempts(logPrefix: String,
+                                       uid: String,
+                                       attempts: [KeeperMutationAttempt],
+                                       formatFailure: (Data) -> String) throws {
+    func attempt(_ step: KeeperMutationAttempt) throws -> (success: Bool, raw: Data) {
+        KeeperAdapterLog.write("\(logPrefix): trying verb=\(step.label) uid=\(uid)")
+        let data = try step.run()
+        KeeperAdapterLog.write("\(logPrefix): \(step.label) response bytes=\(data.count)")
+        let success = (try? JSONDecoder().decode(KeeperExecuteResponse.self, from: data))?.status == "success"
```
gnachman: **Success detection assumes Commander's `message` is a String, but it can be an array.**

This decodes the full `KeeperExecuteResponse` only to read `.status`, and `KeeperExecuteResponse.message` is typed `String?` (line 18). If Commander returns `message` as an array, `decodeIfPresent(String.self, ...)` throws, `try?` collapses that to nil, and `success` becomes false. For a `.classic` password update (single attempt, no retry) that surfaces to the user as "Update failed" even though the record was actually changed.

The array shape is not hypothetical. This PR's own mock returns `{"status":"success","message":["no-op"]}` for `record-update` (commented "Mimic live Commander"), and the existing code already parses `parsed["message"] as? [String]` for `get` responses in `extractPasswordFromRawGet`. The mocks only exercise a string or absent `message` on the genuine-success paths, so this gap isn't covered by the current tests.

The same assumption bites `addRecordOnce`'s `RecordAddResponse.init(from:)` (around line 923), which uses a hard `try container.decodeIfPresent(String.self, forKey: .message)`. There an array `message` throws out of the whole decode; the nsf catch only handles `isNsfUnavailableError`, so it rethrows and the user sees a failure for a record that was actually created (duplicate on retry). And because `nsf-record-add` returns `data: null`, the UID is extracted from `message`, so that path structurally depends on `message` being a string.

Suggested fix: detect success by reading `status` alone (for example `JSONSerialization` plus `["status"] as? String`) rather than a full typed decode, and make the add-response `message` tolerant of both string and array. This is primarily a robustness fix; whether it is currently user-visible depends on Commander's exact success-message shape, which the tests don't pin down.

pwmplugin/Sources/iterm2-keeper-adapter/KeeperCommanderClient.swift
```diff
@@ -343,9 +444,111 @@ private func validatedRecordUID(_ recordUid: String) throws -> String {
     return uid
 }
 
-func listAccountsRecords(apiKey: String, client: KeeperCommanderClient) throws -> [PasswordManagerProtocol.Account] {
-    let data = try client.executeCommand(apiKey: apiKey, command: "ls -R -l")
-    var records: [KeeperRecord]?
+private let keeperEmbeddedRecordUIDRegex = try! NSRegularExpression(pattern: "[A-Za-z0-9_-]{15,}")
```
gnachman: **`extractRecordUID` scraping is loose.**

`keeperEmbeddedRecordUIDRegex` matches `[A-Za-z0-9_-]{15,}` and returns the first run that passes `validatedRecordUID`, which itself has no upper length bound (`^[A-Za-z0-9_-]{15,}$`). When the UID is scraped from a free-text success message, any 15-or-more character token appearing before the real UID (a folder or vault name, for example) would be returned instead. The current mock message happens to lead with the UID ("Record NSFUID123456789 was added successfully."), so tests pass, but the parse is fragile.

Keeper record UIDs are a fixed 22-character base64url encoding of 16 bytes, so anchoring to exactly that would be much safer, for example `(?<![A-Za-z0-9_-])[A-Za-z0-9_-]{22}(?![A-Za-z0-9_-])`. I'm not asserting this misfires today (there's no real Commander message text to confirm it), only that it's worth tightening.

sources/PasswordManager/AdapterPasswordDataSource.swift
```diff
@@ -583,11 +621,25 @@ class AdapterPasswordDataSource: CommandLinePasswordDataSource {
...
                         accountName: entry.accountName,
                         hasOTP: entry.hasOTP,
-                        sendOTP: entry.hasOTP)
+                        sendOTP: entry.hasOTP,
+                        sourceLabel: entry.sourceLabel)
+                }
+
+                if let warning = response.warning?.trimmingCharacters(in: .whitespacesAndNewlines), !warning.isEmpty {
+                    DispatchQueue.main.async {
+                        iTermWarning.show(withTitle: warning,
```
gnachman, line 629: **This NSF-listing warning can repeat, and can fire spuriously on non-enterprise accounts.**

It surfaces the plugin's `warning` with `.kiTermWarningTypePersistent`, which is the one warning type with no "don't show again" affordance. Two issues follow:

1. It re-fires on every list-cache invalidation, and this PR now invalidates after every add, delete, setPassword, and custom command (plus the 30-minute expiry). A few password edits in a session means the same unsuppressable modal several times, and consolidated reloads can stack more than one.
2. Nested Shared Folders is an enterprise-tier feature. For an account that simply doesn't have NSF, `nsf-list` plausibly fails on every call, so a benign and expected condition becomes a recurring modal.

Worth scoping so it (a) doesn't show too frequently and (b) doesn't fire spuriously when NSF just isn't available. The plugin already distinguishes "NSF genuinely unavailable" from a real error via `isNsfUnavailableError` on the add path; reusing that discrimination here (stay silent for the expected no-NSF case, warn only on a genuine transient failure) plus a silenceable warning type would cover both.

sources/PasswordManager/AdapterPasswordDataSource.swift
```diff
@@ -375,14 +379,48 @@ class AdapterPasswordDataSource: CommandLinePasswordDataSource {
...
+            let current = loadPersistedCredentials()
+            if current == nil || current!.isEmpty {
+                persistCredentialsToKeychain(legacy)
+                if masterPassword == nil || masterPassword!.isEmpty {
+                    masterPassword = legacy
+                }
+            }
+            _ = SSKeychain.deletePassword(forService: keychainCredentialServiceName, account: "apiKey")
+        }
+        defaults.set(true, forKey: flagKey)
```
gnachman: **The migration flag is set even when the legacy keychain read fails.**

`migrateLegacyApiKeyIfNeeded` reads the legacy key with `try?` (around line 406) and then unconditionally sets the "migrated" flag here. If that read returns nil because the keychain was momentarily locked or access was denied (not because the key is absent), the migration body is skipped but the flag is still set, so migration never retries. The old key still exists under the `apiKey` account, but the new build only reads the master-password slot, so the user is forced to re-enter it.

More broadly, this stores derived state (a bool in user defaults) to answer a question the keychain already answers: is the destination slot populated? A lazy, idempotent shape avoids the failure mode entirely. On hydrate, if the destination slot is empty, try the declared legacy source and copy it up. The keychain read is the check; any cached flag then becomes a pure optimization, set only after a successful confirming read, never correctness. That also removes the need to clear the flag on the settings-reset path.

Optional but cleaner: the Keeper-specific knowledge here (the literal `apiKey` account, the fact of the migration) could be declared by the plugin in the handshake, so the host runs one generic, adapter-agnostic migrator. That mirrors the existing "plugin declares, host executes" pattern and keeps adapter specifics out of the shared `AdapterPasswordDataSource`.

sources/PasswordManager/iTermPasswordManagerWindowController.m
```diff
@@ -1720,11 +1882,18 @@ - (void)tableView:(NSTableView *)aTableView
...
-                        [[weakSelf currentDataSource] addUserName:userName
-                                                      accountName:accountName
-                                                         password:password
-                                                          context:context
-                                                       completion:^(id<PasswordManagerAccount> _Nullable replacement, NSError * _Nullable error) {
+                        // Preserve vault on delete-then-re-add (Classic vs Nested).
+                        NSDictionary *readdFlags = nil;
+                        if ([(id)entry respondsToSelector:@selector(sourceLabel)]) {
+                            NSString *source = entry.sourceLabel;
+                            if ([source caseInsensitiveCompare:@"Classic"] == NSOrderedSame) {
```
gnachman, line 1892: **`caseInsensitiveCompare:` on a nil `source` takes the Classic branch.**

`entry.sourceLabel` can be nil even when the account responds to the selector. `[nil caseInsensitiveCompare:@"Classic"]` is a message to nil, which returns 0, and `NSOrderedSame` is also 0, so this condition is true when `source` is nil. The intended "vault unknown, fall through to a plain add" path (`readdFlags == nil`) is therefore dead for the nil case: a nil source silently forces `useClassicPermission:@YES`.

This is reachable. A freshly added account comes back as a `CommandLineProvidedAccount` with `sourceLabel == nil` (the add recipe doesn't thread it through) until the list re-fetches. Add a record into a Nested Shared Folder, then rename it before the refresh lands, and it gets re-added into the classic vault. Guarding on `source.length > 0` before the comparisons fixes the immediate bug.

Zooming out, this whole block is a symptom of implementing rename as delete-then-re-add. That path is lossy (folder membership, sharing, custom fields, and TOTP are discarded; only name, username, and password are reconstructed), it has a data-loss window (the record is deleted before the re-add, with the password held only in memory), and it forces Keeper vault vocabulary ("Classic", "Nested", "useClassicPermission") into the generic password-manager window. An optional in-place "update" verb in the adapter protocol, which Keeper can implement cheaply via `record-update` / `nsf-record-update` exactly like `setPassword` already does, with the current read-delete-write kept as the fallback for adapters that don't implement it, would retire all of the above at once: no delete, no vault to reconstruct, no metadata loss, a stable UID, and vault routing that lives in the plugin. Happy to sketch the protocol addition if useful.

gnachman (commented):
Reviewed the Keeper Nested Shared Folder changes. The routing-by-vault approach is sound and the test coverage is good. I've left five inline comments: two in the plugin (Commander response-shape assumptions) and three host-side (a warning that can spam, the credential migration, and a rename bug). None are hard blockers on their own, but the nil-compare rename bug and the warning behavior are worth fixing before this ships.

A theme across several of them: some of the friction here (vault preservation on rename, the credential migration wedged into the auth flow) comes from generic host code carrying Keeper-specific knowledge. Where relevant I've noted a deeper direction (an optional in-place "update" verb in the adapter protocol, and declaring migrations in the handshake) that would let these special cases go away rather than accrete. Those are forward-looking, not required for this PR. Thanks for the thorough work.

pkamble-ks:
@gnachman Thanks for the notes — pushed fixes for the five inline comments. Handshake-declared credential migration and an in-place adapter `update` verb are deferred, as you suggested.

### Plugin
- Success decode: `keeperCommandStatusIsSuccess` now reads `status` via `JSONSerialization` only, so an array `message` no longer makes a real success look like a failure. `RecordAddResponse.message` accepts a string or `[String]`.
- UID scrape: embedded scrape is exactly 22 chars with boundaries (`(?<![A-Za-z0-9_-])[A-Za-z0-9_-]{22}(?![A-Za-z0-9_-])`). Left the general `keeperRecordUIDRegex` `{15,}` as-is so existing identifiers still validate.

### Host
- NSF list warning: plugin stays silent when `isNsfUnavailableError` (expected on non-enterprise accounts). Genuine `nsf-list` failures still warn, with identifier `NoSyncKeeperListPartialFailure` and `kiTermWarningTypePermanentlySilenceable`.
- Legacy Settings `apiKey`: on hydrate, copy into the master-password slot only if that slot is empty. No migrated flag; a failed/nil keychain read does not lock out retry. Reset Configuration still deletes the orphan.
- Nil `sourceLabel` on rename: `source.length > 0` before `caseInsensitiveCompare:`, so nil no longer takes the Classic branch.

Also merged current `upstream/master` so the conflict check is clear.

### Verification
- `swift test --filter KeeperIntegrationTests`
- Live Commander: list / add / rename / set password / delete for Classic and Nested

Happy to adjust anything else.

gnachman:
I tested this against a live Keeper Commander Service Mode instance (Commander 18.1.1, the current release), and the Nested Shared Folder path fails because the `nsf-*` commands are not recognized by stock Commander.

Reproduction:

- In the Commander interactive shell, `nsf-list --records --format=json` prints Commander's full command list, which is its behavior for an unknown command. There is no `nsf-list`, `nsf-record-add`, `nsf-record-update`, or `nsf-rm` in that list. The record command family is `record-add`, `record-update`, `rm`, `get`, `list`, `list-sf`, and `share-folder`.
- Replaying the adapter's exact call against the service (`POST /api/v2/executecommand-async` with `{"command":"nsf-list --records --format=json"}`, then `GET /api/v2/result/<id>`) returns HTTP 400 with:

```json
{"command":"nsf-list --records","data":null,"error":"Invalid JSON response","status":"error"}
```

The "Invalid JSON response" is generated by the Commander service itself: it runs the unknown command, gets non-JSON output, and cannot parse it.

Two consequences for a user on stock Commander:

1. Every account-list load shows the "Nested Shared Folder listing failed" warning, because the error text ("Invalid JSON response") does not match `isNsfUnavailableError`, so it is treated as a real failure rather than an expected absence.
2. Adding an account is broken by default. New records default to Nested, so `addRecord` runs `nsf-record-add` first. That fails, and the fallback to classic `record-add` only triggers when `isNsfUnavailableError` matches, which it does not here, so the default add path throws and no record is created. Only checking "Use classic permission model" (classic `record-add`) works.

Could you clarify what provides the `nsf-*` commands in your environment? For example a Commander fork or plugin, a particular Commander version, or a specific enterprise/PAM tier. I could not find them in upstream Keeper Commander, so I would like to understand how the Nested Shared Folder path was validated before we decide how to handle stock-Commander users (graceful fallback so the warning stays silent and add still works, a capability probe, or gating the feature).

pkamble-ks:
@gnachman Thanks for testing against stock Commander 18.1.1 — your repro matches what we see on non-NSF tenants.

**Where `nsf-*` comes from**

Not a fork or plugin. Nested Shared Folder commands (`nsf-list`, `nsf-record-add`, `nsf-record-update`, `nsf-rm`, …) are part of Keeper Commander, but they are invitation-only / feature-flag gated (https://docs.keeper.io/enterprise-guide/sharing/nested-share-subfolders). Without the flag, Commander never registers those verbs, so interactive `nsf-list` dumps the help catalog and Service Mode wraps the non-JSON output as `{"error":"Invalid JSON response"}` — exactly what you hit. We validated the NSF path on an enterprise tenant where that flag is enabled (interactive `nsf-list` works there). Service Mode still needs those verbs in the API key `command_list` when NSF is on.

**Stock / trial behavior (your two consequences)**

Agreed: `"Invalid JSON response"` and the help-catalog dump were not treated as “NSF unavailable,” so list warned every refresh and default add never fell back to `record-add`.

**Fix (now on this PR)**

Widened `isNsfUnavailableError` to treat those stock shapes as expected absence (including `invalid json response`, unknown/unrecognized command, and help-catalog markers like `available commands` / `type help <command>`). Then:

1. List: no Nested-list warning when NSF isn’t there; classic list still works.
2. Default add: `nsf-record-add` fails → classic `record-add` fallback runs.

Verified on a trial Service Mode instance (no NSF): list stays quiet, default Add creates a classic record. NSF-enabled tenants are unchanged.

Happy to adjust further if anything still looks off on your 18.1.1 setup.

gnachman:
Thanks for the Keeper Nested Shared Folder work! I merged it and then built a set of follow-on changes on top of it: in-place record editing, a unified Edit panel, secure 1Password add/edit, and a round of hardening from code review plus field testing. Rather than push onto your branch, I put them on a branch in this repo so you can review them:

- **Branch:** `pwm-inplace-editing`
- **Compare (our commits on top of your merged work):** https://github.com/gnachman/iterm2/compare/1eb9eb165...pwm-inplace-editing

Since you own the Keeper adapter, I'd really appreciate your eyes on it — especially anything touching NSF, which I can't exercise without a Nested-enabled tenant.

**Keeper adapter (`KeeperCommanderClient.swift`, `main.swift`):**
- In-place edit via `record-update` / `nsf-record-update`, sending only the fields that changed. Title uses the `--title="…"` equals form (so a value starting with `-` binds correctly).
- `normalizedKeeperSourceLabel` now returns `nil` for an unrecognized `record_category` so the record falls through to list-based Classic/Nested tagging. Previously an unknown category (e.g. "shared") made the record unroutable and every edit failed with "vault is unknown".
- Mutation success now unwraps the `{status:"success", result:"…"}` envelope and evaluates the *inner* status.
- `extractPasswordFromRawGet` rejects error/status envelopes instead of returning them verbatim as the password.
- Diagnostics: the adapter streams command + per-poll status to stderr (the host logs it), and a request stuck in `queued` past ~10s fails fast with a "restart the Commander service" message (we hit a wedged Service Mode worker during testing).

**Protocol (`PasswordManagerProtocol.swift`) — all additive/optional, backward compatible:**
- `canEditInPlace`, `requiresPasswordForAdd`, and `ErrorResponse.needsAuthentication`.

**The main thing I can't verify myself:** please confirm on a real NSF tenant that Classic and Nested records still list, add, edit (rename + password change), and delete correctly and route to the right vault. The `isNsfUnavailableError` classic-fallback behavior is unchanged from your version.

Happy to adjust anything. Thanks again for the contribution!

pkamble-ks:
@gnachman Thanks for the follow-up changes. I tested the latest nightly build with a Nested-enabled Keeper tenant and verified the Keeper functionality you mentioned, including Classic and Nested record listing, add, edit, delete, and routing to the correct vault.

Everything is working correctly on my side, and I didn’t find any issues with the Keeper adapter or Nested Shared Folder functionality.

Since these are new enhancements to the existing Keeper Password Manager integration in iTerm2, I also wanted to ask if you have an approximate ETA for when these changes are expected to be included in a standard iTerm2 release from iterm2.com, and, if applicable, the Mac App Store build.

We’re trying to plan when we can point users to a stable build with the updated Keeper functionality instead of the nightly build, so any rough guidance on the release timeline would be appreciated.

Thanks!

gnachman:
I plan to release 3.7.0 in a few weeks. I’m currently waiting for app
review to complete on the companion iPhone app.

PR #687, opened 2026-06-17
Sort JSON keys in Responses API request for prompt-cache stability

sumleo:
The Responses API request body is encoded with a bare `JSONEncoder()` that doesn't set `.sortedKeys`. The tools array serializes Swift dictionaries (`JSONSchema.properties: [String: Property]` and `rawJSONStorage: [String: AnyCodable]`), and dictionary iteration order is randomized per process. So the same tool list can serialize with its keys in a different order from one run to the next.

OpenAI prompt caching keys on a byte-exact prefix match, and the tools live in that cached prefix. Reordered keys change the bytes without changing the meaning, so the cache_read path silently stops firing and every request pays full price.

The fix is to set `outputFormatting = [.sortedKeys]` on that encoder, mirroring what's already done on the Anthropic path in `CompletionsAnthropic.swift`, which has a comment explaining the same caching concern. The wire shape is unchanged; only key order becomes deterministic.

I couldn't build or run this locally since it needs Xcode, so I've kept the change minimal and am relying on CI to verify it compiles. Two edits only: the encoder change in `ResponsesAPIRequest.swift` and a one-line release-notes entry.

### DIFF
--- a/docs/notes-3.7.txt
+++ b/docs/notes-3.7.txt
@@ -1452,3 +1452,7 @@ Bug Fixes:
   status rollup was borrowing the
   session's identity and clearing it
   when the visible peer changed.
+- Sort JSON keys in the Responses API
+  request so OpenAI prompt caching gets
+  a stable prefix and reliably hits the
+  cache.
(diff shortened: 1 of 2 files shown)


sumleo:
Hi @gnachman, gentle nudge on this when you have a moment. It's a small, self-contained prompt-caching fix, and I'm happy to rebase or tweak anything if that would make review easier. Thanks for the project and your time!

gnachman:
Sorry for the delay. Merged, and thanks for the PR!

PR #688, opened 2026-06-21
Add screenshot background color, save destination, and Python API capture

kud:
## Summary

Improves the Make Screenshot feature and exposes screenshot capture to the Python API. Three independent changes:

1. **Configurable background colour.** The screenshot panel previously rendered transparent margins/uncovered areas. A new “Background:” colour well (a `CPKColorWell` with alpha) lets you fill behind the content; lower the opacity for a transparent background. The last-used colour is remembered (`NoSyncScreenshotBackgroundColor`) and defaults to the terminal’s background colour. The fill is composited at every render site for WYSIWYG: the standard save/copy path, the streaming/multi-part encoder (per batch), and the live preview.

2. **Configurable save destination.** Adds a `screenshotSaveLocation` advanced setting (empty ⇒ Desktop) used by the one-click **Save** button, plus a new **Save As…** button that opens `iTermModernSavePanel`, remembers the last folder, and works for multi-part output. Save logic is unified behind a small `SaveDestination` (directory + base name).

3. **Scriptable capture via the Python API.** New `ScreenshotRequest`/`ScreenshotResponse` proto messages (field 134), regenerated ObjC + Python gencode, server dispatch + `apiServerScreenshot:` delegate (reusing `PTYSession`’s existing render path), and a `Session.async_screenshot(background_color=None) -> bytes` client method. The API returns PNG bytes only — it does not write to a caller-supplied path, by design.

CleanShot 2026-06-22 at 01 06 12@2x

## Implementation notes

- Background compositing is a new `iTermAnnotatedScreenshot.applyBackground(to:color:)` helper (returns the image unchanged when transparent). The panel keeps using the no-margin `renderLines(toImage:)` and composites afterward, so layout/size are unchanged.
- The API handler runs on the main thread (request dispatch hops to the main queue), so AppKit/`backingScaleFactor` access during rendering is safe.
- Default change (surfaced intentionally): the panel’s default background is now the terminal’s colour rather than transparent; transparent remains reachable by lowering the well’s opacity.

## Testing

- `tools/build.sh Development` succeeds with no new warnings.
- The regenerated Python SDK imports and round-trips the new messages; `screenshot_request` is field 134 and the status enum resolves.
- Release notes added to `docs/notes-3.7.txt`.

Interactive runtime checks (panel colour/preview/Save As; end-to-end `async_screenshot`) are still recommended before merge.


api/library/python/iterm2/iterm2/session.py
```diff
@@ -453,6 +453,31 @@ async def async_send_text(
...
+        :param background_color: An optional web-style ``"#rrggbb"`` colour to
+            fill behind the content. When `None`, the session's own background
+            colour is used.
+
+        :returns: PNG-encoded image data.
+
+        :throws: :class:`~iterm2.rpc.RPCException` if something goes wrong.
+        """
+        # pylint: disable=no-member
+        result = await iterm2.rpc.async_screenshot(
```
gnachman, line 466: This new API method needs to go through the capabilities mechanism like the rest of the client. Right now there's no version gate, and the server still advertises `X-iTerm2-Protocol-Version: 1.16` (`iTermWebSocketConnection.m:464`), so a script can't detect the feature, and against an older iTerm2 the request falls into the malformed-request path and surfaces as a generic `RPCException("Invalid request. Upgrade iTerm2...")` rather than the typed `AppVersionTooOld` we raise everywhere else (compare `load_url`, `move_session`, `file_panels`).

Three changes to match convention:
1. Bump the header to `1.17` in `iTermWebSocketConnection.m:464`.
2. Add `supports_screenshot` / `check_supports_screenshot` (min_ver `(1, 17)`) to `capabilities.py`.
3. Call `iterm2.capabilities.check_supports_screenshot(self.connection)` at the top of this method.

sources/PTYSession/PTYSession.m
```diff
@@ -8288,6 +8288,34 @@ - (NSImage *)terminalContentSnapshot {
...
+        return nil;
+    }
+    NSColor *backgroundColor = nil;
+    if (colorString.length > 0) {
+        backgroundColor = [NSColor colorFromHexString:colorString];
+    }
+    if (!backgroundColor) {
+        backgroundColor = [_textview.colorMap colorForKey:kColorMapBackground];
+    }
+    NSImage *image = [_textview renderImageWithLines:NSMakeRange(range.location, range.length)
```
gnachman: This duplicates the render path in `terminalContentSnapshot` just above, but without its fallbacks: when `renderImageWithLines` returns nil (browser sessions, for instance) `terminalContentSnapshot` falls through to metal / `[_view snapshot]`, whereas here we return nil and the API reports INTERNAL_ERROR. Please reuse or generalize `terminalContentSnapshot` rather than standing up a parallel render path. (`terminalContentSnapshot` already fills with the terminal background, which is the behavior we want once the configurable color is gone.) Also note you call `[_textview refresh]` here, which `terminalContentSnapshot` deliberately skips for being side-effecting.

sources/PTYSession/PTYSession.m
```diff
@@ -8288,6 +8288,34 @@ - (NSImage *)terminalContentSnapshot {
...
+                                     includeMargins:YES
+                                    backgroundColor:backgroundColor
+                                         showCursor:YES];
+    if (!image) {
+        return nil;
+    }
+    NSData *tiff = image.TIFFRepresentation;
```
gnachman: This reimplements the exact TIFF -> NSBitmapImageRep -> PNG conversion that this same PR centralizes as `iTermAnnotatedScreenshot.pngData(from:)` (it's `@objc`, callable as `+pngDataFrom:`). Call the shared helper instead of re-rolling it here.

gnachman (changes requested):
Thanks for this, there's good work here. After discussing it on our end, we've decided the screenshot background should not be a user-selectable color. The problem genuinely worth fixing is that screenshots come out with transparent areas (not just the margins, but anywhere the renderer leaves uncovered). The right fix for that is to fill behind the content with the terminal's own background color, with no new UI or options. We'll handle that fill on our side so it lands consistently across all the render sites (panel, streaming encoder, preview) and avoids the premultiplied-alpha and preview-vs-output mismatches the configurable version runs into.

Could you revert the customizable-background-color parts of this PR and keep the rest?

**Please remove:**
- The "Background:" color well, its label/tooltip/layout, and the `backgroundColorChanged(_:)` action in `iTermScreenshotPanel.swift`.
- The `NoSyncScreenshotBackgroundColor` persistence: `initialBackgroundColor()`, `persistBackgroundColor()`, `selectedBackgroundColor`, `terminalBackgroundColor`.
- `iTermAnnotatedScreenshot.applyBackground(to:color:)` and its call sites in `renderFinalImage()`, `iTermStreamingScreenshotEncoder` (the `fillColor` parameter and per-batch fill), and the preview's `backgroundFillColor` / `effectiveBackgroundColor`.
- The `background_color` parameter on `Session.async_screenshot` / `rpc.async_screenshot`, the `background_color` field in `ScreenshotRequest`, and the `colorString` plumbing through `apiServerScreenshot:` and `-screenshotPNGDataWithBackgroundColorString:` (a plain `-screenshotPNGData` is fine). Applying "not selectable" consistently means API callers shouldn't pick a fill either.

**Please keep:**
- The Save / Save As destination work and the `screenshotSaveLocation` setting. These are nicely done.
- The `async_screenshot()` capture itself (returning PNG bytes), with the capabilities gate added per the inline comment.
- The `iTermAnnotatedScreenshot` refactor (`defaultScreenshotFilename`, `pngData(from:)`, `save(nsImage:inDirectory:)`, `write(nsImage:to:)`), which is good cleanup on its own.

That leaves the PR focused on two solid, independent features. Sorry to ask for a revert after the effort you put in here, but it keeps the surface area small and lets us fix the transparency at the rendering layer for every code path at once.

gnachman:
Quick update: I landed the margin/uncovered-area fill on `master` separately (it now composites the session's real background, including a background image with its mode and blend, behind the content). So when you revert the configurable-color work there's no need to worry about transparent margins coming back. Heads up that `master` now touches `iTermAnnotatedScreenshot`, `iTermStreamingScreenshotEncoder` (its initializer changed), and the preview, so you'll want to rebase.

api/library/python/iterm2/iterm2/session.py
kud: Done. Bumped the protocol header to `1.17` (`iTermWebSocketConnection.m`), added `supports_screenshot` / `check_supports_screenshot` (min_ver `(1, 17)`) to `capabilities.py`, and call `check_supports_screenshot` at the top of `async_screenshot`.

sources/PTYSession/PTYSession.m
kud: Done. The API path now goes through `terminalContentSnapshot` (so it gets the metal / `[_view snapshot]` fallbacks and skips the side-effecting `refresh`) instead of the parallel render path.

sources/PTYSession/PTYSession.m
kud: Done. Now calls the shared `+[iTermAnnotatedScreenshot pngDataFrom:]` helper instead of re-rolling the TIFF→PNG conversion.

kud:
Done — and rebased onto your transparent-area fill so this sits on top of it.

Removed all the customizable-background-color work: the Background colour well (label/tooltip/layout) + `backgroundColorChanged(_:)`, the `NoSyncScreenshotBackgroundColor` persistence, `applyBackground(to:color:)` and its call sites, the `fillColor` parameter on `iTermStreamingScreenshotEncoder` and the preview's `backgroundFillColor`, and the `background_color` field on `ScreenshotRequest` with its `colorString` plumbing (now a plain `screenshotPNGData`).

Kept the Save / Save As destination work + `screenshotSaveLocation`, the gated `async_screenshot()` capture, and the `iTermAnnotatedScreenshot` refactor. Since master now fills behind the content at the render layer, the panel/encoder/preview just use your `screenshotBackgroundSlice(...)` + `compositing(...)`.

gnachman:
Landed on master as f4ca0004a. I squashed the branch into a single commit (kept you as author) and made a few adjustments while reviewing: removed the leftover applyBackground(to:color:) helper from the reverted background-color work, fixed tilde expansion for the screenshotSaveLocation setting so a ~ path no longer silently falls back to the Desktop, and reverted Copy closing the panel (it was destructive). Thanks for the contribution!

kud:
@gnachman Thank you!

PR #689, opened 2026-06-25
Fix crash when tmux returns a nil state value

jdfalk:
## Problem

iTerm2 aborts (`SIGABRT`) when attaching with tmux integration (`tmux -CC`) to a remote SSH server. The whole app dies, making integration unusable against that server. Plain `tmux` (without `-CC`) is unaffected.

GitLab issue: https://gitlab.com/gnachman/iterm2/-/work_items/12908

Backtrace:
```
-[__NSDictionaryM setObject:forKey:]   *** object cannot be nil
+[TmuxStateParser dictionaryForState:workAroundTabBug:]
-[TmuxStateParser parsedStateFromString:forPaneId:workAroundTabBug:]
-[TmuxWindowOpener dumpStateResponse:pane:]
-[TmuxGateway invokeCurrentCallbackWithError:]
-[TmuxGateway executeToken:]
-[VT100ScreenMutableState(TerminalDelegate) terminalHandleTmuxInput:]_block_invoke
```

## Cause

`+[TmuxStateParser dictionaryForState:workAroundTabBug:]` inserts each parsed pane-state field into an `NSMutableDictionary` with `-setObject:forKey:` and no nil check. tmux is an external process — and with `tmux -CC` to a remote host it may be a different version than expected — so a field can come back in a form that yields a nil object, which makes `-setObject:forKey:` raise `NSInvalidArgumentException`. Nothing catches it, so the app aborts.

## Fix

Guard the insertion in `sources/tmux/TmuxStateParser.m`:
- Skip KVPs with an empty key
- If the value (or its converter result) is nil, log it and skip the field rather than crashing

A bad/unknown field now degrades to iTerm2's existing defaults instead of killing the app. A `NSLog` diagnostic identifies the exact tmux field causing the nil when it fires:
```
Ignoring nil value for key "<KEY>" in tmux state "<KVP>"
```

## Tests

Added `iTerm2XCTests/TmuxStateParserTest.m` with five tests — all pass:

| Test | What it covers |
|------|----------------|
| `testNormalParsing` | Known integer/string fields round-trip correctly (regression) |
| `testEmptyKeyIsSkipped` | A `=value` field (empty key) is silently dropped, not inserted |
| `testUnknownKeyIsPreservedAsString` | Unknown future tmux fields pass through as raw NSString |
| `testFieldWithNoEqualsIsIgnored` | KVPs with no `=` don't crash (pre-existing guard, regression) |
| `testCorrectPaneIsSelected` | Multi-pane state returns only the requested pane |

Run with:
```
xcodebuild test -project iTerm2.xcodeproj -scheme iTerm2Tests \
  -only-testing:iTerm2XCTests/TmuxStateParserTest \
  CODE_SIGN_IDENTITY="" CODE_SIGNING_REQUIRED=NO CODE_SIGNING_ALLOWED=NO
```

### DIFF
--- a/docs/notes-3.7.txt
+++ b/docs/notes-3.7.txt
@@ -497,6 +497,11 @@ Bug Fixes:
   prefer rich text.
 - Clicking in a porthole in an inactive split
   pane now activates that pane.
+- Fixed a crash (SIGABRT) when attaching with tmux
+  integration (tmux -CC) to a remote server whose
+  tmux version returns an unexpected state field.
+  iTerm2 now logs the bad field and continues
+  instead of aborting.
 - Fixed several crashes: registering a profile
   hotkey for a profile with no GUID, loading a
   profile whose default-bookmark value is a
--- a/iTerm2.xcodeproj/project.pbxproj
+++ b/iTerm2.xcodeproj/project.pbxproj
@@ -700,6 +700,7 @@
 		38C1ECCA9DCF67E64C50E1BC /* iTermRenderingComparer.swift in Sources */ = {isa = PBXBuildFile; fileRef = 80A64098B399CE71B981E2C7 /* iTermRenderingComparer.swift */; };
 		398447BD6E0F4CA316672068 /* CompanionPairingWindowController.swift in Sources */ = {isa = PBXBuildFile; fileRef = E8D4F38B871E9DA9C0AA5E10 /* CompanionPairingWindowController.swift */; };
 		3B049F12D648A9AFEEFCA7A1 /* OpenDirectory.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = A60BD9181B3F5D76007D7F11 /* OpenDirectory.framework */; };
+		3B7D01668006EAF08D152B0C /* TmuxStateParserTest.m in Sources */ = {isa = PBXBuildFile; fileRef = AB48E99BCC4A9B8B45479CA0 /* TmuxStateParserTest.m */; };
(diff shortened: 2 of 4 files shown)


gnachman:
Thanks for digging into this and for the nice test coverage.

After investigating, the SIGABRT here is a symptom of a deeper macOS 26+ regression. `TmuxStateParser.m` declares a private `NSString (TmuxStateParser)` category with a `-numberValue` method, and `dictionaryForState:` dispatches to it dynamically via `performSelector:`. macOS 26 added its own `-[NSString numberValue]` to the framework, which shadows our category and returns nil. So every numeric pane field (`cursor_x`, `cursor_y`, the mode flags, scroll regions, etc.) converted to nil, which is what then hit `-setObject:forKey:` and aborted.

That means the nil-guard alone stops the crash but leaves tmux integration silently broken on macOS 26+: the numeric fields get dropped and the pane falls back to defaults (wrong cursor position, modes, scroll region). The real fix is to keep the framework from shadowing us.

Fixed on master:
- `452f8482` renames the category methods to `it_numberValue` / `it_intlistValue` / `it_paneIdNumberValue` so they can never collide with current or future framework methods. This is the actual fix for the regression.
- `0ff695af` folds in your defensive ideas as defense-in-depth: skip empty-key KVPs, and never insert a nil converter result. It also adds a ModernTests test (this repo's test target) covering numeric/paneId/intlist conversion plus the malformed-input cases. The numeric-conversion test would have regressed to nil under the collision, so it guards the root cause directly.

Closing in favor of those commits, but genuinely appreciate the report and the clear repro. Crediting your contribution in the change.

PR #690, opened 2026-06-25
fix(tmux integration): reordered the format string for `resize-pane`.

daneofmanythings:
`resize-pane` has undergone some slight api changes. Moved the format specifier for the size arguement so that it is recognized by the direction flag.


daneofmanythings:
Fixing upstream. Sorry for the noise!

PR #691, opened 2026-06-30
Add "Force new terminal keyboard" preference and related functionality

gfreezy:
This update introduces a new setting in the Keys Preferences to enable a specific keyboard layout for newly created terminal sessions. The implementation includes:

- New preference keys: `kPreferenceKeyForceNewTerminalKeyboard` and `kPreferenceKeyNewTerminalKeyboardLocale`.
- UI elements in the Keys Preferences view for user interaction.
- Logic in `InputSourceForcer` to apply the selected keyboard layout when a new terminal session is created.
- Updates to session management to ensure the new setting is respected.

This feature allows users to have a consistent keyboard layout experience across new terminal sessions, enhancing usability.

### DIFF
--- a/docs/notes-3.7.txt
+++ b/docs/notes-3.7.txt
@@ -1520,6 +1520,12 @@ Bug Fixes:
   silenceable, so you can make it
   always save to Downloads or always
   show a save panel. Issue 12897.
+- Settings > Keys > System adds “Force new
+  terminal keyboard”, which selects a
+  keyboard layout when a newly created
+  terminal first receives keyboard
+  focus. Restored sessions are not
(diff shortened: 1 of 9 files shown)


gnachman:
Squash-merged to master as 0a5f589e6, with a redesign: instead of a separate "Force new terminal keyboard" preference, this is now a refinement of the existing "Force keyboard" setting via an "Only force once per session when it first receives keyboard focus" sub-option. This avoids the two features conflicting (global forcing is a sustained force that would otherwise override a one-shot). Thanks @gfreezy, your authorship is preserved on the commit.

PR #693, opened 2026-06-30
Inherit tab color in split-pane title bar for Minimal theme

wagnerand:
Fixes https://gitlab.com/gnachman/iterm2/-/work_items/12903.

This is my first contribution to iTerm2, I obviously don't have a deep understanding of the architecture and whether I got this right. Any feedback is appreciated!

I tested this locally and it appears to work fine. Kudos to making the local development environment very easy to set up!

### DIFF
--- a/sources/TerminalView/iTermTheme.m
+++ b/sources/TerminalView/iTermTheme.m
@@ -129,7 +129,7 @@
         }
     }
     iTermPreferencesTabStyle preferredStyle = [iTermPreferences intForKey:kPreferenceKeyTabStyle];
-    if (!tabColor || self.useMinimalStyle) {
+    if (!tabColor) {
         return [self dimmedBackgroundColorWithAppearance:effectiveAppearance
                                   sessionBackgroundColor:sessionBackgroundColor
                                         isFirstResponder:isFirstResponder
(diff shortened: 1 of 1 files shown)


gnachman:
What's the motivation for the change? Is it because you have multiple sessions within a tab that have different colors? I'd prefer to narrow the scope if possible because at least to my eyes it looks better when the per-pane titlebar matches the background color.

wagnerand:
When I drafted the issue, it was indeed a feature request initially, but while typing I realized that the tab color is inherited for all other themes, which led me to believe this would rather a bug.

The use case is indeed several different sessions within a tab with at least some of them having different color presets and tab colors. To be concrete, it's a 4-pane tab that looks like this. Each number represents a unique profile.
(I would share my main settings file, but it contains a bunch of sensitive information; the issue itself can be reproduced with the profile attached on gitlab though.)
```
|=========================================================|
|                                             |           |
|                                             |           |
|                                             |     3     |
|                      1                      |           |
|                                             |           |
|                                             |-----------|
|                                             |           |
|---------------------------------------------|           |
|                                             |     3     |
|                      2                      |           |
|                                             |           |
|=========================================================|
```

I've been using this workflow for a couple of months using a saved window arrangement and I kept noticing that I sometimes realized I am in the wrong pane after I started typing. The more visual distinction between the profiles would certainly help.

gnachman:
Thanks for the detailed justification, and sorry for not circling back here sooner. You actually convinced me: the case where a tab holds panes with different tab colors is one the shared tab bar genuinely can't represent, so tinting each pane's title bar with its own color is the right call there.

I ended up implementing this in `4bcd8bf1c` (issue 12903), scoped to exactly that situation: when a tab has two or more distinct tab colors, each pane's Minimal title bar is tinted with its own tab color; when the panes share a color (or there's a single one), the title bar keeps blending with the session background, since the tab bar already conveys the color there. That covers the 4-pane, different-profiles workflow you described while keeping the default look unchanged.

Since that shipped, I'm going to close this PR as addressed. Thank you for pushing on it and for the clear writeup of the use case. If the landed behavior doesn't match what you need in some configuration, please reopen or file a follow-up.

wagnerand:
Thank you! I will check it out in the next stable update.

PR #694, opened 2026-07-01
Improve tab dragging performance and drop-target stability

gfreezy:
## Summary
- Keep Metal enabled during normal tab drags. Only tmux split-pane drags fall back to legacy drawing. This avoids forcing every session onto Core Text for the whole drag, which starved the main thread and made the floating tab preview lag behind the cursor.
- Use the live pointer for split-pane drop highlights during tab drags. AppKit’s `draggingLocation` in `draggingUpdated:` can be stale while `PSMTabDragAssistant` is updating the floating drag window.
- Add hysteresis when switching split-half highlights and tab reorder slots (around tab midpoints instead of placeholder edges) to reduce flicker.

## Test plan
- [x] Drag a tab within the tab bar and confirm reorder slots feel stable and switch near tab centers
- [x] Drag a tab onto a pane and confirm the split highlight tracks the cursor without flicker
- [x] Confirm Metal stays active during plain tab drags (no global fallback to Core Text)
- [x] Drag split panes in a tmux tab and confirm layout still updates correctly

### DIFF
--- a/ThirdParty/PSMTabBarControl/source/PSMTabDragAssistant.m
+++ b/ThirdParty/PSMTabBarControl/source/PSMTabDragAssistant.m
@@ -1209,14 +1209,15 @@ static CVReturn DisplayLinkCallback(CVDisplayLinkRef displayLink,
     // mouse at beginning of tabs
     NSPoint mouseLoc = [self currentMouseLoc];
     PSMTabBarCell *proposedTarget = nil;
+    NSRect overCellRect = NSZeroRect;
+    PSMTabBarCell *overCell = nil;
 
     if ([self destinationTabBar] == control) {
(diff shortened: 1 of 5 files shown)


gnachman:
Thank you, Alex. This is a nice improvement to the drag feel, and keeping Metal enabled during plain tab drags is the right call. I merged it (squashed) as 491b36899, with you as the author.

One heads-up: the commit had re-saved docs/notes-3.7.txt in a non-UTF-8 encoding, which turned every curly quote, en dash, and ellipsis in the whole file into question marks. I restored the file and kept just your tab-dragging entry, so nothing to do on your end, but you may want to check your editor's encoding setting to avoid it next time.

Thanks again for the contribution!

PR #695, opened 2026-07-01
Fix cmd-click crash when command-mark data is unavailable

onemedicine:
## Summary
- guard `VT100ScreenState` command-mark lookup against out-of-bounds coordinates, missing marks, and missing interval-tree entries
- let command-click URL resolution fall through cleanly when command-mark data is unavailable instead of crashing
- add targeted `VT100ScreenTests` coverage for the out-of-bounds and no-mark cases, and document the fix in the 3.7 release notes

## Test plan
- [x] `make paranoid-deps`
- [x] `tools/run_tests.expect ModernTests/VT100ScreenTests/testCommandMarkAtReturnsNilWhenLineIsOutOfBounds ModernTests/VT100ScreenTests/testCommandMarkAtReturnsNilWhenNoMarkOnLine`

## Notes
- I rebased this to the current `master` so the PR is a single fix commit on top of upstream.

### DIFF
--- a/ModernTests/VT100ScreenTests.swift
+++ b/ModernTests/VT100ScreenTests.swift
@@ -679,6 +679,41 @@ class VT100ScreenTests: XCTestCase {
         })
     }
 
+    func testCommandMarkAtReturnsNilWhenLineIsOutOfBounds() {
+        let screen = self.screen(width: 10, height: 4)
+
+        screen.performBlock(joinedThreads: { _, mutableState, _ in
+            var range = VT100GridWindowedRange()
+            let negative = mutableState.commandMark(at: VT100GridCoordMake(0, -1),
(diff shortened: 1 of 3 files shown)


hiSandog:
For the orphaned modal sheet deadlock, the most valuable test would destroy or close the parent window while the command-click modal is still pending. That should assert the sheet is dismissed or safely detached instead of leaving the app in a blocked modal state with no owning window.

gnachman:
Thank you! This is a clean, well-guarded fix, and I appreciate the thorough regression tests covering the out-of-bounds, no-mark, and mark-without-command paths (and for addressing the review note about mustHaveCommand). All three tests pass here. Merged (squashed) as f036dee2d, with you as the author.

PR #697, opened 2026-07-04
Fix VoiceOver QuickNav, stale content, and review-cursor jumps in the terminal

Deon588:
Fixes three VoiceOver problems when interacting with terminal content, none of which occur in Terminal.app. Reported in https://gitlab.com/gnachman/iterm2/-/issues/12920.

1. QuickNav navigation does not engage after VO-interacting with the terminal until QuickNav is toggled off and on.
2. Content goes stale during an active interaction: after Cmd+K and a new command, VoiceOver keeps reading the pre-clear text and cannot reach the new output without re-interacting.
3. The VoiceOver review cursor jumps to the top of the terminal whenever a command produces output.

Causes and fixes:

- The terminal advertises an AXDocument attribute and a settable AXValue (settable only because of the no-op setter kept for Full Keyboard Access, Issue 10023). That makes VoiceOver treat the view as more than a plain text area, which wedges QuickNav on interaction. Fixed by hiding both from the accessibility surface while VoiceOver is running via isAccessibilitySelectorAllowed:. Full Keyboard Access is unaffected while VoiceOver is off; when both run at once, hiding AXValue wins because QuickNav breaks otherwise.
- The announce-new-output path suppresses value-changed notifications during an output burst and posted layout-changed with NSAccessibilityUIElementsKey once output settled. That notification directs VoiceOver to move its cursor to the listed element — the jump to the top — and since no content notification was re-posted at settle, VoiceOver's cached text went stale during interaction. Fixed by replacing the layout-changed with a value-changed + selected-text-changed pair posted after the announcement: the pair makes VoiceOver refresh its cached content and extend its navigable model (value-changed alone does not) without moving the review cursor. Terminal.app posts the same pair for output and never posts layout-changed for it.

Testing: verified with VoiceOver on macOS 26.5.2 — QuickNav engages immediately on interaction; output announcements still speak; the review cursor holds its position when new output arrives and the new output is reachable by navigating down; shell command history at the prompt and deleted-character announcements are unaffected. Release notes updated.

### DIFF
--- a/docs/notes-3.7.txt
+++ b/docs/notes-3.7.txt
@@ -2243,3 +2243,14 @@ Bug Fixes:
   these keys to tmux by name so tmux encodes
   them just as it does outside control mode,
   honoring its extended-keys-format.
+- Fix VoiceOver QuickNav navigation not
+  engaging after interacting with the
+  terminal, which previously required
+  toggling QuickNav off and on.
+- Fix VoiceOver reading stale terminal
+  text when the screen was cleared and
+  new output arrived while interacting.
+- Fix VoiceOver focus jumping to the top
+  of the terminal when new output
+  arrived, and make new output reachable
+  without re-interacting.
--- a/sources/TerminalView/PTYTextView.m
+++ b/sources/TerminalView/PTYTextView.m
@@ -3215,17 +3215,6 @@ static NSString *iTermStringForEventPhase(NSEventPhase eventPhase) {
     _accessibilityPendingOutputCursorLineString = nil;
 }
 
-- (void)accessibilityPostOutputLayoutChangedNotification {
-    NSDictionary *info = @{
-        NSAccessibilityUIElementsKey: @[ self ],
-        NSAccessibilityPriorityKey: @(NSAccessibilityPriorityLow)
-    };
-    NSAccessibilityPostNotificationWithUserInfo(
-        self,
-        NSAccessibilityLayoutChangedNotification,
-        info);
-}
-
 + (NSArray<NSString *> *)accessibilityAnnouncementLinesForTrimmedLines:(NSArray<NSString *> *)trimmedLines
                                                      firstAbsoluteLine:(long long)firstAbsoluteLine
                                                        oldAbsoluteCursorY:(long long)oldAbsoluteCursorY
@@ -3278,12 +3267,8 @@ static NSString *iTermStringForEventPhase(NSEventPhase eventPhase) {
 
     AccLog(@"Output settle: firstAbsY=%lld lastAbsY=%lld cursorAbsY=%lld overflow=%lld", firstAbsY, lastAbsY, absCursorY, overflow);
 
-    if (firstAbsY > lastAbsY) {
-        AccLog(@"No lines to announce (firstAbsY > lastAbsY)");
-        [self accessibilityClearPendingOutputState];
-        return;
-    }
-
+    // When firstAbsY > lastAbsY (e.g. a bare clear) the loop collects nothing and no
+    // announcement is made, but the clear+refresh tail below must still run.
     NSMutableArray<NSString *> *trimmedLines = [NSMutableArray array];
     for (long long absLine = firstAbsY; absLine <= lastAbsY; absLine++) {
         int relativeIndex = (int)(absLine - overflow);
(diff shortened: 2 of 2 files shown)


gnachman:
Thanks for this, and for the unusually precise writeup. I investigated the same three symptoms independently on master and want to share one finding before it lands, since it may affect the QuickNav (step 1) fix.

I reproduced your attribute-surface approach (drop AXDocument, make AXValue non-settable to VoiceOver via `isAccessibilitySelectorAllowed:`) and confirmed it fixes the initial QuickNav wedge and preserves Full Keyboard Access spaces (Issue 10023) when VoiceOver is off.

But I hit a residual leak the initial-interaction test does not cover: if you run a second command while still interacting and then press Up/Down, the arrows leak to the shell again (history navigation), even with AXValue hidden. It is essentially step 1 recurring after output.

The debug log shows the trigger is independent of AXValue settability. Right before the leaked arrow:

```
accessibilityRole -> AXTextArea
accessibilityNumberOfCharacters -> 162
accessibilitySelectedTextRange -> [Empty range at 162]   (caret at end)
Received KeyDown ... keyCode=126   (up arrow, no VO modifier, leaks)
```

VoiceOver reads us as an editable AXTextArea whose caret just moved, switches into text-editing mode, and defers the arrow as caret movement. The thing that flips it into that mode is the value-changed + selected-text-changed pair that `refreshAccessibility` already posts whenever the terminal cursor moves during output. Toggling QuickNav off/on clears it because that resets VoiceOver's browse-vs-edit mode, which matches your original step 1 description.

Why I am flagging it here: your steps 4/5 fix posts a value-changed + selected-text-changed pair at settle, which is the same class of notification I saw pulling VoiceOver into caret mode. So it is worth verifying this specific repro on your branch:

1. VoiceOver + QuickNav on
2. `seq 1 40`, VO-interact
3. navigate up into the output
4. run another command (e.g. `ls`)
5. press Up/Down: do they still move the VO cursor, or leak to the shell?

I tested an approximation on master, not your exact branch, so your full combination (especially the ordering of the announcement vs the notification pair) may behave differently, but we are on the same OS (macOS 26.5.2) so I suspect it reproduces. If it does, the deeper issue is that a focused editable AXTextArea with a caret inherently makes VoiceOver treat arrows as caret keys, and the fix may need to suppress the cursor-follow notifications while VoiceOver is browsing (or move away from one editable AXTextArea toward per-line elements, which is possibly how Terminal.app avoids it).

Happy to pair or dig in the debugger if useful.

Deon588:
Thanks for the careful analysis. I ran your exact reproduction steps on this branch (rebased onto 4416dd38c today), macOS 26.5.2, VoiceOver with QuickNav on: `seq 1 40`, interact with the terminal, navigate up, then `ls`, then `seq 1 20`, staying interacted throughout. Bare Up/Down kept moving the VoiceOver cursor after the second and third bursts. There was no leak to shell history and no QuickNav toggle was needed. One VoiceOver cursor effect does occur at settle: focus follows to the newest output (caret at end), which as a VoiceOver user I actually want after running a command. It does not flip VoiceOver into caret mode here.

An AXObserver capture over the whole reproduction steps/flow shows why I think your master test behaved differently. On this branch the only notifications at settle are a single pair of AXValueChanged and AXSelectedTextChanged, and zero AXLayoutChanged. Master still posts AXLayoutChanged with NSAccessibilityUIElementsKey at settle (`accessibilityPostOutputLayoutChangedNotification`). Per the AppKit docs that notification directs VoiceOver to move its cursor to the listed element, and while diagnosing the jump to top symptom I could reproduce VoiceOver cursor and state disruption from exactly that post. Since your approximation kept master's settle path, I suspect the leak you saw was triggered by AXLayoutChanged rather than by the pair. Would you mind rerunning your sequence on the branch itself?

Two data points from the diagnosis that may save you time. First, Terminal.app does not expose the terminal as separate elements per line: its accessibility surface is a single AXTextArea with zero children and the same parameterized attributes as iTerm's (no text markers). Second, what Terminal does differently, per an AXObserver capture, is preserve the caret VoiceOver sets in AXSelectedTextRange across output (it held the review position while content grew), whereas iTerm snaps the reported caret to the shell cursor on every move. If an edit mode flip does show up in some configuration, holding the caret VoiceOver set rather than suppressing notifications looks like the principled direction.

If it still reproduces for you on the branch, I would take you up on pairing. One thing to note: I think we might be in different time zones, I am based in South Africa, so UTC+2. Could differences in VoiceOver settings (for example “Synchronize keyboard focus and VoiceOver cursor”) explain the divergence?

On a side note: it would be amazing to get this landed. With these fixes iTerm2 becomes, for the first time a genuinely usable terminal for blind users, a real alternative to macOS Terminal with all of iTerm2's power on top. That would mean a lot to me and to other VoiceOver users...

gnachman:
Landed in master as 82f7d8866, thank you.

I reran your exact sequence on the branch itself (not my earlier master-based approximation): seq 1 40, interact, navigate up, ls, then seq 1 20, staying interacted throughout, VoiceOver on macOS 26.5.2. All three fixes check out: QuickNav engages immediately on interaction, fresh output after a clear is reachable and read, and the review cursor follows to the newest output at the end rather than jumping to the top. Crucially the shell-history leak I saw before did not reproduce, so your diagnosis was right: what I hit on master was the retained AXLayoutChanged post with NSAccessibilityUIElementsKey, not the value-changed plus selected-text-changed pair. My approximation had kept master's settle path, which is exactly why it diverged.

I squash-merged this manually rather than with the button because the branch had drifted behind master and hit a trivial conflict in docs/notes-3.7.txt (both sides appended release-note entries). I resolved it as the union of both and kept your PTYTextView.m change byte for byte.

One observation for a possible follow-up: in my VoiceOver config the terminal still gets web-style single-key navigation (typing l says 'link not found', 1 says 'heading not found'), which suggests it is still being classified as document/web content in some setups. That is pre-existing and unaffected by this change, and your suggestion to hold the caret VoiceOver sets in AXSelectedTextRange rather than snapping to the shell cursor looks like the principled direction for it. Thanks again for the careful diagnostic work here.

Deon588:
Thank you for testing the branch, confirming the diagnosis, and landing this! Really happy to see it in master.

About the observation that typing l says “link not found” and 1 says “heading not found”: that is almost certainly not iTerm being classified as web or document content. It is a VoiceOver feature called single key Quick Nav (Apple calls it “single key webpage navigation”). When that setting is enabled and Quick Nav is on, VoiceOver captures single letters as navigation keys (l for link, h for heading, 1 for heading level one, and so on) anywhere Quick Nav is active, not just in real web areas. The announcements you heard are VoiceOver reporting that no such element exists nearby, which is expected inside a terminal. It also means letters do not reach the shell while Quick Nav is active, which is why VoiceOver users who live in the terminal generally keep it off.

To turn it off: open VoiceOver Utility (VO F8, that is Control Option F8, or Caps Lock F8 if you use the Caps Lock modifier), go to Commanders, then the Quick Nav tab, and uncheck “Enable single key webpage navigation”. Arrow based Quick Nav keeps working exactly as before; only the single letter capture stops. Turning Quick Nav off entirely (Left and Right arrows together) also suspends it.

In fairness, the VoiceOver team made this confusing when they shipped it: the setting is named after webpages but fires anywhere Quick Nav is active, and it has been enabled by default, so it often reads as the app being treated as a web page. So no iTerm change is needed for that symptom, it is VoiceOver configuration.

Thanks again for the review and the merge. This means a lot.

PR #698, opened 2026-07-05
Confirm screenshot Copy to Clipboard visually

kud:
## Problem

The screenshot window's **Copy to Clipboard** button gives no UI feedback. It writes the image to the pasteboard and returns — no panel close, no label change, nothing. Because the sibling **Save** button *looks* like it worked only by closing the panel, Copy reads as a dead button even though the clipboard is correctly populated.

## Fix

Add a transient confirmation: on copy, the button briefly becomes a green checkmark + "Copied!" for ~1.2 s, then reverts to its normal title. Wired into **both** copy paths (`copyClicked(_:)` and `copyToClipboardFromFile(_:)`), so it fires regardless of which one handles the image.

Implementation notes:

- A `copyFeedbackGeneration` counter invalidates a prior revert timer, so a rapid second copy can't leave the label stuck or revert early.
- The checkmark uses `SFSymbol.checkmarkCircleFill.rawValue` rather than a raw SF Symbol string literal, per the project's convention.
- Not called from the multi-part path — Copy is disabled when `cachedNumberOfParts > 1`.

## Notes

- Release note added to `docs/notes-3.7.txt`.
- Verified with a Development build (`tools/build.sh`) — compiles clean.


kud:
Heads-up: the failing `xcode-tests` here isn't from this change. `master`'s own Tests workflow has been red since ~11 Jul — `ModernTests/WorkgroupAutoSendClippingsTests.swift` no longer compiles because a call to `iTermWorkgroupPeerPort.clippingsToAutoSend(...)` is missing the newly-added `mainSessionState:` argument. This PR only touches `sources/Screenshots/iTermScreenshotPanel.swift` (+ release notes), so the red is inherited from the base branch, not caused here. Happy to rebase once mainline is green.

gnachman:
Thank you! Nice, self-contained fix for a button that really did read as dead. I like the generation-counter guard against a rapid second copy, and following the SFSymbol convention. Merged (squashed onto current master) as 4416dd38c, with you as the author. And you were right that the CI red was inherited from a broken test on master, not this change.

PR #701, opened 2026-07-08
Add support for videos as background images

OneTrickSupport:
## Summary

Adds support for using short videos (mp4/mov) as terminal background images, addressing the long-standing request for animated backgrounds (GitLab issue #7920).

- The Settings &gt; Profiles &gt; Window file picker accepts videos alongside images, validates that the file is playable and has a video track, previews the first frame, and shows a permanently-silenceable notice about the energy cost of high-resolution / high-frame-rate files.
- Videos play muted in a seamless loop (AVPlayerLooper) and respect blend, transparency, and the scaling modes. Tile is not offered for videos, since neither renderer can tile one.
- Works with **both renderers**:
  - **GPU renderer:** `iTermImageWrapper` owns the frame pull — it dequeues from the shared `AVPlayerItemVideoOutput` at most once per display refresh and wraps that frame as a BGRA Metal texture (zero-copy, via `CVMetalTextureCache`) once per GPU. Every pane showing the video samples that same texture for a given refresh. The update cadence treats a playing video like an inline animated GIF so frames draw at display rate.
  - **Legacy renderer:** an `AVPlayerLayer` in `iTermImageView` plays the video; playback position carries over seamlessly when the renderer switches because both paths share one player.
- `iTermImageWrapper` owns a single shared muted `AVQueuePlayer` + `AVPlayerItemVideoOutput` per video and loads a poster frame for image-only consumers (blending, snapshots). Consumers hold playback interests, so decoding stops when no window can show the video and split panes showing the same video share a single decoder.

## Test plan

Manually verified on macOS 26 (Apple Silicon), Xcode 26.6:
- Picking mp4/mov via the picker; picker still accepts images and folders; corrupt-file warning still works
- Playback under the GPU renderer (confirmed active) and under the legacy renderer (transparency enabled)
- Split panes: shared decoder, synchronized playback with no seam at the dividers; per-pane and whole-window background modes
- Blend slider and scaling modes live-update
- Six panes under sustained heavy output (~100% CPU for 75s) with no crash and memory returning to baseline
- Removing the video restores normal behavior

Not yet re-verified in the UI after the review fixes: the occlusion/miniaturize pause, the idle-session first frame, and the three settings changes (source-mode refresh, energy-notice suppression, Tile disabled).

## Known limitations

- Drag-and-drop of a video onto the settings image well is not supported (picker only) since NSImageView filters non-image drags
- The Metal path samples raw BGRA without color-space conversion, so colors can differ very slightly from the legacy path

### DIFF
--- a/docs/notes-3.7.txt
+++ b/docs/notes-3.7.txt
@@ -91,6 +91,13 @@ Major New Features:
   code review or perform in-terminal diffs of a
   git repo.
 
+- Background images can now be videos (mp4 or
+  mov). Videos play muted in a seamless loop
+  and respect the blend and scaling settings.
+  They work with both the GPU and legacy
+  renderers, and panes showing the same video
+  share one decoder.
+
 - Clippings is a new side panel that contains a
   list of items formatted as markdown. They're a
   useful place to store text that you need
--- a/docs/notes-3.7.txt
+++ b/docs/notes-3.7.txt
@@ -93,10 +93,14 @@ Major New Features:
 
 - Background images can now be videos (mp4 or
   mov). Videos play muted in a seamless loop
-  and respect the blend and scaling settings.
-  They work with both the GPU and legacy
-  renderers, and panes showing the same video
-  share one decoder.
+  and respect the blend and scaling settings,
+  except tiling, which is not offered for
+  videos. They work with both the GPU and
+  legacy renderers, and panes showing the same
+  video share one decoder. Playback pauses
+  when no window can show it, so a covered,
+  minimized, or off-Space window costs
+  nothing.
 
 - Clippings is a new side panel that contains a
   list of items formatted as markdown. They're a
--- a/sources/Drawing/iTermSharedImageStore.h
+++ b/sources/Drawing/iTermSharedImageStore.h
@@ -6,6 +6,8 @@
 //
 
 #import <AppKit/AppKit.h>
+#import <AVFoundation/AVFoundation.h>
+#import <UniformTypeIdentifiers/UniformTypeIdentifiers.h>
 
 NS_ASSUME_NONNULL_BEGIN
 
@@ -18,9 +20,32 @@ NS_ASSUME_NONNULL_BEGIN
 @property (nonatomic, readonly) CGImageRef cgimage;
 // Size of largest rep
 @property (nonatomic, readonly) NSSize scaledSize;
+// When the wrapped file is a video, image holds a placeholder (the poster
+// frame, once it loads asynchronously) and videoURL locates the file so views
+// can play it. Playback happens in iTermImageView; renderers that can only
+// draw still images fall back to the poster frame.
+@property (nonatomic, readonly, nullable) NSURL *videoURL;
+@property (nonatomic, readonly) BOOL isVideo;
 
 + (instancetype _Nullable)withContentsOfFile:(NSString *)path;
 + (instancetype)withImage:(NSImage *)image;
+
+// Container formats AVFoundation can reliably decode for background videos.
++ (NSArray<UTType *> *)videoContentTypes;
++ (BOOL)pathIsVideo:(NSString *)path;
+
+// One muted, looping player is shared by every consumer of this wrapper —
+// AVPlayerLayers in image views and the Metal renderer alike — so a video
+// used in several panes decodes once. Playback runs while at least one
+// consumer holds an interest. Call these on the main queue.
+- (void)retainVideoPlaybackInterest;
+- (void)releaseVideoPlaybackInterest;
+// Nil for non-videos. Created on first access; main queue only.
+@property (nonatomic, readonly, nullable) AVQueuePlayer *videoPlayer;
+// The output vends BGRA pixel buffers for the Metal renderer. Nil until
+// videoPlayer or a playback interest has created the player. Safe to read
+// from the render thread; AVPlayerItemVideoOutput itself is thread-safe.
+@property (atomic, readonly, nullable) AVPlayerItemVideoOutput *videoOutput;
 
 - (instancetype)initWithImage:(NSImage *)image NS_DESIGNATED_INITIALIZER;
 - (instancetype)init NS_UNAVAILABLE;
--- a/sources/Drawing/iTermSharedImageStore.h
+++ b/sources/Drawing/iTermSharedImageStore.h
@@ -7,7 +7,11 @@
 
 #import <AppKit/AppKit.h>
 #import <AVFoundation/AVFoundation.h>
+#import <CoreVideo/CoreVideo.h>
+#import <CoreVideo/CVMetalTextureCache.h>
 #import <UniformTypeIdentifiers/UniformTypeIdentifiers.h>
+
+@protocol MTLDevice;
 
 NS_ASSUME_NONNULL_BEGIN
 
@@ -22,8 +26,9 @@ NS_ASSUME_NONNULL_BEGIN
 @property (nonatomic, readonly) NSSize scaledSize;
 // When the wrapped file is a video, image holds a placeholder (the poster
 // frame, once it loads asynchronously) and videoURL locates the file so views
-// can play it. Playback happens in iTermImageView; renderers that can only
-// draw still images fall back to the poster frame.
+// can play it. iTermImageView plays it with an AVPlayerLayer and the GPU
+// renderer samples frames through copyVideoPixelBufferForHostTime:generation:;
+// consumers that can only draw still images fall back to the poster frame.
 @property (nonatomic, readonly, nullable) NSURL *videoURL;
 @property (nonatomic, readonly) BOOL isVideo;
 
@@ -42,10 +47,32 @@ NS_ASSUME_NONNULL_BEGIN
 - (void)releaseVideoPlaybackInterest;
 // Nil for non-videos. Created on first access; main queue only.
 @property (nonatomic, readonly, nullable) AVQueuePlayer *videoPlayer;
-// The output vends BGRA pixel buffers for the Metal renderer. Nil until
-// videoPlayer or a playback interest has created the player. Safe to read
-// from the render thread; AVPlayerItemVideoOutput itself is thread-safe.
-@property (atomic, readonly, nullable) AVPlayerItemVideoOutput *videoOutput;
+
+// Vends the Metal texture for the video frame that should be visible at
+// hostTime, for GPU renderers drawing on the given device.
+//
+// An AVPlayerItemVideoOutput is single-consumer: copying a pixel buffer marks
+// it acquired, so if each pane's renderer pulled from the output directly, only
+// the first pane to ask would get any given frame and the others would be stuck
+// one frame behind — visible as a flickering seam at split dividers when panes
+// share one background image. Instead this owns the pull, dequeueing at most
+// once per display refresh, and vends one texture per frame so every pane
+// sampling this wrapper in a given refresh samples the identical texture.
+//
+// generation identifies the frame the caller is showing: pass 0 if it has none.
+// Returns NULL when no frame has decoded yet or when the visible frame is still
+// the one identified by *generation — either way the caller should keep showing
+// what it has. Otherwise returns a texture the caller owns (release it with
+// CFRelease, and keep it alive for as long as the GPU may sample it) and sets
+// *generation to identify it.
+//
+// A texture is cached per device, so a second GPU — an eGPU, or the discrete
+// GPU on a dual-GPU Mac — gets its own. Safe to call from any thread; does not
+// create the player, so playback must already be under way (see
+// retainVideoPlaybackInterest).
+- (nullable CVMetalTextureRef)copyVideoMetalTextureForHostTime:(CFTimeInterval)hostTime
+                                                        device:(id<MTLDevice>)device
+                                                    generation:(inout NSInteger *)generation CF_RETURNS_RETAINED;
 
 - (instancetype)initWithImage:(NSImage *)image NS_DESIGNATED_INITIALIZER;
 - (instancetype)init NS_UNAVAILABLE;
--- a/sources/Drawing/iTermSharedImageStore.m
+++ b/sources/Drawing/iTermSharedImageStore.m
@@ -11,6 +11,7 @@
 #import "NSArray+iTerm.h"
 #import "NSImage+iTerm.h"
 #import "NSObject+iTerm.h"
+#import <AVFoundation/AVFoundation.h>
 #import <QuartzCore/QuartzCore.h>
 
 @interface NSFileManager(CachedImage)
@@ -111,13 +112,25 @@
 
 @end
 
+static void *iTermImageWrapperCurrentItemContext = &iTermImageWrapperCurrentItemContext;
+
 @implementation iTermImageWrapper {
     id _cgimage;
     NSMutableDictionary<NSNumber *, NSImage *> *_tilingImages;
     NSMutableDictionary<NSString *, NSBitmapImageRep *> *_reps;
+
+    AVQueuePlayer *_videoPlayer;
+    AVPlayerLooper *_videoLooper;
+    AVPlayerItemVideoOutput *_videoOutput;
+    NSInteger _videoPlaybackInterestCount;
 }
 
+@synthesize videoOutput = _videoOutput;
+
 + (instancetype)withContentsOfFile:(NSString *)path {
+    if ([self pathIsVideo:path]) {
+        return [[self alloc] initWithVideoURL:[NSURL fileURLWithPath:path]];
+    }
     NSImage *image = [[NSImage alloc] initWithContentsOfFile:path];
     if (!image) {
         return nil;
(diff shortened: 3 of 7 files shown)


gnachman (changes requested):
Thanks for this, it is a genuinely nice feature and the shared-decoder design is the right idea. I reviewed it in depth (and traced a few paths through the surrounding renderer/session code). There are a couple of blocking correctness problems, most importantly visible tearing under the GPU renderer when panes share one background image, plus several smaller issues. Requesting changes.

## Blocking

### 1. GPU renderer: cross-pane frame desync / tearing when a background image is shared across panes (confirmed)

This is the most serious one and it reproduces clearly in practice (very noticeable seam tearing).

With `Settings > Appearance > per-pane background images` turned off, every pane resolves `effectiveBackgroundImage` to the same shared `iTermImageWrapper` (via `sessionBackgroundImage` -> `tabBackgroundImage` -> `currentSession.backgroundImage`), which is the intended "one decoder for the whole window." But under Metal, each `SessionView` has its own `iTermMetalDriver` on its own private serial queue (`SessionView.m`, `_driver = [[iTermMetalDriver alloc] ...]`), and each pane's `iTermBackgroundImageRenderer` independently pulls frames from the single shared `AVPlayerItemVideoOutput` in `dequeueVideoTextureForImage`.

`AVPlayerItemVideoOutput` is single-consumer and stateful: `hasNewPixelBufferForItemTime:` reports "new since the last buffer you acquired," and `copyPixelBufferForItemTime:` marks it acquired. So for each newly decoded frame, whichever pane's queue asks first gets the frame; the other panes then see no newer frame, get nil, and keep their previous texture. On any given refresh pane A can be on frame T while pane B is on frame T-1. Because per-pane-off means one continuous image spanning the divider, that shows up as a flickering discontinuity at the split seam, and it gets worse with more panes.

The legacy renderer is immune because per-pane-off draws a single window-level `AVPlayerLayer` (`setSharedBackgroundImage`), one continuous surface with no seams.

Right-altitude fix: dequeue each frame once and share it. Have the wrapper (or a small per-video frame cache) own the per-refresh pull and vend a single `CVMetalTexture` keyed by the acquired item time, so all panes sample the same frame for a given refresh, rather than each renderer racing `copyPixelBufferForItemTime:` on the shared output. Giving each pane its own `AVPlayerItemVideoOutput` would de-race acquisition but multiplies buffer copies and can still drift by a frame on timing.

### 2. iTermSharedImageStore: video output is re-added to each looper replica but never detached from the previous item

In `observeValueForKeyPath:` the handler does `[item addOutput:_videoOutput]` on each `currentItem` change, guarded only by `![item.outputs containsObject:_videoOutput]` (checks the new item). An `AVPlayerItemVideoOutput` may be attached to only one `AVPlayerItem` at a time, and `AVPlayerLooper` keeps live replica items and rotates `currentItem` among them. When the loop wraps, the output is added to the new replica while still attached to the one being left. Depending on whether the old replica is still retained, that either stops the output vending buffers (video freezes after the first loop under Metal) or throws for a double attachment.

Fix: observe with `NSKeyValueObservingOptionOld` too and `removeOutput:` from the previous item before `addOutput:` on the new one.

### 3. Metal renderer: CVMetalTexture keep-alive is one generation too shallow for triple buffering

`textureFromPixelBuffer:` keeps only the current and previous `CVMetalTexture` (`_videoTextureKeepAlive` / `_previousVideoTextureKeepAlive`), but `iTermMetalDriverMaximumNumberOfFramesInFlight == 3` and the `MTLTexture` handed to the transient state does not retain its backing `CVMetalTexture`. A 60fps video rolls the keep-alive every draw, so frame N's texture can be released (when frame N+2 runs) while frame N is still queued on the GPU, and the GPU samples a recycled IOSurface. Result: intermittent flicker/garbage frames or a crash under load.

Fix: keep `iTermMetalDriverMaximumNumberOfFramesInFlight` generations alive (e.g. a small FIFO), or tie release to the command-buffer completion handler.

### 4. Settings: video validation only checks `isPlayable`, so a file with no usable video track is accepted

In `checkVideo:` the gate is `!loaded || !asset.isPlayable`. `isPlayable` is true for a playable asset that has no video track (audio-only `.mov`, or a container whose video track the decoder cannot use). Such a file passes validation, gets written to `KEY_BACKGROUND_IMAGE_LOCATION`, poster generation yields a nil thumbnail, and the background renders black with no error shown.

Fix: also require a video track, e.g. load `tracks` and check `[asset tracksWithMediaType:AVMediaTypeVideo].count > 0`.

## Should fix

### 5. Playback keeps decoding when the window is fully occluded or on another Space

The Metal interest gate is `_useMetal && effective.isVideo` (`updateVideoBackgroundPlaybackInterest`) and the legacy gate is `!isHiddenOrHasHiddenAncestor && window != nil` (`updateVideoPlaybackState`). Neither predicate is set by full occlusion, a background Space, or (as far as I can tell from the code) miniaturization, so the shared player keeps decoding frames nobody can see, burning battery. The test plan says miniaturize pauses, so it is worth double-checking which of these actually stop playback. Related: under Metal with per-pane-off, the window-level `AVPlayerLayer` is occluded by the opaque metal views but stays not-hidden (`hidden = image == nil`, since `iTermTextIsMonochrome()` is always YES on macOS 12), so it holds an interest and decodes needlessly behind Metal.

### 6. Enabling a video on an idle session under Metal can leave it frozen

The Metal video only advances because `updateDisplayBecause:` sets `self.active |= videoBackgroundPlaying` and calls `setNeedsDisplay:`. Nothing kicks that first tick when the terminal is quiet: `setBackgroundImage:` sets the interest but does not force a redraw, and `setUseMetal:` draws two frames then `changeCadenceIfNeeded` reads a still-NO `_active`. So setting/enabling a video on an idle session decodes (holding an interest) but shows a frozen frame until an unrelated event (keystroke, blink) triggers `updateDisplayBecause:`.

### 7. Settings: the source-mode control is not refreshed when switching to a video

The image and folder branches of `applyBackgroundImageSelectionAtPath:` fall through to a shared tail that calls `updateControlForKey:KEY_BACKGROUND_IMAGE_SOURCE_MODE`. The video branch returns early and replicates the tail but omits that one call, so switching a profile from folder-rotation to a single video writes `SingleImage` to the key while the source-mode popup keeps showing folder-rotation until the pane reloads.

### 8. Settings: the energy notice re-nags on re-selecting the already-configured video

The energy warning is shown on the success path of every apply, keyed only by the global silenceable identifier. Re-picking the video that is already configured (a normal action) pops the dialog again until the user silences it. Consider skipping it when the chosen path equals the currently stored location.

### 9. Settings UI does not reflect that not all scaling modes work with videos

Tile silently falls back to Scale to Fill for videos (`desiredVideoGravity`, and the Metal path). The scaling-mode popup still offers Tile with no indication, so the control claims a behavior it does not deliver. The popup should reflect this, for example by disabling the Tile item while a video is the selected background (without silently rewriting the stored mode).

## Minor / description

- The "Known limitations" note about the non-monochrome manual drawing path showing the static poster frame is effectively moot on the macOS 12 deployment target: `iTermTextIsMonochrome()` is hardwired to YES for >= macOS 10.16 (`FutureMethods.h`, "Issue 9209"), so `shouldDrawBackgroundImageManually` collapses to just `[NSView iterm_takingSnapshot]`. The only live consumer is snapshotting, where a still poster frame is the correct result anyway. Worth trimming that limitation, and the "e.g. minimum contrast" example is misleading since minimum contrast does not route through that path.
- `CVMetalTextureCache` is created lazily but never flushed (`CVMetalTextureCacheFlush(cache, 0)` once per frame) and only released in dealloc. Memory hygiene, not a correctness issue.
- The deprecated AVFoundation APIs (`loadValuesAsynchronouslyForKeys:`, `statusOfValueForKey:error:`, `isPlayable`) are the correct choice here since the async `load(_:)` replacements are macOS 13+, so no change needed given the macOS 12 target.

## What looked solid

The playback-interest reference counting is balanced across renderer switches and split panes (same shared wrapper, each holder does its own retain/release, `MAX(0, ...)` guards underflow), the MRC retain/release in `PTYSession` is correct, and the zero-copy `CVMetalTextureCache` approach for the GPU path is the right design. Item 1 above is the main thing standing between this and shipping.

OneTrickSupport:
Thanks for the thorough review — the analysis on item 1 in particular saved me a lot of guessing.

Pushed three commits: `8124aec` for the blockers, `76da0a6` for the rest, and `b5b34e1` for a follow-up on item 9.

All nine items are now verified by hand on a debug build, on an M5 with a 1080p25 clip. Details per item below.

## Blocking

**1. Cross-pane frame desync.** Fixed as you suggested, and I took the "vend a single `CVMetalTexture`" wording literally rather than stopping at the pixel buffer. `iTermImageWrapper` now owns the pull: it dequeues at most once per display refresh and wraps that frame exactly once per GPU, so every pane sharing the video samples the *same* `MTLTexture` for a given refresh, not merely the same pixels. `iTermBackgroundImageRenderer` no longer has a texture cache of its own; it asks for the current frame and gets NULL back when it already has it. The texture cache is keyed by `MTLDevice.registryID` so a second GPU gets its own, and since it's now one cache in one place I added the `CVMetalTextureCacheFlush` you mentioned under Minor, on frame retirement.

Verified at 6 panes with a high-motion clip and per-pane images off: the seam is gone.

**2. Video output re-added without detaching.** Now observing with `NSKeyValueObservingOptionOld` and calling `removeOutput:` on the outgoing item before `addOutput:` on the new one. Verified across many loop wraps.

**3. Keep-alive depth.** Rather than deepening the FIFO to `iTermMetalDriverMaximumNumberOfFramesInFlight`, I tied lifetime to the work: the transient state holds a strong reference to the `CVMetalTexture` it samples, and `iTermMetalFrameData` is released from the command buffer's completion handler. The reference now lives exactly as long as the GPU needs it, so there's no depth constant left to get wrong.

Soaked it at 6 panes with saturated output for 75s at ~100% CPU: no crash, and memory returned to baseline afterwards.

**4. Video validation.** `checkVideo:` loads `tracks` alongside `playable` and requires `[asset tracksWithMediaType:AVMediaTypeVideo].count > 0`. Error text updated to mention a missing video track. Verified with an AAC audio file in a QuickTime container: previously accepted and rendered black, now refused.

## Should fix

**5. Playback while occluded.** Both gates now require the window to be visible, via `occlusionState`, which covers all three cases you listed — fully covered, miniaturized, and on another Space — none of which the view hierarchy exposes. `PTYSession` and `iTermInternalImageView` both take the gate and re-evaluate on `NSWindowDidChangeOcclusionStateNotification`.

To answer your question about which predicate actually stopped playback: none of them did. The Metal gate was `_useMetal && effective.isVideo`, with no visibility term at all, so nothing in that path could have paused anything. The test plan's claim that miniaturizing paused playback was simply wrong, and I've removed it from the PR description.

Measured on the fixed build, one pane, same clip, iTerm2's own process:

| window state | CPU |
|---|---|
| visible, video playing | 2.4 – 2.8% |
| miniaturized | 0.2% |
| covered by another window | 0.2 – 0.4% |

The drop lands within a few seconds of ⌘M and reverses on restore.

**One part of this I deliberately left alone, and I'd like your call on it.** The window-level `AVPlayerLayer` sitting behind opaque Metal views: with the occlusion gate in place there is no longer a wasted *decode* there, because the shared player has to run for the Metal path anyway and both interests now drop together when the window is hidden. What remains is a redundant layer composite while the window is visible. Hiding that view would fix it, but the condition is only safe if *every* pane is using Metal — a mixed-renderer window would lose its background entirely — and that seemed like a worse failure than the cost. Happy to do it if you have a reliable way to express "all panes are on Metal", or to leave it.

**6. Frozen first frame on an idle session.** Taking a playback interest now marks the session active and calls `setNeedsDisplay:`. I used the `active` setter rather than `updateDisplayBecause:` because the latter calls back into `updateVideoBackgroundPlaybackInterest`, and relying on the early-return to break that cycle felt fragile.

**7.** The video branch now calls `updateControlForKey:KEY_BACKGROUND_IMAGE_SOURCE_MODE` before returning.

**8.** The energy notice is skipped when the chosen path equals the stored one.

**9.** Tile is disabled in the scaling popup while a video is selected. The stored mode is untouched, so switching back to a still image restores the user's choice. This needed a second pass (`b5b34e1`): `NSPopUpButton` autoenables its items, so menu validation re-enabled Tile when the popup opened, leaving it clickable while the button drew its title in the disabled color. Turning off `autoenablesItems` makes the disabled state stick.

Note that if a profile already has Tile stored and a video is then selected, the popup shows Tile greyed as the current selection. That seemed more honest than silently rewriting the stored mode, but say the word if you'd rather it displayed the effective mode instead.

## Minor

Trimmed the "Known limitations" note about the manual drawing path — you're right that it's moot at the macOS 12 deployment target — and dropped the misleading minimum-contrast example. Left the deprecated AVFoundation calls as they are.

## Cost

6 panes, 1080p25, per-pane images off, on an M5: iTerm2 ~2.5–3% CPU, a single shared decoder, WindowServer ~14%, GPU ~11%, memory flat across many loop wraps.

I want to be careful not to overclaim on the before/after — I didn't run a properly paired benchmark, so I'd treat the shared-texture change as structurally better (one decode and one texture wrap per frame regardless of pane count) rather than as a measured win. The WindowServer and GPU share is the display-rate redraw inherent to playing a video, which is what item 5 now avoids paying when nothing can see it.

gnachman:
Thanks for the thorough turnaround. The shared-texture rework for item 1 and the lifetime change for item 3 both look like the right altitude. I did a fresh correctness pass over the updated code and found a few things still open, including a couple that sit right on the handlers this round touched. Two of them are thread-safety races and would be intermittent in the field, so they are worth a careful look before shipping.

## Blocking

### A. KVO handler mutates the shared AVPlayerItemVideoOutput without the lock (race)

`iTermImageWrapper observeValueForKeyPath:` (the item-2 fix). Detaching and reattaching the output is now correctly ordered, but the handler reads `_videoOutput` and calls `[previous removeOutput:_videoOutput]` / `[item addOutput:_videoOutput]` without taking `_videoLock`. Meanwhile `latchVideoFrameForHostTime:` calls `hasNewPixelBufferForItemTime:` / `copyPixelBufferForItemTime:` on that same output from the Metal driver's private queue while holding `_videoLock`. Because the KVO side never locks, the lock provides no mutual exclusion here. `AVPlayerLooper` rotates `currentItem` on every loop, so at each loop boundary while the GPU renderer is drawing, the output's single-consumer attachment is reconfigured concurrently with a frame pull. That is a data race on non-thread-safe AVFoundation objects: intermittent crash, exception, or a garbled frame at the wrap. This also violates the invariant documented on `_videoOutput` (off-main access must hold the lock). Taking `_videoLock` around the detach/attach in the handler closes it.

### B. Poster-frame swap races the Metal texture builder (race)

`setPosterFrame:` reassigns `_image` and does `[_reps removeAllObjects]` (plus `_cgimage = nil`, `[_tilingImages removeAllObjects]`) on the main queue with no lock. `bitmapInColorSpace:` reads `_image` and reads and writes `_reps`, and it runs on the metalDriver private queue via `textureFromImage:` while the renderer builds a texture from the 1x1 placeholder (the gap the poster is meant to cover). Before this feature `_image` was assigned once in init and never mutated, so this cross-queue mutation is new. If the async poster lands during a placeholder texture build, it is concurrent unsynchronized access to an ARC object pointer and a non-thread-safe NSMutableDictionary: intermittent crash. The legacy path is safe here because `reallyUpdate` returns early into `loadVideo` for video, but the Metal path is not.

### C. Legacy renderer: N AVPlayerLayers bound to one shared AVQueuePlayer

`iTermInternalImageView loadVideo` attaches an `AVPlayerLayer` to the wrapper's shared `videoPlayer`. The store returns one cached wrapper per path and the wrapper vends a single player, so more than one image view showing the same file all attach their own layer to the same player. An AVPlayer's visual output can only be presented by one AVPlayerLayer at a time, so only one shows moving video and the others go black or freeze. Two reachable repros in the legacy (non-Metal) renderer: two windows using the same video as a whole-window background, or per-pane backgrounds with a split showing the same file. The Metal path was deliberately engineered around this same single-consumer limitation, so the layer path is the one gap left. (One link rests on documented AVFoundation behavior rather than something I could exercise here, so a quick two-layer smoke test would confirm before you fix it.)

## Should fix

### D. Metal renderer never falls back Tile to Fill for a video

Tile handling is `tState.repeat = (_mode == iTermBackgroundImageModeTile)` in `iTermBackgroundImageRenderer` with no `isVideo` override, and the item-9 change deliberately leaves a stored Tile mode in place. So a profile that had Tile stored from a prior still image and then switches to a video renders as a 1:1 `address::repeat` tiled or cropped video under Metal, while the legacy path forces `AVLayerVideoGravityResizeAspectFill`. The two renderers diverge for the same profile, which is the opposite of the "both renderers fall back to Scale to Fill" claim. This also answers your open question (b): displaying Tile greyed is fine once the renderer actually honors the fallback, but today the Metal renderer does not, so the effective mode is genuinely different between paths. The fix belongs in the renderer (force fill for `isVideo` in Tile mode), not only in the menu-item disable.

### E. Video validation lives only in the prefs UI, so the escape-sequence and profile-load paths bypass it

`checkVideo:` and the energy notice are in `ProfilesWindowPreferencesViewController`. But `terminalSetBackgroundImageFile:` and profile load both reach `iTermImageWrapper withContentsOfFile:`, which produces a video wrapper and starts continuous decode for any mp4/mov, with no video-track check and no energy notice. An audio-only or corrupt file set this way renders black with no poster and nothing to explain why, which is exactly the case the item-4 fix was written to reject. (This is not a DoS concern, the escape-sequence path is still behind the existing per-file permission prompt.) The right altitude is to hoist the video-track validation into the shared load path so every entry point gets it, rather than guarding only the settings controller.

## Minor

### F. Poster-frame generation is duplicated verbatim

`iTermImageWrapper loadPosterFrame` and `ProfilesWindowPreferencesViewController loadBackgroundVideoPreviewWithFilename:` build the same AVAssetImageGenerator (appliesPreferredTrackTransform, kCMTimeZero, wrap CGImageRef as NSImage). These are the only two such call sites in the tree, so a future change to one (a non-zero seek because frame 0 is black, a tolerance tweak) has to be made in both. A single class method on `iTermImageWrapper` would serve both, and would also give the escape-sequence path in E a natural home for validation.

### G. Per-frame interest reconciliation dereferences the delegate for every session

`updateVideoBackgroundPlaybackInterest` is called unconditionally from `updateDisplayBecause:` (a per-cadence path) and evaluates `self.effectiveBackgroundImage` (a delegate hop) and a window occlusion read before the `desired == _videoBackgroundInterestImage` early-out. That runs for every session on every display update, including the large majority that have no video background. A cheap upfront guard (bail unless `_useMetal && _backgroundImage.isVideo`) keeps the delegate hop and occlusion read off the cadence path.

---

A meta note rather than a code point: A and D in particular are places where a claimed fix is either incomplete (A left the same handler racy) or rests on an assumption that does not hold in the other renderer (D). It would be worth having your AI do a focused correctness pass over the updated diff, specifically the cross-queue access to `_videoOutput` / `_image` / `_reps` and the Tile-mode-with-video path across both renderers, before the next round. Happy to re-review after.

PR #703, opened 2026-07-20
Fix settings search field focus ring corner artifacts on macOS 26

aashutoshrathi:
On macOS 26 (Tahoe), the system automatically draws an accent ring around the search field capsule when it is first responder. The custom focus ring in iTermPrefsSearchFieldCell.drawWithFrame: extends 4pt below the cell frame (NSInsetRect(cellFrame, -0.5, -4)) to work around a missing or broken focus ring in earlier Tahoe builds.

This extra ring is no longer needed on current Tahoe — it leaves thin angled line fragments protruding below the bottom corners of the capsule fill.

The fix skips the custom ring on macOS 26+, letting the system provide the focus indicator naturally.

### DIFF
--- a/docs/notes-3.7.txt
+++ b/docs/notes-3.7.txt
@@ -2163,3 +2163,6 @@ Bug Fixes:
   user name and password when a site returns
   an HTTP authentication challenge (Basic,
   Digest, or NTLM) instead of failing to load.
+- Settings search field no longer draws a
+  custom focus ring on macOS 26+ that left
+  corner artifacts under the capsule.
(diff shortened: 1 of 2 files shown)


aashutoshrathi:
Closing — bug does not reproduce on current master (3.7 dev). The original issue was on 3.6.11; something in master already resolves it.

gnachman:
Thanks for spotting this! Landed in fad51a375, with one adjustment: the fix is gated on macOS 26.5.1+ rather than all of macOS 26.

The custom ring exists because the system focus ring was actually broken on 26.2 (it hugged the internal editor instead of the capsule, which is what issue 12708 / FB21919006 was about). It was fixed by 26.5.1, where, as you found, the custom ring instead leaves angled fragments below the capsule corners. So on 26.2-26.4 the custom ring is still needed; only on 26.5.1+ do we skip it and let the system draw the ring. Verified on 26.5.1: centered text and a clean native ring with no corner artifacts.

aashutoshrathi:
@gnachman hey, is this related and fixed too.

tab

gnachman:
@aashutoshrathi That's horrible. Would you mind filing a bug and include your plist file as described in the new-issue template? I wonder if you have a custom tabbar height or something that's causing problems. You can file issues at https://iterm2.com/bugs

PR #708, opened 2026-07-29
feat: add Profiles menu search

suikammd:
Adds Search Profiles… to the Profiles menu with a lightweight ProfileListView-based search panel. Reuses the existing Profiles menu open path and includes ModernTests for GUID/open-action behavior.

### DIFF
--- a/ModernTests/iTermProfilesMenuOpenActionTests.m
+++ b/ModernTests/iTermProfilesMenuOpenActionTests.m
@@ -1,0 +1,84 @@
+//
+//  iTermProfilesMenuOpenActionTests.m
+//  ModernTests
+//
+//  Created by OpenAI on 7/23/26.
+//
+
+#import <XCTest/XCTest.h>
+
+#import "iTermController.h"
+
+@interface iTermController (iTermProfilesMenuOpenActionTests)
+- (void)openProfileFromProfilesMenuWithGuid:(NSString *)guid action:(SEL)action;
+@end
+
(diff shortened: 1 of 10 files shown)


gnachman:
Thanks for the PR! Before going further on the implementation, I want to understand the motivation.

The existing "Open Profiles…" window (`iTermProfilesWindowController`, opened via Cmd-O) already embeds the same `ProfileListView` with its built-in search field, and it focuses that search field automatically when the window becomes key (`windowDidBecomeKey:` calls `[tableView_ focusSearchField]`). So today you can open that window and immediately start typing to filter by name or tag, press Enter to open the match (Shift+Enter for a new window), and there's even a "Close after opening" checkbox for dismiss-on-launch.

As far as I can tell the only real differences in this panel are:
- it honors the `openProfilesInNewWindow` preference with Option to toggle (matching the Profiles menu convention) instead of Enter/Shift+Enter,
- it always closes after opening, and
- lighter/utility-window chrome.

Each of those seems achievable as an option on the existing window rather than a second, near-duplicate `ProfileListView` host (which also loses editing, open-all, the tags toggle, and multi-select, and adds a second menu item to keep in sync).

Did you look at the existing Open Profiles window and its search field? What workflow does it not cover for you that made a separate panel the better approach? If there's a real gap (e.g. a Spotlight-style keyboard-only quick launcher), I'd rather we generalize the existing window than maintain two.

PR #709, opened 2026-07-31
Right-to-left (Persian/Arabic/Hebrew) rendering, selection, and copy

nuved:
Right-to-left (Persian, Arabic, Hebrew) rendering, selection, and copy for the
terminal, matched to how macOS / TextEdit behave (with one deliberate exception
for quotation marks, noted below).

## Inert unless right-to-left support is on

Everything here is gated behind the existing **Bidi** preference
(`kPreferenceKeyBidi`). Per-line `BidiDisplayInfo` is only computed when that
preference is enabled, so with it off every mirror/reorder path is a no-op
(`bidiInfo` is nil, `visualX == logicalX`), and the Advanced Settings
introduced here default to off as well. For a left-to-right-only user, this is
a behavioral no-op.

## What it fixes

**Rendering**
- Parentheses, brackets, and other mirror pairs no longer render reversed in
  right-to-left text (`)تهران(` → `(تهران)`), on both the CoreGraphics and the
  Metal (GPU) renderer. Uses CoreText's own bidi resolution, so brackets around
  an embedded Latin run stay correct.
- Guillemets « » ‹ › are drawn **as typed** rather than bidi-mirrored, so
  `«متن»` keeps its marks hugging the text instead of pointing outward
  (`»متن«`). This is a deliberate departure from CoreText/TextEdit; Persian and
  Arabic readers expect the as-typed form. Brackets still mirror. Guillemets
  directly wrapping a Latin island join it, so `«machine learning»` keeps its
  marks on their typed sides too.
- A combining mark merged into a cell no longer perturbs the display order.
  The fathatan riding the lam-alef in `کاملاً` gave the alef's cell an absolute
  position inside the ligature's advance and the word drew scrambled; only a
  cell's base character sets its position now.
- Arabic cursive joining survives a selection or a mid-word SGR color change.
  Runs that differ only in color are shaped as one string with per-range
  colors, and the Metal glyph path shapes the whole row in one call instead of
  one call per background-color run (a selection changes the background, which
  severed the joining exactly when the user selected part of a word).
- Persian and Arabic letters join with the profile's **Use Ligatures** option
  turned on. The no-shaping fast path exists because CoreText is slow, and with
  ligatures enabled it accepted any `isalpha`/`isnumber` character. Those are
  locale-aware and return true for non-Latin letters, so cursive scripts were
  routed onto the path that draws each isolated glyph and every word came out
  disconnected. The fast path is ASCII-only now; everything else is shaped by
  CoreText.
- List markers like `۸)` and `۱۰)` render in place. A digit-plus-paren marker
  is a single left-to-right run whose columns the surrounding right-to-left line
  lays out in reverse. The grid-alignment step anchored each glyph to the
  previous glyph's trailing edge, correct only when glyph order matches display
  order, so the run's glyphs piled onto one column: the `۱` of `۱۰)` printed
  over the next letter and the paren detached to the line's edge. Each column
  now anchors to its own base glyph, which is unchanged for ordinary runs and
  correct for reversed ones.
- A line's layout and its justification use the same base direction. They
  could disagree on a line that opens left-to-right but is mostly Persian (a
  shell prompt holding a pasted Persian sentence), which detached the
  sentence-final period from its sentence.

**Paragraph direction**
- A line lays out right-to-left when it opens with a right-to-left character
  or when its strong right-to-left characters outnumber the left-to-right
  ones. Either signal alone misfires: first-strong lays out
  `Berlin شهری بزرگ در آلمان` left-to-right, and majority alone flips a
  Persian line that quotes a lot of English. An English sentence containing
  one Persian word stays left-to-right.

**Scrollback & TUIs**
- Wrapped right-to-left lines keep their reorder map on every row once scrolled
  into history, so Persian no longer scrambles in `less`, agent CLIs, and other
  full-screen programs.
- `scrollRect` carries the per-line right-to-left annotation with the content,
  so rows scrolled within an alt-screen region aren't left drawn in logical
  order.

**Selection & copy**
- Mouse character selection is **visual**: the drag selects exactly the
  columns under the pointer, the highlight stays contiguous, and when the drag
  ends the visual span decomposes into logical runs so the copy is in reading
  order. This is the TextEdit convention discussed in the review thread
  (an earlier revision stored a logical range, the browser convention, and a
  contiguous drag over mixed-direction text could highlight a discontiguous
  span and made the edges of a right-justified line unreachable). The per-line
  share of a multi-line selection follows each line's direction, so extending
  upward from a right-justified line keeps the anchor line's words.
- Word, line, and smart selection still operate on logical ranges, and copy
  mode remains fully logical.
- Copy extracts the logical text of what was selected: a full sweep of a
  right-justified line yields the sentence with its trailing period attached;
  a partially swept embedded Latin run contributes exactly its selected cells.
- Word/double-click selection no longer double-converts the coordinate (it
  arrives logical from the mouse handler), which landed on the wrong word.
- Selection highlight on the Metal renderer tests the logical cell index, not
  the visual column (which lit the mirror-image cells).
- When "Require Cmd modifier to drag text from a selection" is on, pressing on
  the selection without Cmd starts a new selection instead of dead-ending.
  Re-selecting from the first letter of a right-to-left line always presses on
  the previous selection, which read as selection being frozen.

**Optional (all default off)**
- Treat Latin words as left-to-right islands in right-to-left text.
- Strip zero-width bidi/format characters on paste (also gated behind the Bidi
  preference); line editors such as zsh render ZWNJ as a visible `<200c>`
  because macOS classifies it as a control character.
- Do not reorder right-to-left text in the alternate screen, for full-screen
  programs that already order their own output for display.
- Treat a line as right-to-left when it contains at least N right-to-left
  words. At 1, a shell prompt holding a pasted Persian word right-justifies
  the way the same text renders in command output. Too aggressive for a
  default, but the behavior right-to-left users tend to want day to day.

## Tests

The `ModernTests` bidi suites cover the reorder map on wrapped/padded lines,
combining marks on ligatures, selection highlight, drag selection driven
end-to-end through the real `iTermSelection` (margin drags, trailing-period
drags, mixed-line decomposition, multi-line direction-aware spans), logical
copy of full visual sweeps, mirroring, and the paragraph-direction rules.

## Column-positioned TUI output (agent CLIs)

- Empty cells left by a program that positions words with absolute-column moves
  (`ESC[<n>G`, as some agent CLIs do) held code 0, which stringified to U+0000
  (bidi class BN). A BN next to a number or a ZWNJ perturbed the reorder map and
  words drew with spaces inside them; empty cells are now treated as spaces
  before layout.
- With Latin-run isolation on, a leading list-marker period (`۱. Word`) was
  absorbed into the following Latin island, tearing the period from its number
  so the marker drew after the word (`Word۱ .`). A leading punctuation separated
  from the word by a space now stays outside the island.
- The alternate-screen bidi option is now a single Advanced Setting (the
  redundant `alternateScreenBidi` was folded into `disableBidiInAlternateScreen`).

New `RTLBidiCaseTests` characterize these plus related cases (period after an
English island, percent in a mixed run, combining marks), with a Safari-vs-iTerm2
reference page under `tests/`.

---

*This consolidates the work that was briefly split across #710, #713, #717,
#718, and #719 back into this PR. All five are closed; everything lives here
and this is the only PR that needs review.*


gnachman:
Following up on the selection behavior, because I dug into this and hit something I want your read on.

I tested how the platform actually handles mouse selection in mixed-direction text, and macOS is not internally consistent about it. There are two real, deliberate conventions, and this PR follows one of them. I want us to pick the team on purpose rather than by accident.

Using `سلام Hello دنیا` (on-screen order `donya  Hello  salaam`), drag-selecting from the middle of `Hello` (starting at the second `l`) rightward into the Persian word to its right:

- **TextEdit (native `NSTextView`):** the highlight follows the pointer and stays contiguous on screen. Copy serializes the glyphs you actually swept into reading order, giving `سلام llo`. The `He` you did not drag over is excluded. Visual selection, logical-order copy.
- **Safari, Chrome, and Firefox (I checked all three):** selection is a logical range between the endpoints. The highlight can jump and go discontiguous, and copy returns the whole logical interval, giving `سلام Hello`, including the `He` you did not sweep. This is what the PR does today.

So the split is native macOS text on one side and every major browser on the other. Both are legitimate conventions, so this is a genuine judgment call and not a bug on your part. If anything, matching all three browsers is a defensible thing to have done.

Since you read and write RTL and I don't, I'd rather understand how you weighed this than impose a preference. A few questions:

- How did you analyze the tradeoff? Did you model the selection on browser behavior deliberately, or did it fall out of storing the selection logically for the copy fix?
- As an RTL user, which do you actually reach for and expect day to day, and does it change by context (selecting prose versus selecting a command or a path in a terminal)?
- Is there something better about the browser/logical model for RTL specifically that a non-RTL user like me would miss?

For whatever a single non-expert data point is worth: I personally find the browser behavior hard to use, because the highlight stops tracking the mouse and I can't predict what I'll get. But I hold that loosely and I'm very willing to be told it's just my LTR intuition and that RTL users expect the logical model.

One note regardless of direction: iTerm has copy mode, our keyboard-driven selection and the natural analog of shift-arrow, where macOS itself does logical, discontiguous selection even in native text views. So if we land on visual for the mouse, logical selection still has a clean home there rather than being lost. But I want your analysis first.

(There are a couple of implementation-level issues too, independent of this question, but I'll hold those until we've settled the model.)

nuved:
> Following up on the selection behavior, because I dug into this and hit something I want your read on.
> 
> I tested how the platform actually handles mouse selection in mixed-direction text, and macOS is not internally consistent about it. There are two real, deliberate conventions, and this PR follows one of them. I want us to pick the team on purpose rather than by accident.
> 
> Using `سلام Hello دنیا` (on-screen order `donya Hello salaam`), drag-selecting from the middle of `Hello` (starting at the second `l`) rightward into the Persian word to its right:
> 
> * **TextEdit (native `NSTextView`):** the highlight follows the pointer and stays contiguous on screen. Copy serializes the glyphs you actually swept into reading order, giving `سلام llo`. The `He` you did not drag over is excluded. Visual selection, logical-order copy.
> * **Safari, Chrome, and Firefox (I checked all three):** selection is a logical range between the endpoints. The highlight can jump and go discontiguous, and copy returns the whole logical interval, giving `سلام Hello`, including the `He` you did not sweep. This is what the PR does today.
> 
> So the split is native macOS text on one side and every major browser on the other. Both are legitimate conventions, so this is a genuine judgment call and not a bug on your part. If anything, matching all three browsers is a defensible thing to have done.
> 
> Since you read and write RTL and I don't, I'd rather understand how you weighed this than impose a preference. A few questions:
> 
> * How did you analyze the tradeoff? Did you model the selection on browser behavior deliberately, or did it fall out of storing the selection logically for the copy fix?
> * As an RTL user, which do you actually reach for and expect day to day, and does it change by context (selecting prose versus selecting a command or a path in a terminal)?
> * Is there something better about the browser/logical model for RTL specifically that a non-RTL user like me would miss?
> 
> For whatever a single non-expert data point is worth: I personally find the browser behavior hard to use, because the highlight stops tracking the mouse and I can't predict what I'll get. But I hold that loosely and I'm very willing to be told it's just my LTR intuition and that RTL users expect the logical model.
> 
> One note regardless of direction: iTerm has copy mode, our keyboard-driven selection and the natural analog of shift-arrow, where macOS itself does logical, discontiguous selection even in native text views. So if we land on visual for the mouse, logical selection still has a clean home there rather than being lost. But I want your analysis first.
> 
> (There are a couple of implementation-level issues too, independent of this question, but I'll hold those until we've settled the model.)

The timing is funny: your comment describes the model the PR had until this afternoon, and this afternoon I replaced it, for exactly the reason you named. The highlight stopped tracking the mouse and I couldn't predict what I'd get. That isn't your LTR intuition missing something; I read and write Persian daily and hit the same wall.

How it happened: the logical model wasn't a deliberate choice. It fell out of storing the selection logically for the copy fix, as you guessed. It survived as long as my test lines were pure Persian, where a logical interval and a visual span are the same set of cells. It fell apart on real terminal content, which is never pure: paths, numbers, English words inside Persian sentences. Selecting «کد پستی 12345 است.» from the left lit 345 but not 12. Dragging from the right margin toward the first letter selected the entire row, because the logical end of a right-justified line sits at the visual far left. The trailing period was unreachable from either direction. Each of those had a boundary-condition explanation, and each fix broke a different gesture. When fixes start playing whack-a-mole like that, the model is wrong, not the arithmetic.

So the PR now implements the TextEdit convention, and your سلام Hello دنیا example behaves exactly as you described it there: sweeping from the second l of Hello into the Persian word highlights contiguously under the pointer and copies «سلام llo», reading order, the He excluded. The shape of it: a live character selection stores the visual columns that were actually dragged; the per-line highlight maps them to logical cells through the bidi table, and both renderers consume that one index set; when the drag ends, the visual span decomposes into contiguous logical runs, one subselection each, merged across soft-wrap boundaries so wrapped lines don't grow newlines in the copy. Multi-line drags open the first and last lines toward each line's own reading end, so extending upward from a right-justified line keeps the anchor line's words instead of dropping them. Word, line, and smart selection still operate on logical ranges, and copy mode stays fully logical, which I agree is the natural home for that model.

Your three questions, directly: deliberate now, accidental before. Day to day I reach for visual, and the terminal context strengthens that preference, because what I'm selecting is usually a path or a number I'm aiming at with my eyes, not prose. The one real advantage of the logical model, always-complete intervals in the copy, is available under the visual model by sweeping the whole phrase, and RTL users do that instinctively. The reverse isn't true: under the logical model there was no gesture that produced the visual result at all.

elikogol:
Hi @nuved, I built a focused follow-up on top of this PR for mixed RTL/LTR terminal input, based on real Hebrew and Arabic usage. I opened a draft PR against your `bidi-mirroring` branch here: https://github.com/nuved/iTerm2/pull/1

I’d appreciate your feedback on whether this direction fits your implementation. If you’d prefer changes or a different approach, no worries at all.

nuved:
@elikogol's follow-up is now merged into `bidi-mirroring`, so it is part of this PR as of b914452eb. Summary for this thread: with "auto-detect paragraph writing direction" OFF (the default), a shell command with a quoted Persian, Arabic, or Hebrew argument now keeps the prompt and command anchored on the left while the argument itself reads in natural RTL order. Previously the default configuration rendered mixed payloads in scrambled word order, and natural reading required opting into the experimental paragraph-direction settings.

I reviewed and tested it with real Persian before merging, and fixed three implementation bugs on top (full details in nuved/iTerm2#1): the isolate ran past the closing quote to the end of the line, so `echo 'سلام' && ls` pulled the tail into the RTL block; a bracket-mirroring exemption rendered `(hello)` as `)hello(`; and an English apostrophe could open a phantom quote and rearrange plain prose like `he said 'سلام' to me`. Each fix has a test that failed before it and passes after, and the full ModernTests run is green (4,446 executed, 0 failures), verified on both renderers.

Thanks @elikogol, this closes the biggest usability gap the default configuration had.

gnachman:
Thanks @nuved. This is a real, useful RTL feature and the core is solid. I have landed a rebased and reworked version directly on `master` (commit range `d61f9f9b8..095009a67`, 18 single-purpose commits) rather than merging this branch, because I trimmed several over-scoped pieces, fixed a number of bugs, and added a broad consumer audit. I am closing this PR in favor of those commits, with credit to you. Here is a full accounting of what landed, what I added, and what I removed and why.

## What landed (the worthwhile core)

- Bracket and mirror-pair mirroring in RTL text on both the CoreGraphics and Metal renderers, driven by CoreText's own bidi resolution (`isRTL && mirrorable`, i.e. UBA L4).
- The generated `BidiMirroring.txt` pair table (BMP-only, asserted).
- Reorder LUT plus the combining-mark fix (the fathatan on the lam-alef in `کاملاً` no longer perturbs display order).
- The ASCII-only shaping fast-path fix (this was a pre-existing bug: `isalpha`/`isnumber` are locale-aware and were routing cursive scripts onto the isolated-glyph path).
- List-marker grid alignment (`۱۰)` anchors each glyph to its own column's base glyph).
- Cursive joining that survives a selection or a mid-word SGR color change (one shaped string with per-range colors; the Metal path shapes the whole row at once).
- Paragraph direction by pure first-strong (UBA P2/P3), with a scalar walk so supplementary-plane RTL scripts work.
- Neutral infrastructure: the wrapped-line and scrollback reorder map, the scrollRect RTL flag, NUL-to-space, and the justification guard.
- The selection model reworked so committed subselections are logical while an in-progress bidi character drag stays visual and is decomposed to logical runs on copy.
- The opt-in extras, off by default: strip zero-width bidi/format characters on paste, and the consolidated alternate-screen bidi opt-out.

## What I added beyond this PR

Making selection use logical coordinates touched far more consumers than the PR updated. I audited every reader of the selection and routed each through logical coordinates:

- Accessibility: coordinate-for-point, frame-for-range, set-selected-range, and the IME and VoiceOver read paths.
- Find-on-page shift-extend (the find cursor is visual, so it needed conversion).
- The context-menu and gesture smart-select entries, and drag-endpoint handling.
- The Python API `get_selection`.
- Copy mode (character and line selection now select in logical order regardless of visual direction).
- The Companion selection range (sourced from logical subselections, with the visual span clamped to the line width so end-of-line padding cannot inflate it).

Plus a few fixes discovered during the audit:

- The alternate-screen bidi migration was running from `+[iTermAdvancedSettingsModel initialize]`, which fires before custom-folder (Dropbox) prefs are copied into local defaults, so a copied-down opt-out was silently reverted on upgrade. Moved it into `performMigrations`, which runs after.
- Additional zero-width bidi and format characters added to the paste strip set (the LRE/RLE/PDF/LRO/RLO range, the isolate controls, and ALM).
- CoreText baseline tie-break and a stale-predecessor lookup when a double-width right-half precedes a combining mark.

New test coverage: `CopyModeSelectionDirectionTests`, `BidiSelectionConsumerAuditTests`, `BidiCompanionSelectionTests`, `BidiAccessibilityCoordTests`, `BidiFindOnPageSelectionTests`, `BidiFindOnPageAnchorTests`, `BidiPasteStripTests`, and `BidiAltScreenMigrationTests`.

## What I removed, and why

The guiding principle was to trust the Unicode bidi algorithm and CoreText rather than reimplement or outguess them. Each removed piece was a deviation from that or an unrelated rider.

1. Guillemet mirror-suppression. The PR special-cased `« » ‹ ›` to not mirror. Removed; guillemets now follow UBA L4 like every other mirrorable character. At the terminal's default LTR base a whole-line `«متن»` still reads as typed for free; at RTL base it mirrors like TextEdit and Safari.
2. `isolateRTLSuffixAfterLTRPrefix` (about 150 lines of shell-quote lexing in the display layer to keep LTR command prefixes anchored). Fragile, only ran when detect-paragraph-direction was off (the default), and broke guillemets, curly quotes, and nested cases. Removed; detect-off is now plain LTR-base bidi.
3. The Latin-islands feature (`isolateLatinRunsInRTL` and friends): opt-in LRI-insertion heuristics. Removed along with its setting and tests.
4. The majority and minimum-RTL-words paragraph rules. Reduced to pure first-strong (UBA P2/P3). First-strong alone lays out `Berlin شهری بزرگ در آلمان` LTR correctly, and the majority rule alone misfired on lines that open LTR. Setting removed.
5. Glyph-comparison mirror detection (the "did CoreText draw a different glyph" heuristic). Replaced with `isRTL && mirrorable`. Verified equivalent on the existing mirror tests including the English-bracketing N0 case, and strictly more correct (it fixed math `≤` to `≥`, which the glyph heuristic missed). The heuristic existed to compensate for the now-removed islands distortion.
6. The `cellForIndex`/`deltas` plumbing in the lookup table, an islands artifact. Reverted; the sanitized string is NUL-to-space only.
7. The `iTermCharacterSource` full-backing-store `memset`. This reverted the Issue 12763 CPU optimization (clear only the drawn frame). I restored the optimization. See the open question below.

## One open question for you

For item 7: I reverted the glyph-clear back to clearing only the drawn frame, which reintroduces the bug you were fixing (overflowing cursive or wide-ligature ink, plus the DEBUG one-pixel-overflow abort). Could you share an exact repro (glyph, font, size, and whether the visible symptom is ink bleed or only the DEBUG abort)? The proper fix is likely to clear the glyph's ink bounds including overflow rather than the whole bitmap, but I need the repro to build a regression test. If you open a focused follow-up PR for just that, I will review it.

Thanks again for the contribution. All bidi test suites are green on master.

nuved:
> Thanks @nuved. This is a real, useful RTL feature and the core is solid. I have landed a rebased and reworked version directly on `master` (commit range `d61f9f9b8..095009a67`, 18 single-purpose commits) rather than merging this branch, because I trimmed several over-scoped pieces, fixed a number of bugs, and added a broad consumer audit. I am closing this PR in favor of those commits, with credit to you. Here is a full accounting of what landed, what I added, and what I removed and why.
> 
> ## What landed (the worthwhile core)
> * Bracket and mirror-pair mirroring in RTL text on both the CoreGraphics and Metal renderers, driven by CoreText's own bidi resolution (`isRTL && mirrorable`, i.e. UBA L4).
> * The generated `BidiMirroring.txt` pair table (BMP-only, asserted).
> * Reorder LUT plus the combining-mark fix (the fathatan on the lam-alef in `کاملاً` no longer perturbs display order).
> * The ASCII-only shaping fast-path fix (this was a pre-existing bug: `isalpha`/`isnumber` are locale-aware and were routing cursive scripts onto the isolated-glyph path).
> * List-marker grid alignment (`۱۰)` anchors each glyph to its own column's base glyph).
> * Cursive joining that survives a selection or a mid-word SGR color change (one shaped string with per-range colors; the Metal path shapes the whole row at once).
> * Paragraph direction by pure first-strong (UBA P2/P3), with a scalar walk so supplementary-plane RTL scripts work.
> * Neutral infrastructure: the wrapped-line and scrollback reorder map, the scrollRect RTL flag, NUL-to-space, and the justification guard.
> * The selection model reworked so committed subselections are logical while an in-progress bidi character drag stays visual and is decomposed to logical runs on copy.
> * The opt-in extras, off by default: strip zero-width bidi/format characters on paste, and the consolidated alternate-screen bidi opt-out.
> 
> ## What I added beyond this PR
> Making selection use logical coordinates touched far more consumers than the PR updated. I audited every reader of the selection and routed each through logical coordinates:
> 
> * Accessibility: coordinate-for-point, frame-for-range, set-selected-range, and the IME and VoiceOver read paths.
> * Find-on-page shift-extend (the find cursor is visual, so it needed conversion).
> * The context-menu and gesture smart-select entries, and drag-endpoint handling.
> * The Python API `get_selection`.
> * Copy mode (character and line selection now select in logical order regardless of visual direction).
> * The Companion selection range (sourced from logical subselections, with the visual span clamped to the line width so end-of-line padding cannot inflate it).
> 
> Plus a few fixes discovered during the audit:
> 
> * The alternate-screen bidi migration was running from `+[iTermAdvancedSettingsModel initialize]`, which fires before custom-folder (Dropbox) prefs are copied into local defaults, so a copied-down opt-out was silently reverted on upgrade. Moved it into `performMigrations`, which runs after.
> * Additional zero-width bidi and format characters added to the paste strip set (the LRE/RLE/PDF/LRO/RLO range, the isolate controls, and ALM).
> * CoreText baseline tie-break and a stale-predecessor lookup when a double-width right-half precedes a combining mark.
> 
> New test coverage: `CopyModeSelectionDirectionTests`, `BidiSelectionConsumerAuditTests`, `BidiCompanionSelectionTests`, `BidiAccessibilityCoordTests`, `BidiFindOnPageSelectionTests`, `BidiFindOnPageAnchorTests`, `BidiPasteStripTests`, and `BidiAltScreenMigrationTests`.
> 
> ## What I removed, and why
> The guiding principle was to trust the Unicode bidi algorithm and CoreText rather than reimplement or outguess them. Each removed piece was a deviation from that or an unrelated rider.
> 
> 1. Guillemet mirror-suppression. The PR special-cased `« » ‹ ›` to not mirror. Removed; guillemets now follow UBA L4 like every other mirrorable character. At the terminal's default LTR base a whole-line `«متن»` still reads as typed for free; at RTL base it mirrors like TextEdit and Safari.
> 2. `isolateRTLSuffixAfterLTRPrefix` (about 150 lines of shell-quote lexing in the display layer to keep LTR command prefixes anchored). Fragile, only ran when detect-paragraph-direction was off (the default), and broke guillemets, curly quotes, and nested cases. Removed; detect-off is now plain LTR-base bidi.
> 3. The Latin-islands feature (`isolateLatinRunsInRTL` and friends): opt-in LRI-insertion heuristics. Removed along with its setting and tests.
> 4. The majority and minimum-RTL-words paragraph rules. Reduced to pure first-strong (UBA P2/P3). First-strong alone lays out `Berlin شهری بزرگ در آلمان` LTR correctly, and the majority rule alone misfired on lines that open LTR. Setting removed.
> 5. Glyph-comparison mirror detection (the "did CoreText draw a different glyph" heuristic). Replaced with `isRTL && mirrorable`. Verified equivalent on the existing mirror tests including the English-bracketing N0 case, and strictly more correct (it fixed math `≤` to `≥`, which the glyph heuristic missed). The heuristic existed to compensate for the now-removed islands distortion.
> 6. The `cellForIndex`/`deltas` plumbing in the lookup table, an islands artifact. Reverted; the sanitized string is NUL-to-space only.
> 7. The `iTermCharacterSource` full-backing-store `memset`. This reverted the Issue 12763 CPU optimization (clear only the drawn frame). I restored the optimization. See the open question below.
> 
> ## One open question for you
> For item 7: I reverted the glyph-clear back to clearing only the drawn frame, which reintroduces the bug you were fixing (overflowing cursive or wide-ligature ink, plus the DEBUG one-pixel-overflow abort). Could you share an exact repro (glyph, font, size, and whether the visible symptom is ink bleed or only the DEBUG abort)? The proper fix is likely to clear the glyph's ink bounds including overflow rather than the whole bitmap, but I need the repro to build a regression test. If you open a focused follow-up PR for just that, I will review it.
> 
> Thanks again for the contribution. All bidi test suites are green on master.

Thanks for the detailed write-up. The accounting is fair, and I agree with
most of the trims, in particular collapsing the paragraph rules to pure
first-strong and replacing the glyph-comparison heuristic with
isRTL && mirrorable. The consumer audit caught real gaps I had missed, and
the windowed-selection clamp is a good catch.

Two things from my side.

On the open question in item 7: I'll open a focused PR for the glyph clear
with an exact repro (glyph, font, size, and whether the symptom is ink bleed
or only the DEBUG abort) and a regression test.

And a small request. Because the work landed as new commits rather than a
merge, I don't appear in the repository's history or in the contributors
list, so the only record of the contribution is this closed PR. Would you be
willing to add a credit line to the RTL entries in docs/notes-3.7.txt, in the
same style as the existing "Credit to NSEcho for the find"? I'm not asking
you to rewrite history on master, I understand that isn't reasonable on a
public branch. A line in the release notes would mean a lot to me.

Thanks again for taking the feature seriously.

nuved:
**Attribution for the RTL work landed in d61f9f9b8..095009a67**

Following up on my request of Aug 31 for a credit line. Since the work was landed as new commits rather than merged, none of the 18 commits carries my author field or a `Co-authored-by:` trailer, so the repository history and the contributors graph record none of it. To make the extent concrete, I compared the tree at this PR's head (b914452eb) with the landing commit (095009a67) by git blob hash.

Of the 49 files this PR changed, **24 are byte-identical** in the landed tree. Same content, same blob hash, now under your commits:

| File | Blob hash (identical in both trees) | My commit | Landed in |
|---|---|---|---|
| `ModernTests/BidiCombiningMarkTests.swift` | `dc92575ab6f4` | 93357f413 | ffa168ad2 |
| `ModernTests/BidiDragSelectionTests.swift` | `e0f404c45705` | bfd08c9c1 | 96a822b79 |
| `ModernTests/BidiMirrorSelectionTests.swift` | `fe12137b356a` | 0f8ee532b | 96a822b79 |
| `ModernTests/BidiMirroringTests.m` | `9af7454b69b6` | 0f8ee532b | ffa168ad2 |
| `ModernTests/BidiPaddedHighlightTests.swift` | `607b7138bd0d` | d2965f8d6 | 96a822b79 |
| `ModernTests/BidiSelectionCrashGuardTests.swift` | `7cc64a3b7ad1` | b82b6a77f | 96a822b79 |
| `ModernTests/BidiSelectionHighlightTests.swift` | `e4a59ddbc6ce` | b82b6a77f | 96a822b79 |
| `ModernTests/BidiWordSelectionTests.swift` | `cb821da767e3` | 1481af5a3 | 96a822b79 |
| `ModernTests/iTermLineAttributeTests.swift` | `149330da8c11` | 1e9da7cd4 | 96a822b79 |
| `sources/CharacterSets/iTermCharacterSets.h` | `9a521c107186` | 0f8ee532b | 98682fc3d |
| `sources/CharacterSets/iTermCharacterSets.m` | `1718c0dc81d1` | 0f8ee532b | 98682fc3d |
| `sources/Drawing/iTermBackgroundColorRun.m` | `a6b4e750278e` | b82b6a77f | 96a822b79 |
| `sources/Drawing/iTermMutableAttributedStringBuilder.h` | `c248e34034ac` | 8f940001c | 076890622 |
| `sources/Drawing/iTermMutableAttributedStringBuilder.m` | `e53fe42898c4` | a7ea22030 | 829b09cdd |
| `sources/LineBuffer/LineBlock+SwiftInterop.m` | `f2d01889a6e1` | 3b52ec372 | f0478886e |
| `sources/Selection/SelectionExtraction.swift` | `45c3bc404dae` | 1e9da7cd4 | 96a822b79 |
| `sources/TerminalView/PTYMouseHandler.h` | `f1274a4f8bc4` | b82b6a77f | 96a822b79 |
| `sources/VT100/VT100Grid.m` | `c0b53c979e48` | edd576fdb | aa43e6cc0 |
| `tests/rtl-bidi-reference.html` | `635b087987b5` | 70289a00b | ffa168ad2 |
| `tests/rtl-showcase.sh` | `ad72336e395b` | b5e5cb4de | ffa168ad2 |
| `tests/rtl-showcase.txt` | `929bf6b24c42` | b5e5cb4de | ffa168ad2 |
| `tests/rtl-tui-repro.sh` | `d2c65a5aa947` | edd576fdb | aa43e6cc0 |
| `tests/zwnj-compare.txt` | `24ae6eb15546` | 3d02228fb | 1f4b1bd9a |
| `tools/iTermCharacterSets.m.template` | `fecd0c2b41f2` | 0f8ee532b | 98682fc3d |

Six more files differ from mine by fewer than ten lines (`iTermMetalPerFrameState.m`, `iTermPasteHelper.m`, `BidiCopyLogicalRangeTests.swift`, `PTYTextView+Private.h`, `generate_nscharacterset.py`, the bridging header), and the remaining source files use my version as their base with your edits on top. Anyone can verify a row with:

```
git rev-parse b914452eb:sources/CharacterSets/iTermCharacterSets.m
git rev-parse 095009a67:sources/CharacterSets/iTermCharacterSets.m
```

I'm not asking for history to be rewritten. I'm asking for two things: a credit line on the RTL entries in `docs/notes-3.7.txt`, in the same style as the existing "Credit to NSEcho for the find", and a mention in the 3.7 release announcement. The closed-PR comment is currently the only place my contribution is recorded, and it is not visible from the code, the history, or the release notes.

Thanks.

gnachman:
You're right, and I'm sorry. I should have credited you in 3.7.0 and I didn't. Landing the work as fresh commits meant none of it carried your author field or a `Co-authored-by:` trailer, and you shouldn't have had to document it blob by blob to get that noticed. Thank you for the patience, and for the care you put into the contribution.

Two of your commits were cherry-picked onto the release branch and did ship in 3.7.0:

- the `VT100Grid.m` fix that carries `rtlFound` with row content in `scrollRect`, so right-to-left text no longer scrambles when a full-screen program scrolls part of the screen
- the `LineBlock+SwiftInterop.m` fix for the double-applied offset that left the continuation rows of a wrapped right-to-left line unreordered when drawn from scrollback

The 3.7.0 changelog had no bug-fix entries for either one, so I've added them, each with a credit line to you.

The bulk of your work is not in a stable release yet. The early 3.7.x releases are built off a fork rather than off master, to keep risk down while the release settles, so only a small set of cherry-picks made it over. Everything else is on master and has been in the nightlies since `v20260829-nightly`.

When a release does include the bulk of it, you will be credited at the top of its release notes. That credit is already written into the notes file for the next release.

Thanks again for pushing on this.

PR #710, opened 2026-07-31
Move rtlFound with row content in scrollRect (fixes RTL scrambling in TUIs)

nuved:
## Problem

With right-to-left support enabled, RTL text printed by full-screen programs and AI agent CLIs comes out scrambled (drawn in raw logical order), while the same text printed by `cat` renders correctly. Persian/Arabic/Hebrew users chatting with terminal AI agents see most long paragraphs reversed.

## Cause

`scrollRect:downBy:softBreak:` moves characters (memmove) and external attributes between rows, but not the per-line `rtlFound` metadata. Bidi analysis is gated on that flag: `populateRTLState` skips paragraphs whose lines lack it and actively nulls their display info. So any RTL row moved by a region scroll (DECSTBM margin scrolling, IND at the bottom margin, IL/DL) is left with a stale `rtlFound=NO`, loses its BidiDisplayInfo on the next populate pass, and renders in logical order. TUIs scroll and repaint their regions constantly; `cat` never does, which is why plain output was unaffected.

## Fix

Carry `rtlFound` with the content in the move loop: copied exactly for a full-width move, ORed in for a partial-width move (a stale YES only costs a wasted bidi analysis; a stale NO renders RTL text backwards).

## Testing

- New `ModernTests/BidiTUIRepaintTests.swift` replays TUI write patterns end to end: repaint cycles (cursor up + EL + rewrite), chunked mid-row streaming, erase-and-rewrite split across frames, region scrolls (grid-level metadata check plus screen-level populate check — both red before the fix), scrollback commits, and a renderer-consistency test that runs the real `synchronizeWithConfig` cadence and asserts the SCA handed to the renderer carries bidi info matching its own content.
- Note: the shared test `FakeSession` never implemented `screenRestore(_:)` (so `_sharedStateCount` never returned to zero and all syncs short-circuited) or `screenUpdateDisplay(_:)`. The new tests use a subclass implementing both the way PTYSession does; the existing fixture is unchanged.
- `tests/rtl-tui-repro.sh` reproduces the bug visually with plain escape sequences (repainted and scrolled Persian paragraphs vs. plain-printed controls).
- Full VT100GridTests suite (158 tests) passes with the change.

### DIFF
--- a/ModernTests/BidiTUIRepaintTests.swift
+++ b/ModernTests/BidiTUIRepaintTests.swift
@@ -1,0 +1,338 @@
+//
+//  BidiTUIRepaintTests.swift
+//  ModernTests
+//
+//  A TUI (like an AI agent CLI) repaints its output in place: cursor up,
+//  erase line, rewrite, then eventually the rows scroll into history. Rows
+//  written this way must get the same bidi treatment as rows printed once
+//  and never touched. These tests replay that write pattern at the
+//  VT100ScreenMutableState level, running populateRTLStateIfNeeded between
+//  cycles the way the per-frame sync does, and assert that the per-row
+//  BidiDisplayInfo survives each cycle.
+//
+
+import XCTest
+@testable import iTerm2SharedARC
+
+// FakeSession leaves screenRestore(_:) and screenUpdateDisplay(_:)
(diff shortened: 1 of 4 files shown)


nuved:

PR #711, opened 2026-08-01
Fix Kitty keyboard protocol SET behavior

rschmitt:
The SET form should modify the flags on the current stack frame, not destroy the entire stack. Compare the Ghostty [implementation](https://github.com/ghostty-org/ghostty/blob/4d605bf0d819df901a0332bbb320dc849fdd82e4/src/terminal/kitty/key.zig#L27):

```zig
    /// Perform the "set" operation as described in the spec for
    /// the CSI = u sequence.
    pub fn set(
        self: *FlagStack,
        mode: SetMode,
        v: Flags,
    ) void {
        switch (mode) {
            .set => self.flags[self.idx] = v,
            .@"or" => self.flags[self.idx] = @bitCast(
                self.flags[self.idx].int() | v.int(),
            ),
            .not => self.flags[self.idx] = @bitCast(
                self.flags[self.idx].int() & ~v.int(),
            ),
        }
    }
```

### DIFF
--- a/sources/VT100/VT100Terminal.m
+++ b/sources/VT100/VT100Terminal.m
@@ -2511,17 +2511,30 @@ static BOOL VT100TokenIsTmux(VT100Token *token) {
 
         case VT100CSI_SET_KEY_REPORTING_MODE:
             self.dirty = YES;
-            [self.currentKeyReportingModeStack removeAllObjects];
-            switch (token.csi->p[1]) {
-                case 1:  // all set bits are set and all unset bits are reset
-                    _keyReportingFlags = token.csi->p[0];
-                    break;
-                case 2:  // all set bits are set, unset bits are left unchanged
-                    _keyReportingFlags |= token.csi->p[0];
-                    break;
-                case 3:  // all set bits are reset, unset bits are left unchanged
-                    _keyReportingFlags &= ~token.csi->p[0];
-                    break;
+            {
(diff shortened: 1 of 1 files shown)


rschmitt:
Here's a minimal reproducer that exposes the difference in behavior between iTerm2 and Ghostty:

```bash
#!/bin/bash

echo "=== Kitty Keyboard Protocol SET-form persistence test ==="
echo ""
echo "Step 1: PUSH flags=1 (enable protocol)"
printf '\033[>1u'

echo "Step 2: SET flags=1 on current entry (simulates resize re-assert)"
printf '\033[=1;1u'

echo "Step 3: POP (should fully disable protocol)"
printf '\033[<u'

echo ""
echo "Protocol should now be OFF. Type Ctrl-C then Enter:"
echo "  - If you see '^C' → correct (protocol disabled)"
echo "  - If you see '^[[99;5u' → BUG (SET-form flags persisted after POP)"
echo ""
echo "Reading one line (if stuck, close the tab and run: printf '\\033[=0u'):"
read -r line
echo ""
echo "Raw input: $(printf '%s' "$line" | cat -v)"

# Clean up in case the protocol is still active
printf '\033[=0u'
```

gnachman:
Thanks for the fix, @rschmitt! Nice catch matching the SET semantics to the spec (and the Ghostty reference was helpful). Merged. 🎉

PR #713, opened 2026-08-03
Option: treat Latin words as left-to-right islands in RTL text

nuved:
Stacked on #709 (the first commit here is that PR's mirroring change; review the second commit, "Add option to treat Latin words as LTR islands in RTL text").

In a right-to-left paragraph the bidi algorithm reorders neutral characters and mirrors brackets around English exactly as it does around Persian. That's correct per Unicode, but in a terminal full of English identifiers, file paths, and code mixed into Persian or Arabic it reads wrong: «(English)» becomes «)English(», and ~/projects loses its order.

This adds an Advanced setting, off by default, "Treat Latin words as left-to-right islands in right-to-left text." When it's on, each run of non-space ASCII containing a Latin letter is wrapped in isolate controls (LRI/PDI) before layout, so CoreText lays those runs out as left-to-right islands. English words, multi-word phrases, paths, and code keep their order and un-flipped brackets; the surrounding right-to-left text is untouched. Guillemets keep their natural glyph, so « still opens on the right in Persian.

`BidiDisplayInfo` now threads an explicit UTF-16-index → source-cell map (inserted controls map to no cell) rather than the delta formula, so the extra controls don't disturb cell indexing, the lookup table, RTL runs, or mirror detection.

The setting deviates from standard bidi on purpose, which is why it's opt-in and off by default. It's aimed at terminal use, where English tokens are conceptually left-to-right islands. `tests/rtl-showcase.sh` prints every case (single and multi-word English, code, paths, quotes, Persian brackets) for manual review.


nuved:

PR #715, opened 2026-08-05
Fix Tahoe dark tab chrome ignoring custom tab height

maskshell:
## Summary
- On macOS Tahoe dark themes, selected-tab chrome (`TahoeDarkLeftTab` / `Mid` / `Right`) was drawn at the images’ intrinsic size instead of the pill rect, so left/right tabs with a custom **Default tab bar height** showed a fixed-height foreground over a taller cell (masked only when an opaque custom tab color painted over the caps).
- Cap/mid images now scale to the pill height (with width clamped so tall narrow left tabs cannot overlap caps). Pill fill/stroke/progress ring use a true stadium radius (`height/2`) so curvature matches the scaled chrome.
- Clarify the Advanced Pref help text: top tabs ignore `defaultTabBarHeight` on Tahoe; left/right use `max(26, value)`.

## Test plan
- [ ] macOS 26, Theme Dark (or Dark High Contrast), tabs on the **left**
- [ ] Advanced Prefs → set **Default tab bar height** above 26 (e.g. 36 or 40)
- [ ] Without custom tab color: selected tab chrome should match cell height (no dual-layer, no gray crescents at corners)
- [ ] Tall left tabs at default ~50pt bar width: caps must not overlap in the middle
- [ ] With Use Tab Color: still looks correct; selected fill aligned with chrome
- [ ] Tabs on **top**: pill corners remain stadium-shaped (~12pt), no squaring vs previous Tahoe look
- [ ] Light Tahoe theme: no regression (stroke outline, no cap images)


ThirdParty/PSMTabBarControl/source/PSMTahoeTabStyle.swift
```diff
@@ -439,6 +439,13 @@ class PSMTahoeTabStyle: NSObject, PSMTabStyle {
     // bar inside it, so it never crowds the tab title or status subtitle.
     private var progressRingWidth: CGFloat { 2 }
 
+    /// Corner radius for a drawn pill, derived from the pill's own height so it
+    /// stays stadium-shaped when cell height changes (vertical tabs + custom
+    /// `defaultTabBarHeight`).
+    private func pillCornerRadius(for rect: NSRect) -> CGFloat {
+        max(0, rect.height / 2.0 - 2.5)
```
gnachman: `pillCornerRadius` regresses the corner radius of every Tahoe pill, including the untouched horizontal top-tab path. The old `barRadius - 2.5` = 11.5 was calibrated against the 28pt bar height, but this rect is only 24pt tall (36pt container minus 8pt insets, minus 1 in `adjustedCellRect`, minus 3 in `backgroundRect`), so `rect.height/2 - 2.5` = 9.5. That squares the corners of every horizontal selected-tab fill, the light-theme stroke, the tab-color tint, and the progress ring (`progressBarClipPath` uses this helper too) by about 2pt; default-height vertical pills drop from a true stadium (11) to 8.5. The comment claims it stays stadium-shaped, but `height/2 - 2.5` is never a stadium (a stadium is `height/2`). Suggest dropping or recalibrating the 2.5 so the horizontal path stays at 11.5.

ThirdParty/PSMTabBarControl/source/PSMTahoeTabStyle.swift
```diff
@@ -835,8 +855,8 @@ class PSMTahoeTabStyle: NSObject, PSMTabStyle {
                                           isHighlighted: Bool) {
         backgroundColorSelected(selected, highlightAmount: isHighlighted ? 1.0 : 0.0).set()
         
-        let radius = barRadius - 2.5
         let rect = backgroundRect(for: cellFrame)
+        let radius = pillCornerRadius(for: rect)
```
gnachman, line 848: In the dark theme this fill radius (9.5, per the `pillCornerRadius` note) no longer matches the corner curvature baked into the cap images that `drawCellOutline` scales onto the same rect (a stadium, about `height/2`). The squarer fill pokes past the caps rounded corners, leaving gray crescents (or, with a per-tab color, colored crescents) at all four corners of a selected dark tab. Rendering the real TahoeDark cap assets at the two radii confirms roughly 1 to 2pt of fill leaking beyond the chrome. Fixing the radius formula resolves this as well.

ThirdParty/PSMTabBarControl/source/PSMTahoeTabStyle.swift
```diff
@@ -1813,17 +1833,26 @@ class PSMTahoeDarkTabStyle: PSMTahoeTabStyle {
...
+        // dual-height selected tab that vanished only when an opaque tab color
+        // painted over the caps.
+        let left = Self.leftTabCap
+        let scale = left.size.height > 0 ? rect.height / left.size.height : 1.0
+        let capWidth = left.size.width * scale
+
+        let leftDest = NSRect(x: rect.minX, y: rect.minY, width: capWidth, height: rect.height)
+        left.draw(in: leftDest)
+
+        let rightDest = NSRect(x: rect.maxX - capWidth, y: rect.minY, width: capWidth, height: rect.height)
```
gnachman, line 1835: Scaled caps can overlap because only `midDest` is clamped. `capWidth` is about `(2/3) * rect.height` with no bound against `rect.width`, and `leftDest`/`rightDest` keep the full `capWidth`, so when `2 * capWidth > rect.width` the two cap images overprint in the middle (and `rightDest.origin.x` goes left of `rect.minX`). This is reachable with left tabs: at the 50pt minimum vertical bar width the pill is about 42pt wide, so overlap begins once the tab height passes about 31.5pt (Default tab bar height around 36), exactly the large-height case this PR targets. On master the fixed 16pt caps could never overlap given the 42pt floor, so this is a new regression. Clamp `capWidth` to `rect.width / 2` before building the three rects.

ThirdParty/PSMTabBarControl/source/PSMTahoeTabStyle.swift
```diff
@@ -579,7 +586,20 @@ class PSMTahoeTabStyle: NSObject, PSMTabStyle {
         }
     }
 
-    private let barHeight = 28.0
+    /// Drawn pill height inside the horizontal tab-bar container (36pt). Vertical
+    /// tabs use the live cell height from `tabBar.height` so chrome tracks
+    /// Advanced Pref `defaultTabBarHeight` instead of staying pinned at 28.
+    private static let horizontalBarHeight: CGFloat = 28.0
+
+    private var barHeight: CGFloat {
```
gnachman: This vertical branch is dead code. After this PR migrated `progressBarClipPath` and `drawCellBackgroundSelected` to `pillCornerRadius`, the only remaining reader of `barHeight`/`barRadius` is `clippingPath`'s `if orientation == .horizontalOrientation` block, so this `.verticalOrientation` branch can never execute and the property always returns `horizontalBarHeight` (28). `private let barHeight = 28.0` would be behaviorally identical. That means the doc comment here, the release-note line ("vertical bar height tracks the container"), and part of the PR description document behavior that never runs. If the branch were made live, it would also round-trip through `tabBar.height` to recover a value the style already computes as `tabBarHeight` (max(26, defaultTabBarHeight)). Suggest reverting `barHeight` to the constant and trimming the release note; the real fix is carried entirely by `pillCornerRadius` and the cap scaling.

ThirdParty/PSMTabBarControl/source/PSMTahoeTabStyle.swift
```diff
@@ -1813,17 +1833,26 @@ class PSMTahoeDarkTabStyle: PSMTahoeTabStyle {
...
-        
-        dest.origin.x = rect.maxX - image.size.width
-        Self.rightTabCap.draw(in: dest)
-        
-        dest.origin.x = rect.minX + image.size.width
-        dest.size.width = rect.width - (2 * image.size.width)
-        Self.tabMid.draw(in: dest)
+        // Scale the 9-slice-like cap/mid images to the pill rect. Previously
+        // dest.size was image.size, so chrome stayed at the asset's intrinsic
+        // height while the cell/backing grew with defaultTabBarHeight — a
```
gnachman: This comment contains an em dash (U+2014, between `defaultTabBarHeight` and `a`); the convention here is to avoid em dashes in comments, so replace it with a comma or a period. The comment also narrates change history ("Previously dest.size was image.size ..."); once merged, a comment that states the constraint (the caps must scale to the pill rect height so the chrome matches the cell height) will serve future readers better than the before/after story.

gnachman (commented):
Automated review of this PR. The cap-scaling approach is right, but the new radius formula regresses geometry on paths the change did not intend to touch, and the `barHeight` change turns out to be inert. Findings are inline. One additional cleanup that has no changed line to attach to:

- `drawCellOutline(path:rect:radius:)`: the `radius` parameter is now unused by both implementations (the base only strokes `path`; the dark override uses only `rect`). Since this PR already touches the call site and the override, consider dropping the parameter.

Two help-text accuracy concerns I checked turned out fine: compact windows never carry Tahoe-style tabs, and no reachable path delivers a vertical Tahoe height below the 26pt floor, so the revised advanced-setting description is accurate.

maskshell:
Addressed the review:

1. **`pillCornerRadius`** → `height/2` (true stadium). Horizontal ~24pt pills go to 12 (was 11.5 with `barRadius - 2.5`), so fill/stroke/tint/progress ring match scaled dark caps and no longer leak crescents.
2. **Cap overlap** → `capWidth = min(scaled, rect.width / 2)`.
3. **`barHeight`** → reverted to `private let barHeight = 28.0`; release note no longer claims vertical barHeight tracking (that branch was dead).
4. **Comment** → constraint-only, no em dash / history narration.
5. **`drawCellOutline`** → dropped unused `radius` parameter on both implementations and the call site.

ThirdParty/PSMTabBarControl/source/PSMTahoeTabStyle.swift
maskshell: Done in db4477a18.

`pillCornerRadius` is now a true stadium: `max(0, rect.height / 2.0)` (no `- 2.5`). For the ~24pt horizontal pill that is 12, within ~0.5pt of the old calibrated 11.5, without squaring top-tab corners. Vertical pills scale with cell height the same way.

ThirdParty/PSMTabBarControl/source/PSMTahoeTabStyle.swift
maskshell: Done in db4477a18.

Same stadium `height/2` radius is shared by the selected fill, tab-color tint, and progress ring, so it matches the curvature baked into the scaled dark cap images. That should remove the gray/colored crescents at the corners.

ThirdParty/PSMTabBarControl/source/PSMTahoeTabStyle.swift
maskshell: Done in db4477a18.

`capWidth` is now `min(left.size.width * scale, rect.width / 2.0)` before building left/right/mid dest rects, so tall left tabs at the ~50pt minimum bar width can no longer overprint caps in the middle.

ThirdParty/PSMTabBarControl/source/PSMTahoeTabStyle.swift
maskshell: Done in db4477a18.

Reverted to `private let barHeight = 28.0`. The vertical branch was dead after the radius move, as you noted. Release note and PR description no longer claim that vertical `barHeight` tracks the container; the real fix is cap scaling plus stadium `pillCornerRadius`.

ThirdParty/PSMTabBarControl/source/PSMTahoeTabStyle.swift
maskshell: Done in db4477a18.

Replaced with a constraint-only comment (caps must scale to the pill height so chrome matches the cell). No em dash and no before/after narration.

gnachman:
Thanks for the fix, @maskshell! Nicely diagnosed and cleanly done. Scaling the cap and mid images to the pill height and deriving the corner radius from height/2 is exactly the right approach here, and it makes the fill, stroke, progress ring, and dark-theme caps all share one curvature. I built and reviewed it locally and it looks correct. Merged, and Iʼve added you to the Credits page. 🎉

PR #716, opened 2026-08-05
Right-to-left (Persian/Arabic/Hebrew) rendering, selection, and copy

nuved:
**Draft — work in progress, opening for early feedback.**

This adds right-to-left (Persian, Arabic, Hebrew) rendering, selection, and
copy handling to the terminal, matched to how macOS / TextEdit behave.

## Inert unless right-to-left support is on

Every change here is gated behind the existing **Bidi** preference
(`kPreferenceKeyBidi`). Per-line `BidiDisplayInfo` is only computed when that
preference is enabled, so with it off:

- every mirror/reorder path is guarded by `bidiInfo` (nil → no-op) or
  `mirrorsSourceCell:` (returns NO on nil);
- the logical-vs-visual selection swaps reduce to identity, because with no LUT
  `visualX == logicalX` for every cell;
- the `scrollRect` `rtlFound` carry only moves a per-line flag that nothing
  reads when bidi analysis is gated off.

For a left-to-right-only user with the preference off, these commits are
behavioral no-ops. The two Advanced Settings introduced here both default to
**off** as well.

## What it fixes

**Metal (GPU) renderer**
- RTL selection highlight landed on the mirror-image cells (empty trailing
  padding on the left); now tests the logical cell index, matching the
  CoreGraphics path.
- Parentheses and other mirror pairs rendered reversed in RTL text
  (`)تهران(` instead of `(تهران)`). Applies UBA rule L4 using CoreText's real
  bidi resolution, so brackets around an embedded LTR run stay un-mirrored.

**Scrollback & TUIs**
- Wrapped RTL lines lost their reorder map once scrolled into history, so
  Persian scrambled in `less`, Claude Code, and other full-screen apps. The map
  now survives the wrap/history slice.
- `scrollRect` now carries `rtlFound` with the row content, so RTL rows scrolled
  within the alt-screen aren't left drawn in logical order.

**Selection & copy**
- Mouse clicks map through the bidi reordering, so a click selects the word
  under the cursor rather than its mirror.
- Copy extracts the logical selection range, not the cells that happen to sit
  under the visual span.
- Word selection no longer double-converts the coordinate (it arrives logical
  from the mouse handler), which was landing on the wrong word on RTL lines.

**Optional (both default off)**
- Treat Latin words as LTR islands inside RTL text.
- Strip zero-width bidi/format characters on paste (also gated behind the Bidi
  preference) — line editors such as zsh render ZWNJ as a visible `<200c>`
  because macOS classifies it as a control character, which fills a Persian
  command line with `<200c>`.

## Tests

New `ModernTests` bidi suites cover the reorder map on wrapped/padded lines,
selection highlight, logical copy range, drag selection, and mirroring.


nuved:
Superseded by focused, rebased PRs: #709 (bracket + Metal-paren mirroring), #710 (scrollRect rtlFound), #713 (Latin-islands option), #717 (selection/highlight/copy), #718 (wrapped-scrollback reorder map), #719 (opt-in zero-width strip on paste). Closing this bundle in favor of those.

PR #717, opened 2026-08-05
RTL selection, highlight, and copy in the terminal

nuved:
**Draft — work in progress.**

Selection, highlight, and copy fixes for right-to-left (Persian/Arabic/Hebrew)
text in the terminal. Gated behind the existing **Bidi** preference: with it
off, per-line `BidiDisplayInfo` is nil, so `visualX == logicalX` for every cell
and each change below is a no-op. No behavior change for left-to-right-only
users.

## What it fixes

- **Mouse selection maps through the bidi reordering.** A click is converted
  from its visual column to the logical cell, so character drag, word/line/smart
  selection, the highlight, and copy all operate in the logical space the
  selection model already uses.
- **Selection highlight on the Metal (GPU) renderer** tested the visual column
  against a logically-stored selection, lighting the mirror-image cells (the
  empty trailing padding on the left of a right-to-left line). It now tests the
  logical cell index, matching the CoreGraphics path.
- **Copy extracts the logical selection range**, not the cells that happen to
  fall under a visual span. A partial selection on a mixed line stays a
  contiguous string — an English word embedded in Persian is copied whole
  instead of split into visually-adjacent pieces.
- **Word/double-click selection no longer double-converts the coordinate.** The
  click already arrives logical from the mouse handler, but `rangeForWordAt`
  converted again (and converted the result back to visual), landing on a
  different word on right-to-left lines — double-click «معلم» and «تومن» got
  selected. It now operates on the logical location directly and returns a
  logical range.

## Tests

New `ModernTests` suites: drag-selection highlight, padded (TUI/alt-screen)
highlight, and logical copy range.

### DIFF
--- a/ModernTests/BidiCopyLogicalRangeTests.swift
+++ b/ModernTests/BidiCopyLogicalRangeTests.swift
@@ -99,4 +99,28 @@ class BidiCopyLogicalRangeTests: XCTestCase {
         XCTAssertTrue(text.contains("ر") || text.contains("ز"),
                       "line 1 (روز خوش دوست) dropped from multi-line copy; got >>>\(text)<<<")
     }
+
+    // A partial selection covering an English word embedded in RTL must copy the
+    // whole word contiguously — «Berlin» must not come out as «B … شهری» (the
+    // visually-adjacent-but-logically-split result the user hit).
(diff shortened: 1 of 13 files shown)


nuved:

PR #718, opened 2026-08-05
RTL: keep the reorder map on wrapped lines in scrollback

nuved:
**Draft — work in progress.**

> Depends on #710 (the `scrollRect` rtlFound fix): this branch is stacked on it
> and its test reuses that harness, so the diff currently shows both commits.
> Once #710 lands, a rebase drops the shared commit and only the scrollback fix
> remains.

A wrapped line's bidi is stored once for the whole raw line in the line buffer.
`_wrappedLineWithWrapWidth:` already slices out the requested display row's
portion (range `[offset, offset+width)`, padded to width). The Swift accessor
then split that result a second time by `lineOffset` — which equals that same
offset — so every continuation row asked for `[width, …)` of a slice that only
has `width` cells and got nothing. The row came back with no reorder map, so a
right-to-left line drawn from scrollback rendered its continuation rows in
logical order (reversed/scrambled) while the on-grid copy was fine. This scrambled
Persian in `less`, full-screen agent CLIs, and other programs once their output
scrolled into history.

Return the already-sliced per-row bidi instead of re-splitting it. Gated behind
bidi support, so it is inert when right-to-left support is off.

## Tests

Adds a test that scrolls a wrapped Persian paragraph into history and requires
each row to render identically from scrollback and from the grid; it fails
before this change (continuation row loses its map) and passes after.

### DIFF
--- a/ModernTests/BidiTUIRepaintTests.swift
+++ b/ModernTests/BidiTUIRepaintTests.swift
@@ -1,0 +1,338 @@
+//
+//  BidiTUIRepaintTests.swift
+//  ModernTests
+//
+//  A TUI (like an AI agent CLI) repaints its output in place: cursor up,
+//  erase line, rewrite, then eventually the rows scroll into history. Rows
+//  written this way must get the same bidi treatment as rows printed once
+//  and never touched. These tests replay that write pattern at the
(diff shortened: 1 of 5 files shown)


nuved:

PR #719, opened 2026-08-05
Paste: opt-in strip of zero-width bidi/format characters

nuved:
**Draft — work in progress.**

Adds an opt-in Advanced Setting to strip zero-width bidi/format characters
(ZWNJ, ZWJ, ZWSP, LRM/RLM, word joiner, soft hyphen, ZWNBSP) when pasting.

## Why

Line editors such as zsh/readline and some TUIs render a zero-width format
character as a visible `<200c>`, because macOS classifies U+200C (the Persian
half-space, ZWNJ) as a control character (`iswcntrl(0x200c) == 1`). Pasting
Persian at a shell prompt then fills the command line with `<200c>` markers. The
characters are correct in the pasted data; it's the line editor that draws them.

## Gating

Off by default, and double-gated: it only runs when the **Bidi** preference is
enabled *and* the Advanced Setting is on. These characters are meaningful (the
half-space ZWNJ carries orthographic meaning, the marks carry direction), so
removing them changes the text — it's an explicit opt-in, inert otherwise.


nuved:

PR #720, opened 2026-08-06
Fix typos for word punctuation

michaelbnewman:
Fix multiple occurrences of typo `puncutation` to `punctuation`

* Two corrections in: api/library/iterm2/gen_binding.py
* Two corrections in: api/library/iterm2/iterm2/binding.py

Ironically, no actual punctuation corrected.


gnachman:
Thanks for the fix, @michaelbnewman! Merged. Appreciate you catching the typo in both the generator and the generated binding so it stays consistent. (And yes, the irony of no actual punctuation being corrected is noted.)

PR #721, opened 2026-08-06
Add an option to make the left-side tab bar scrollable

Conceptron:
## Problem

iTerm2 has a vertical tab bar (Appearance > General > Tab Bar Location > Left), but it cannot be scrolled. Tabs that do not fit are moved into the "More tabs" overflow menu, so once you have a lot of tabs open most of them are only reachable through that menu.

The layout in `PSMTabBarControl.m` `reallyUpdate:` walks down from a fixed top margin and breaks at the bottom edge:

```objc
for (int i = 0; i < cellCount; ++i) {
    if (currentOrigin + cellRect.size.height <= [self frame].size.height) {
        [newOrigins addObject:[NSNumber numberWithFloat:currentOrigin]];
        currentOrigin += cellRect.size.height;
    } else {
        // Out of room, the remaining unpinned tabs go into overflow.
        ...
        break;
    }
}
```

There is no scroll offset anywhere in that file, and `scrollWheel:` / `NSScrollView` / `scrollOffset` have no hits in it.

Existing request for this: https://gitlab.com/gnachman/iterm2/-/issues/10944

## What this does

Adds an advanced setting, "Make the left-side tab bar scrollable?" (`scrollableVerticalTabBar`, default off). When it is on and the tab bar is on the left:

* Every cell gets an origin shifted by a scroll offset, instead of the layout stopping at the bottom edge. Nothing goes into the overflow menu, so the scroll wheel can reach every tab.
* `scrollWheel:` moves the offset, clamped at both ends. Precise deltas (trackpad, Magic Mouse) are used as-is; line deltas from a notched wheel are scaled to one tab per line. When everything already fits, the event is passed through to `super` unchanged.
* Selecting a tab that is off-screen scrolls it into view, for example when cmd-shift-[ walks past the top, or when the selection comes from outside the bar.
* `clipsToBounds` is set explicitly, because `NSView` defaults it to `NO` on macOS 14 and later and the scrolled-away cells would otherwise paint over the window chrome.
* The add-tab button is hidden when the tabs run past the bottom, since there is no overflow menu for it to sit against and it would land off-screen.

Default is off, so behaviour is unchanged unless it is turned on.

Files touched: `ThirdParty/PSMTabBarControl/source/PSMTabBarControl.m`, `sources/Settings/iTermAdvancedSettingsModel.{h,m}`.

## Testing

Built Deployment on macOS 26.5, Xcode 26.2, arm64. Tested with 18 tabs in a 420pt tall window:

* Setting off: unchanged, tabs overflow into the menu exactly as before.
* Setting on: the overflow button disappears, the trackpad and scroll wheel reach every tab, cells clip cleanly at the top and bottom edges, and selecting tab 1 scrolls the bar back to the top.

I run with roughly 200 tabs in a left sidebar day to day, which is what prompted this.

## One thing worth flagging

With the setting on, layout gives an origin to every cell rather than only the ones that fit, so `_setupCells:` sets a frame and a close-button tracking rect for all of them, and `scrollWheel:` triggers a relayout per event. At high tab counts that is more work per update than the overflow path does. I have not profiled that difference.

If you would rather have it only lay out the cells in the visible range plus a small buffer, I am happy to rework it that way before this is merged.

### DIFF
--- a/ThirdParty/PSMTabBarControl/source/PSMTabBarControl.m
+++ b/ThirdParty/PSMTabBarControl/source/PSMTabBarControl.m
@@ -10,6 +10,7 @@
 
 #import "DebugLogging.h"
 #import "iTerm2SharedARC-Swift.h"
+#import "iTermAdvancedSettingsModel.h"
 #import "NSBezierPath+iTerm.h"
 #import "PSMTabBarCell.h"
 #import "PSMOverflowPopUpButton.h"
@@ -158,6 +159,11 @@ PSMTabBarControlOptionKey PSMTabBarControlOptionPUAFontProvider = @"PSMTabBarCon
 
     // vertical tab resizing
     BOOL _resizing;
+
+    // How far the vertical (left-side) tab bar is scrolled, in points. 0 means the first tab is at
+    // the top. Only meaningful when the scrollable vertical tab bar is enabled; otherwise tabs that
+    // don't fit go into the overflow menu and this stays 0.
+    CGFloat _verticalScrollOffset;
 
     // animation for hide/show
     int _currentStep;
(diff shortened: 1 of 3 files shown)


gnachman:
Thank you, @Conceptron! Merged. I built a few follow-ups on top of your work: moved the toggle out of Advanced Settings into a real checkbox under Settings > Appearance, made drag-to-reorder and tab activation respect the scroll offset, kept tab content and the activity/progress indicators from drawing under the window decorations (while preserving window transparency), and re-synced the hover state after scrolling. Nice feature - I run with a lot of tabs too. I've added you to the Credits page. Thanks for contributing!

gnachman:
Follow-up: I've since extended this to horizontal (top/bottom) tab bars as well, in 3ec57866c. Thanks again for kicking it off!

PR #723, opened 2026-08-06
Honor OSC 9;4;4 (paused) when the percentage is omitted

semx:
`ESC ]9;4;4 ST` — the ConEmu protocol's "paused" state without a percentage — currently does nothing.

The `pr` value is optional for `st=4` per the [ConEmu spec](https://conemu.github.io/en/AnsiEscapeCodes.html), and the comment on that branch in `-terminalPostUserNotification:` already said so, but the code only acted when `pr` was in 0…100. `st=2` (error) handles the same situation by falling back to the `VT100ScreenProgressError` sentinel. The `case 4` comment was also a copy of `case 2`'s and wrongly said "set error state".

### What this does

When `pr` is absent, keep the percentage that is already showing and just switch it to the warning encoding, so a paused 60% download stays at 60% and only changes color. That is what [Ghostty](https://ghostty.org/docs/vt/osc/conemu) does ("v when specified, otherwise unchanged"), and what Windows Terminal does in `Terminal::SetTaskbarProgress`. When there is no percentage to keep, it falls back to 10% — the progress bar's width is derived from the percentage, so 0 would draw an invisible indicator; Windows Terminal substitutes the same value under the name `TaskbarMinProgress`.

An explicitly out-of-range `pr` is still ignored, exactly as before and as `st=1` does — this distinguishes "the parameter is absent" from "the parameter is present but invalid", so `ESC ]9;4;4;101 ST` remains a no-op rather than changing state.

`VT100ScreenProgressPercentage()` is a new `NS_INLINE` helper in `VT100ScreenProgress.h` that extracts the percentage from whichever base is in use; `VT100ScreenProgressIsVisible` now uses it instead of repeating the three range checks. No enum values were added, so `PSMProgress` needs no corresponding change.

### Testing

Six tests added to `ModernTests/VT100ScreenTests.swift` (no new file, so no project change), driving the real parser through the existing `screen(...)`/`feed(...)` helpers:

- 60% then `9;4;4` → warning encoding at 60%; likewise from an error percentage
- `9;4;4` with nothing showing, and from an explicit 0% → the minimum visible value
- `9;4;4;101` and `9;4;4;-1` → state untouched
- explicit 0, 30, 100 → honored exactly
- st=0/1/2/3 unchanged, including the ignored `9;4;1;101`

`tests/progress-bar-paused.sh` is a manual demonstration in the style of the existing `progress-bar-*.sh` scripts.

Manual check:

```sh
printf '\033]9;4;1;60\a'   # bar at 60%
printf '\033]9;4;4\a'      # stays at 60%, turns orange
printf '\033]9;4;4;101\a'  # ignored
printf '\033]9;4;0\a'      # clear
```

### Verification

`make Development` succeeds on this branch — `** BUILD SUCCEEDED **`, no errors and no new warnings in the touched files. The seven tests above run and pass:

```
xcodebuild test -scheme ModernTests -only-testing:ModernTests/VT100ScreenTests
** TEST SUCCEEDED **
```

### DIFF
--- a/ModernTests/VT100ScreenTests.swift
+++ b/ModernTests/VT100ScreenTests.swift
@@ -1616,6 +1616,116 @@ class VT100ScreenTests: XCTestCase {
                        ".....!")
     }
 
+    // MARK: - ConEmu progress protocol (OSC 9;4)
+
+    private func screenProgress(_ screen: VT100Screen) -> Int {
+        var result = 0
+        screen.performBlock(joinedThreads: { _, mutableState, _ in
+            result = mutableState.progress.rawValue
+        })
+        return result
+    }
+
+    // Mirrors kMinimumVisibleProgressPercentage in
+    // VT100ScreenMutableState+TerminalDelegate.m.
+    private let minimumVisibleProgressPercentage = 10
+
+    // The percentage is optional for the paused state, so pausing without one keeps
+    // the percentage that is already showing and only changes its color.
+    func testProgressPauseWithoutPercentageKeepsCurrentPercentage() {
+        let screen = screen(width: 80, height: 24)
+        feed(screen, "\u{1b}]9;4;1;60\u{07}")
+        XCTAssertEqual(screenProgress(screen), VT100ScreenProgress.successBase.rawValue + 60)
+
+        feed(screen, "\u{1b}]9;4;4\u{07}")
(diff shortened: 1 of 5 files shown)


gnachman:
Thanks for the careful fix and the thorough tests, @semx! Nicely done, especially matching Ghostty and Windows Terminal behavior and documenting the absent-vs-invalid distinction.

PR #724, opened 2026-08-06
Keep a subparameter whose parameter was stored

semx:
Follow-up to #722 and `fd151a4`. Your blank-16th fix made me go back and check the whole family properly instead of the cases I happened to think of, and there is one more: a subparameter written after a full parameter list is discarded even though the parameter it belongs to was stored.

```
CSI 1;2;…;16;:5 m
  before #722:  subparameter 5 attached to parameter 16
  after #722:   discarded
  after fd151a4: still discarded
```

A parameter substring that starts with a colon takes no slot of its own — the implementation treats it as belonging to the parameter before it, which is the documented deviation in that comment block. So here the target is the 16th parameter, which was stored perfectly well. Nothing overflowed; the list merely being full is not a reason to drop it. Note it is honored after fifteen parameters and dropped after sixteen, which is the inconsistency.

### The rule underneath

Both earlier patches, mine included, were approximating the same invariant with a flag derived from the parameter count: **a subparameter belongs to the most recently opened parameter, and survives exactly when that parameter was stored.** Tracking that directly is simpler than tracking whether the list is full, and it decides all four cases without special-casing any of them:

| after 16 parameters | belongs to | outcome |
|---|---|---|
| `;;:5` (blank 16th, i.e. 15 parameters + blank) | the blank, stored | keep |
| `;:5` | the stored 16th | keep |
| `;17:5` | the 17th, discarded | discard |
| `;17;:5` | the 17th, discarded | discard |

So `parameterOverflowed` becomes `lastParameterStored`, set wherever a parameter is opened — including the blank one at a semicolon — and read where subparameters are attached. `openedStorableBlankParameter` goes away with it.

### How I checked it this time

I wrote a model of the documented behavior separately from the parser (splitting on `;`, colon-leading substrings belonging to the previous parameter, at most 16 parameters and 16 subparameters), generated 512 parameter lists around the limit by enumeration rather than by hand — uniform lists of every length from 13 to 19, a run of ordinary parameters followed by one interesting substring at every position, pairs of interesting substrings adjacent, and degenerate forms of bare separators — and compared the model against `VT100CSIParser.m` built at four revisions:

```
before #722   disagrees with the model on 116 inputs
after #722    20        ← my patch fixed the overflow family, broke this one
after fd151a4 18        ← your fix recovered the blank-16th case
this branch    0
```

The parse changes on exactly 18 inputs relative to master, and all 18 are that colon-after-a-full-list family; I checked that programmatically rather than by eye. Two regression tests are added to `VT100CSIParserTest.m`, covering the kept case and the discarded-parameter case.

`make Development` succeeds: `** BUILD SUCCEEDED **`, no errors and no new warnings. I could not run the tests — `iTerm2XCTests` doesn't link from the command line on master either (the `IT2Runner` / `IT2ChannelDisconnect` symbols), as mentioned on #722.


iTerm2XCTests/VT100CSIParserTest.m
```diff
@@ -578,4 +578,26 @@ - (void)testGetCSISubparameterByIndex {
     XCTAssert(iTermParserGetCSISubparameter(token.csi, 0, 3) == -1);
 }
 
+- (void)testSubparameterAfterAFullParameterListBelongsToTheLastParameter {
+    // A parameter substring that starts with a colon doesn't open a parameter of its own, so this
+    // subparameter belongs to the 16th parameter, which was stored. Nothing overflowed, so it must
+    // be kept — the list being full is not by itself a reason to discard it.
```
gnachman, line 584: That's unexpected! Is this in ECMA 43? Do other terminal emulators implement this?

semx: No on both counts — but it predates this PR, and the test only pins down what the parser already does.

**Not in the standard.** ECMA-48 §5.4.2 says parameter sub-strings are separated by 03/11 (`;`), and that 03/10 (`:`) "may be used as a separator in a parameter sub-string, for example, to separate the fractional part of a decimal number from the integer part" — the sub-parameter convention we all implement for SGR comes from ISO-8613-6, which xterm's ctlseqs documents as the source of `38:2:…` and `38:5:…` and notes ECMA-48 5th edition marks those codes "reserved for future standardization". The same clause says "an empty parameter sub-string represents a default value which depends on the control function". So by the standard, `;:5` is its own parameter sub-string whose first part is empty, not a continuation of the parameter before it.

**Where it comes from here.** It isn't a decision about leading colons, it's what falls out of the flat `p[]` + `count` structure: sub-parameters are attached to `param->count - 1`, so whatever came last receives them. `b25a40fb0` ("Fix crash on CSI :", Nov 2014) added the `param->count > 0` guard so that a colon with nothing before it stops crashing, and the comment there is the note about ignoring a parameter string that starts with a colon. Building `VT100CSIParser.m` from before #722 confirms nothing about it is new:

```
CSI 1;:5 m          →  p=[1]      subs: p0.s0=5
CSI 1;2;…;16;:5 m   →  p=[1…16]   subs: p15.s0=5
```

What #722 changed by accident was consistency: the value still attached with fifteen parameters but was dropped with sixteen. This PR removes that split.

**Other terminals — measured, not read.** I first inferred both of these from their source and got both wrong, so these are runs, not readings:

- **kitty 0.48.2**, driven through its own parser (`test_create_write_buffer` / `test_parse_written_data`, the entry points its test suite uses): `CSI 4m` → decoration 1, `CSI 4:5m` → decoration 5, and **`CSI 4;:5m` → decoration 5**. Same result as iTerm2 — the `:5` reaches the `4`, and `4;:5m` is indistinguishable from `4:5m`.
- **Ghostty 1.3.1**, asked with DECRQSS (`ESC P $ q m ESC \`) after each sequence: `CSI 4:5m` and `CSI 4;:5m` both report `0;4m`, so the underline survives — it does *not* treat the empty part as a default parameter that resets. Its DECRQSS reply doesn't carry the underline style, so this measurement can't tell me whether the 5 attached or was discarded; I'd rather say that than guess.

So iTerm2 isn't alone here: at least kitty produces the same result, and Ghostty at minimum doesn't do the spec-literal thing either.

**If you'd rather change the semantics, I'm happy to.** The coherent alternatives are to make a colon-leading sub-string its own default-valued parameter and hang the sub-parameters on that, or to reject the sequence. Either is a deliberate behavior change rather than the fix this PR is, so it belongs in its own commit — and then this test should be rewritten to encode the new intent instead of kept. Tell me which you'd prefer; if the semantics stay as they are, the PR stands.

semx:
No rush on this — just flagging that it's waiting on you, since the question above is answered.

Short version of that answer: ECMA-48 doesn't define a parameter substring that starts with a colon (it documents `:` as a separator *within* a substring, for decimal fractions; the sub-parameter convention we implement comes from ISO-8613-6). The behavior isn't introduced here either — building `VT100CSIParser.m` from before #722 shows `CSI 1;:5 m` already attached the subparameter to the previous parameter, and `b25a40fb0` is where the `count > 0` guard and that comment came from. And on other terminals I measured rather than read: kitty 0.48.2, driven through its own parser, renders `CSI 4;:5 m` exactly like `CSI 4:5 m` — same as iTerm2.

So the PR only removes the split where the same subparameter was honored after fifteen parameters and dropped after sixteen. If you'd rather change the semantics instead — make a colon-leading substring its own default-valued parameter, or reject the sequence — say which and I'll rewrite the fix and the test around it. If you'd rather leave it alone, this stands as is, and I'm happy either way.

PR #725, opened 2026-08-08
Add optional inactivity timeout for tab colors

Chipagosfinest:
## Summary

- add an opt-in advanced setting that hides custom tab colors after a configurable number of hours without terminal output
- preserve the configured color while disabling only its displayed appearance, so fresh output restores it
- keep split-pane inheritance and Minimal-theme pane tinting correct when some colors have expired

The default is 0 hours, which preserves the existing behavior indefinitely.

## Verification

- PASS: deterministic Swift policy smoke covering disabled, before-threshold, exact-threshold, and fresh-activity behavior
- PASS: Swift typecheck and generated Objective-C header selector verification
- PASS: Xcode project structure and target-membership parse
- PASS: git diff --check
- BLOCKED: the focused ModernTests target cannot launch locally because this Mac has Command Line Tools but no full Xcode installation; xcodebuild exits before compilation

Focused tests are included in ModernTests/iTermTabColorInactivityPolicyTests.swift for CI or a machine with Xcode.


gnachman:
Thanks for putting this together, and for the thorough tests and verification notes. The behavior you implemented is exactly right: expire on inactivity, restore on output, and never lose the configured color.

That said, I'd like to steer this out of the app rather than merge it as native code plus an advanced setting. My reasoning:

  - Advanced settings have a high bar, and this is a fairly niche preference. The default is 0 (off), so for everyone who doesn't use it, it's permanent surface area: new per-session state, a displayedTabColor concept that every current and future tab-color call site has to reason about, and a check wired into the per-session display update path.
  - The whole feature is achievable from the Python API without touching the app. Hiding and restoring a tab color is just a session-local profile change (toggling "Use Tab Color" while leaving the color value intact), and resumed output can be detected from screen-update notifications.

So I wrote a small script that does the same thing, and attached it here as a .its installer (in the zip). Double-click the .its to install it, then enable the Python API if you haven't  (Settings > General > Magic > Enable Python API). The timeout is a variable at the top of the script (TIMEOUT_HOURS), matching the hours unit you chose.

How it works, and where it honestly differs from your native version, so the tradeoff is clear:

  - It watches screen-update notifications only for sessions that currently have a custom tab color, so idle plain tabs generate no traffic. There is no "tab color changed" event in the API  (session-local profile edits are written to a model whose change notifications are disabled), so to discover which sessions are colored it does an infrequent, staggered scan of a few profile keys per session. Your native code sees the color change synchronously and reads true last-output time for free; a script structurally cannot, so it approximates both.
  - "Output" is inferred from screen changes, which is a close proxy but also fires on selection changes and blinking text (though not a plain blinking cursor).
  - It only acts while running, and if it is restarted while some colors are hidden it has no record that it hid them, so those tabs stay colorless until their color is re-enabled or the session ends.

Those limitations are the price of doing it in userland, and for a personal preference like this I think that's the right place for it. If you find the script works well for you, I'd like to close this PR in favor of it. If there's a reason the native fidelity genuinely matters for your use case, I'm happy to hear it.

Either way, I appreciate the contribution and the care you put into it.


[tab_color_inactivity_timeout.its.zip](https://github.com/user-attachments/files/31398204/tab_color_inactivity_timeout.its.zip)

PR #726, opened 2026-08-13
Add composer @-routing to send commands to other tabs/panes/windows

jianran:
## Summary
- The composer can send commands to other tabs, panes, and windows using @-prefixed addresses: `@2 ls` types into tab 2's active session, `@2.3 ls` hits pane 3 of tab 2 (panes counted top-left first), `@w3.2 ls` reaches window 3's tab 2, `@all` goes to every other session in the window, `@wall` to every window
- A toast confirms the routed send and the composer stays open
- After an @-addressed send, plain sends stick to the same target until `@. cmd` clears it (running `cmd` locally) or the composer closes
- A literal leading `@` can be escaped with a backslash
- Controlled by the new `composerTabRouting` advanced setting (enabled by default)

## Test plan
- [x] `ModernTests/ComposerTabRouterTests.swift` — unit tests for @-address parsing and sticky-target behavior
- [x] Manual test plan: `tests/composer-tab-routing-manual-test.md`

### DIFF
--- a/ModernTests/ComposerTabRouterTests.swift
+++ b/ModernTests/ComposerTabRouterTests.swift
@@ -90,6 +90,21 @@ class ComposerTabRouterTests: XCTestCase {
         XCTAssertEqual(route.paneNumber, 0)
     }
 
+    func testHereClearsDefaultAndRunsLocally() {
+        let route = ComposerTabRouter.parse("@. ls -l")
+        XCTAssertEqual(route.kind, .here)
+        XCTAssertEqual(route.payload, "ls -l")
+    }
+
+    func testBareHereFallsThrough() {
+        XCTAssertEqual(ComposerTabRouter.parse("@.").kind, .none)
+        XCTAssertEqual(ComposerTabRouter.parse("@.   ").kind, .none)
+    }
+
+    func testDoubleDotFallsThrough() {
+        XCTAssertEqual(ComposerTabRouter.parse("@.. x").kind, .none)
+    }
+
     func testEscapeSendsLiterally() {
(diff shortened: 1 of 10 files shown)


gnachman:
Thank you for this, Jianran. The core idea (address other sessions by name and fan a command out to them) is one we wanted, and it pushed us to build it out. After looking at it we ended up giving the feature a different home, so I want to explain the approach and where it landed.

The concern with hanging it off the composer is that the composer belongs to one session, and if you have a grid of agents each one is small, so a cross-session control surface there is cramped and easy to miss. We already had the Cockpit window (a floating panel that lists every window, tab, and session), which was under-used, so we made it the home for this instead.

What shipped:

- The Cockpit now docks the real multi-cursor composer along its bottom, in your profile's font, and focuses it when you open the panel, so directing a message is keyboard-only from the first keystroke.
- Type `@` to pick a session, tab, or window from the same picker the AI chat uses, with a live preview. Each pick becomes an atomic chip. A tab or window chip fans out to all of its panes. Only chips are targets, so any literal text you type (including a bare `@foo`) is sent verbatim.
- The chips and the list selection are two views of the same thing: selecting rows adds chips, and editing chips reselects rows.
- The list and composer are resizable split panes, and the list shows each session's live status (its own status word and color, plus a detail line) with a dynamic filter bar.

So the routing concept from this PR is in the product, just via the Cockpit rather than the per-session composer. It is on master now and should be in tomorrow's nightly build. I would love for you to try it and tell us how it feels compared to what you had in mind here.

I am closing this PR since the functionality landed in that form, but thank you again for the nudge that got us to build it.

PR #730, opened 2026-08-17
Fix framework builds on case-sensitive filesystems

cklukas:
Every framework sub-build passes `CONFIGURATION_BUILD_DIR=$(SRCROOT)/Build/$(CONFIGURATION)` to
xcodebuild, so the products land in `Build/Release` (or `Build/Debug`) — but the `mv` that collects
each product reads from lowercase `build/Release`. On the default case-insensitive APFS those are
the same directory, so this has always worked; on a case-sensitive volume the `mv` fails with
"No such file or directory" immediately after a **successful** build, which makes the failure
confusing to diagnose.

Affected recipes, all with the same one-word fix (`build/` → `Build/`, matching what xcodebuild
was told to do):

| File | Recipes |
|---|---|
| `Makefile` | `SwiftyMarkdown`, `Highlightr` |
| `BetterFontPicker/Makefile` | `all`, `dev` |
| `ColorPicker/Makefile` | `all` |
| `SearchableComboListView/Makefile` | `all` |

The change is byte-for-byte a no-op on case-insensitive filesystems.

**How found / verified:** building the repo on Case-sensitive APFS (an external volume).
`make paranoid-deps` failed at SwiftyMarkdown, then Highlightr, then BetterFontPicker, each time
after `** BUILD SUCCEEDED **`; with these fixes, a clean `make paranoid-deps` followed by
`make Development` completes on the same volume, and the resulting app runs.

Related but filed separately: the CoreParse submodule has a case-mismatched `#import` that breaks
the *app* build on case-sensitive filesystems in a subtler way (duplicate `CPToken` interface).
That fix is a PR against `gnachman/CoreParse`; this PR is independent of it.

### DIFF
--- a/BetterFontPicker/Makefile
+++ b/BetterFontPicker/Makefile
@@ -13,8 +13,8 @@ endif
 all:
 	xcodebuild -parallelizeTargets -target BetterFontPicker -configuration Release 'CONFIGURATION_BUILD_DIR=$$(SRCROOT)/Build/$$(CONFIGURATION)' $(SIGNING_FLAGS) $(ARCH_FLAGS)
 	rm -rf BetterFontPicker.framework
-	mv build/Release/BetterFontPicker.framework .
+	mv Build/Release/BetterFontPicker.framework .
 dev:
 	xcodebuild -parallelizeTargets -target BetterFontPicker -configuration Debug 'CONFIGURATION_BUILD_DIR=$$(SRCROOT)/Build/$$(CONFIGURATION)' $(SIGNING_FLAGS) $(ARCH_FLAGS)
 	rm -rf BetterFontPicker.framework
-	mv build/Debug/BetterFontPicker.framework .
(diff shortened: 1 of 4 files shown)


cklukas:
Related CoreParse pull request: https://github.com/gnachman/CoreParse/pull/1

gnachman:
Thank you for tracking this down, @cklukas. Diagnosing a build that fails with "No such file or directory" immediately after `** BUILD SUCCEEDED **` on a case-sensitive volume is exactly the kind of subtle, confusing failure that's easy to lose hours to, and your fix is clean and correct: the `mv` now reads from the same `Build/$(CONFIGURATION)` directory that each `xcodebuild` invocation is already told to write to, matching the existing Sparkle recipe. No-op on case-insensitive filesystems, and a real fix on case-sensitive ones. Merging now.

PR #731, opened 2026-08-17
✨ feat(compact-tabs): show the window name beside the tabs in Compact and Minimal

kud:
> [!NOTE]
> Rebased onto current `master` and manually verified since it was first pushed. One bug surfaced during that pass and is fixed here: the space reserved for the tabs was computed from `minCompactTabWidth` (60pt) while the tab bar actually enforces `minTabWidth` (75pt) by default, and the inter-cell spacing was not counted at all — so a long name could take room the tabs needed and push one into the overflow menu. Both sites now read a single shared `-tabBarCellMinWidth`, and the spacing term is included.

### 📄 Description

Under the Compact and Minimal themes a window can be named, and then the name is never shown.

`iTermThemedWindowType()` maps `WINDOW_TYPE_NORMAL` to `WINDOW_TYPE_COMPACT` for both of those tab styles (`ITAddressBookMgr.m:119`), so there is no title bar to draw into. The only fallback is the fake label `_windowTitleLabel`, and its visibility is exactly `-rootTerminalViewShouldDrawWindowTitleInPlaceOfTabBar` (`PseudoTerminal.m:1173`), one of whose conditions is that the tab bar is **not** visible. The title is drawn _in place of_ the tab bar, never alongside it — so the moment a second tab exists, "Set Window Title" has no visible effect at all.

That predicate answers a layout question — _should the title take the tab bar's slot?_ — but it is the only thing gating an identity question — _should the name be visible?_ Those coincide in every theme that has a title bar, which is why this only shows up here.

This adds a second, much smaller home for the name: the left strip of the tab bar, after the stoplight buttons, proxy icon and window number.

That position is not a new mechanism. `-tabBarInsetsForCompactWindow` (`PseudoTerminal.m:5414`) already reserves left-strip space for exactly this class of decoration, and `insets.left` is turned into `leftMarginForTabBarControl` by every tab style. The name becomes a fourth item in an established row. `PSMTabBarControl` is untouched.

Grouping reads left to right as _window identity, then its contents_: stoplights act on the window, the proxy icon is its document, the number names it, then the tabs it holds.

CleanShot 2026-08-18 at 22 05 07@2x

CleanShot 2026-08-18 at 22 05 20@2x

> [!IMPORTANT]
> **This changes a default, deliberately — flagging it rather than burying it.**
>
> The new advanced setting `showWindowNameBesideTabs` defaults to **“When a custom name is set”**, not to off. A window that has never been named renders identically to before, so nobody who does not use Set Window Title sees any change. But a user who _has_ named a window will now see that name where previously they saw nothing.
>
> That is the bug this fixes, so shipping it off by default would mean nobody finds it. Happy to flip the default to `Never` if you would rather it were strictly opt-in.

Showing only the _custom_ name is also the behaviour I would argue for on its own merits: the derived title is already on screen, because it **is** the tab label. Repeating it beside the tabs is the redundancy Minimal exists to remove. A name the user typed exists nowhere else in the interface, so it is the only part of the title that carries new information. `Always` is available for anyone who disagrees.

The two decisions support each other. Because the name only appears once typed, it is a near-static string, which is what makes a space-reserving slot safe. Under `Always` the derived title tracks the running job, so the tab origin shifts as it updates — inherent to reserving space for a live string, and the reason it is not the default.

### 🔍 Scope

The name yields to the tabs rather than competing with them. A tab is somewhere to go; the name is only context. So it truncates with a tail ellipsis before any tab is pushed into the overflow menu, is capped at the lesser of 180pt and a quarter of the strip, and hides entirely below 44pt rather than rendering an uninformative `My…`.

Deliberately out of scope, flagged rather than handled:

- **Vertical and bottom tab bars.** Gated to `PSMTab_TopTab` only.
- **Full screen**, and the tab bar being on loan to the titlebar accessory. Both early-return; a native title bar exists there.
- **Non-compact window types.** They have a title bar already.
- **Per-window colour.** Windows have no colour concept today, and tab groups have just introduced the only hue in that strip. A second colour language at a different scope seemed likely to confuse rather than clarify, so this is name-only.
- **Promoting the setting to Preferences → Appearance.** It is arguably the right long-term home next to the window-number and proxy-icon toggles, but that means a `PreferencePanel.xib` diff, which is hard to review. Advanced setting for now; happy to follow up if you want it surfaced.

One small tidy inside the change: the vertical cap-height centring maths used by the window number label is now `-tabBarStripLabelOriginYForFont:`, shared by both labels, so the two sit on a common baseline by construction rather than by two copies of the same arithmetic.

### ✅ How to Validate

1. Settings → Appearance → Theme: **Minimal**. Open a window and a second tab so the tab bar is showing.
2. Window → Edit Window Title → enter a name. It should appear in the left strip, after the window number, sharing its baseline, with the tabs shifted right to make room.
3. Clear the title — the name disappears and the tabs reclaim the space.
4. Give it a long name, open six tabs, and drag the window narrower. The name should truncate, then disappear, **without** any tab being pushed into the overflow chevron. _This is the step I am least sure of and most want a second pair of eyes on._
5. Settings → Advanced → search `window name beside`. `Never` hides it immediately with no restart; `Always` shows the derived title even with no custom name.
6. Single tab, tab bar hidden → the existing centred fake title still behaves exactly as before.
7. Theme → **Compact** behaves the same; **Dark**/**Light** show no strip label, since the real title bar already has the name.
8. Tab bar position Left / Right / Bottom → no label, no layout change.
9. Double-click the tab bar still zooms; right-click still gives the tab bar menu; dragging the window by the strip — including by the name itself — still moves the window.

### 🛠️ Developer Checklist

- [x] Single atomic commit
- [x] Builds clean with no new warnings (`tools/build.sh Development`)
- [x] Release notes updated in `docs/notes-3.7.txt`
- [x] Manually verified end to end
- [ ] Tests included and passing (if applicable) — there is no test coverage for tab bar layout in this area


gnachman (commented):
Thanks for this, and for the unusually careful writeup and scoping. I reviewed it and tested on macOS 26 (Tahoe). The overall approach (a fourth item in the existing left-strip decoration row, gated to top tabs, reserving inset space so the tabs are not crowded) is sound and reads well. A few things to address before it can go in.

## Blocking

**1. The name does not come back after exiting full screen.**
Set a window name, enter full screen, then exit: the label stays gone. `-rootTerminalViewWindowNameBesideTabs` returns nil during the transition (`togglingLionFullScreen_ || anyFullScreen`), and `-windowNameBesideTabsTextWidth` also returns 0 while the tab bar is on loan to the titlebar. The label caches that empty state in its `stringValue` and `hidden` flag, and nothing re-runs `-updateWindowNameBesideTabsText` once the transition settles: the window title does not change, so `-windowTitleDidChangeTo:` never fires, and the post-exit `-layoutSubviews` calls happen while the tab bar is still being handed back. The window number label survives the same transition only because its visibility is recomputed straight from the delegate on every `-layoutWindowPaneDecorations`, with no cached string gating it. Suggest forcing a name refresh from wherever the tab bar is returned from loan, so it covers full-screen exit and any other borrow/return, rather than a full-screen-specific one-off.

**2. `-allowanceForWindowNameBesideTabs` ignores `extraSpaceBeforeCompactTopTabBar`.**
`-tabBarInsetsForCompactWindow` reserves `... + windowNameWidth + extraSpace`, but the allowance only budgets a fixed 24pt chevron cushion and never subtracts `extraSpace`. With that advanced setting above ~24 and a near-full bar, showing the name can push a tab into the overflow menu, which is exactly the invariant this logic exists to protect.

## Please re-validate

**3. Scrollable tab bars are now on by default on master** (that default flipped after the base you rebased onto). When scrolling is on, tabs scroll instead of overflowing, so the overflow-avoidance path here is bypassed in the default configuration, and the interaction that matters becomes whether the reserved name inset plays well with the scroll viewport rather than the overflow menu. Please re-test with scrollable tab bars on: in particular, scroll so a tab reaches the far-left edge and confirm it is not rendered underneath the name label.

## Smaller items

**4.** `-applyWindowNameBesideTabsAttributes` builds the attributes dictionary with the decoration color; if `-rootTerminalViewTabBarTextColorForWindowNumber` is ever nil the dictionary literal throws, whereas the existing window-number consumer tolerated nil via `.textColor`. Worth a nil guard.

**5.** `-leadingEdgeForWindowNameBesideTabs` omits the `proxyIconExtraPadding` that the no-window-number branch of `-tabBarInsetsForCompactWindow` adds, so the label sits ~4pt off when the window number is hidden and a proxy icon is shown.

**6.** The label sets `lineBreakMode = NSLineBreakByTruncatingTail` in init, but it is always shown via `attributedStringValue`, which carries its own truncation (your comment notes this), so the init line never takes effect and can be dropped.

**7.** The advanced-setting help text and description tell people to name the window with “Window > Set Window Title,” but the actual menu item is “Edit Window Title” (only the dialog is headed “Set Window Title”). Please match the real menu item name so it can be found.

On the default: shipping “When a custom name is set” as the default is fine by me, and thanks for flagging it so clearly.

Separately, if you happened to see the overflow menu drop its first entry while testing on Tahoe, that was a pre-existing bug unrelated to this PR and is already fixed on master.

kud:
Thank you, this is a genuinely useful review, and finding the full-screen case took real effort. All seven are addressed. Rebased onto current `master` first, which also picks up your overflow-menu fix; I did see that empty menu on Tahoe and had put it down to my own change, so it is a relief to learn otherwise.

**1. The name not coming back after full screen.** Fixed, and your diagnosis was the right one. The hook now lives where the tab bar is returned from loan rather than anywhere full-screen-specific.

One wrinkle worth naming, because it changes the shape of the fix: the return can happen while the window is still full screen. `-windowWillExitFullScreen` calls `-hideStandardWindowButtonsAndTitlebarAccessories`, which returns the bar before `anyFullScreen` goes false, so re-reading the delegate at that instant reads the same nil. The refresh is therefore deferred to the next runloop turn, the way `-updateTabBarControlIsTitlebarAccessory` defers its `-repositionWidgets` for issue 12811.

I also widened `-updateWindowNameBesideTabs`, which was gated purely on the string changing. That was the second half of the latch you spotted: after a borrow/return the string can be identical while the label is still hidden from when the width was forced to 0. It now also compares whether the name can be shown at all, so a stale `hidden` flag is enough to trigger a relayout on its own.

**2. `extraSpaceBeforeCompactTopTabBar`.** Fixed, the allowance subtracts it now. You are right that it is the one thing that logic exists to prevent, and the inset was reserving space the allowance was handing to the tabs on paper.

**3. Scrollable tab bars.** Re-tested with them on. The short version is that the overflow-avoidance path is bypassed, as you say, but the reserved inset lands somewhere safe: `-scrollLeadingMargin` is `leftMarginForTabBarControl`, which is `insets.left`, which is the same number the name grows. So the cell clip in `-drawRect:` already excludes the name's band, and a tab scrolled to the far left stops at the name rather than passing under it. I checked the drag auto-scroll path too, since that shifts cells without touching `_scrollOffset`.

**4. Nil decoration colour.** Guarded. The attributes are built mutably now and the colour is only set when it is non-nil, which gives the same tolerance `-setTextColor:` has.

**5. `proxyIconExtraPadding`.** Fixed. Rather than copy the `4` into the second site, I pulled it out as `iTermRootTerminalViewCompactProxyIconExtraPadding` in `iTermRootTerminalView.h` and both places read it. This is the second time these two calculations have drifted apart, the first being the `minTabWidth` one in the note at the top of the PR.

**6. `lineBreakMode` in init.** Dropped.

**7. “Edit Window Title”.** Fixed in the advanced setting's description and in the release note. You are right that only the dialog says “Set”, and a setting that tells you to look for a menu item that is not there is worse than one that says nothing.

Still a single atomic commit, builds clean with no new warnings.

gnachman:
Thanks for the very thorough writeup, especially calling out the width-reservation step as the one you were least sure of. A few things from a review pass, roughly in priority order:

**1. `Always` mode relayouts the whole root view on every title tick.** `windowTitleDidChangeTo:` now unconditionally calls `updateWindowNameBesideTabs`, and in `Always` mode the name is `presentationWindowTitle`, which updates on every job/dir/title change. Each of those makes `updateWindowNameBesideTabsText` return YES, so `updateWindowNameBesideTabs` runs a full `layoutSubviews` (tab bar, toolbelt, status bar, division view, tab style) just to move one label. Before this change, a compact top-tab window with the fake title hidden skipped that work on title changes entirely. The default `WhenCustom` mode is fine because the custom name is static; it's only `Always` that pays this per-tick.

**2. The overflow-avoidance math re-derives what the tab bar already knows, and can drift from it.** `-allowanceForWindowNameBesideTabs` estimates `tabCount * cellMinWidth + intercellSpacing` plus a hardcoded 24pt overflow allowance, but the real usable width lives in `PSMTabBarControl` / the tab style (`leftMarginForTabBarControl`, `rightMarginForTabBarControlWithOverflow:addTabButton:`, and the per-cell widths from `cellWidthsForHorizontalArrangement`). Two concrete ways they disagree:
- It omits the tab bar's own left margin, so it systematically overstates free space near the boundary.
- `numberOfTabViewItems` overcounts when a tab group is collapsed to a single chip cell, and doesn't account for pinned tabs or uneven/optimum sizing, so `widthTabsNeed` can be well off in either direction.

This is the part you flagged, and I think the discomfort is warranted: it's a hand-mirror of the tab bar's fit logic rather than asking the tab bar what fits. If there's a way to route the reservation through `PSMTabBarControl`'s existing available-width calc, that would keep the two from rotting apart.

**3. `-windowNameBesideTabsTextWidth` is recomputed ~3x per layout pass.** Each call runs the full allowance math (frame width, toolbelt frame, tab count, intercell spacing, several advanced-setting reads, `frameForStandardWindowButtons`, `compactProxyIconWidthIncludingMargin`) plus `NSTextField.fittingSize`. In one `layoutSubviews` it's evaluated in `layoutWindowPaneDecorations` for the hidden check, again via `-frameForWindowNameBesideTabsLabel`, and a third time via `-windowNameBesideTabsWidthIncludingMargin` from `-tabBarInsetsForCompactWindow`, all identical. Since `layoutSubviews` runs on every resize/drag frame, computing it once and threading it through would save the repeated text measurement.

**4. The name loses its intended dimming when the window-number color is nil.** `-applyWindowNameBesideTabsAttributes` only sets `NSForegroundColorAttributeName` (and the 0.55 alpha) inside `if (decorationColor)`. When `rootTerminalViewTabBarTextColorForWindowNumber` returns nil (which the code anticipates), the attributed string has no foreground color, so it draws at `labelColor` / alpha 1.0 and reads as a peer of the tab titles rather than quiet context.

**5. Minor: `-leadingEdgeForWindowNameBesideTabs` duplicates the inset arithmetic in `-tabBarInsetsForCompactWindow`.** The "stoplights + proxy icon + (window number box vs. proxy-icon extra padding)" formula is now in both places (the comment says "Mirrors both branches"). A single method returning the post-window-number leading edge, called from both, would keep the drawn label and the reserved space from diverging.

The refactoring is nice, for the record: `-tabBarStripLabelOriginYForFont:` preserves the window-number Y origin exactly (including the macOS 26 minimal shift), and `-tabBarCellMinWidth` is a faithful hoist. No recursive layout loop either, since `layoutSubviews` only calls the text-only `-updateWindowNameBesideTabsText`.

kud:
Thanks, this was a genuinely useful pass. All five are fixed in `db79225b6`. Three of them turned up something underneath that I think you'll want to know about, so those are separated out below.

**1. `Always` mode relayout.** `-updateWindowNameBesideTabs` now compares the reserved width before and after the text changes, and only calls `-layoutSubviews` when that width actually moves. When it hasn't, it sets the label's frame directly. In `Always` mode most title ticks land on the same reserved width, so they no longer cost a layout pass.

**2. The overflow-avoidance math.** You were right that the discomfort was warranted. There's now `-maximumLeftInsetFittingAllCellsMinimallyForWidth:` on `PSMTabBarControl`, backed by a private `-minimumCellAreaWidth` that mirrors the three branches of `-cellWidthsForHorizontalArrangementWithOverflow:` (chips, pinned, plain). `-allowanceForWindowNameBesideTabs` just asks for that number. The hand-rolled tab count, the missing left margin and the hardcoded 24pt allowance are all gone.

Uneven and optimum sizing deliberately have no branch in the minimum: `-variableCellWidthsWithOverflow:` returns nil unless every cell fits at its desired width and desired is already at least `cellMinWidth`, and `-shouldUseOptimalWidthWithOverflow:` is only true when optimum already fits. Neither can overflow a bar that the minimal expression says fits.

I also dropped the "never more than a quarter of the strip" cap. With the fit clamp doing real work it was a second hiding threshold competing with the 44pt minimum, and the 180pt cap already covers proportion. Happy to put it back if you'd rather keep it.

**3. Triple recomputation.** `-windowNameBesideTabsTextWidth` is memoised for the duration of a layout pass now, invalidated at the top of `-layoutSubviews` and whenever the text changes.

**4. Lost dimming.** The colour falls back to `labelColor` and the alpha applies unconditionally, so the branch is gone entirely.

**5. Duplicated leading-edge arithmetic.** Now `-widthOfDecorationsBeforeWindowNameBesideTabs` in `iTermRootTerminalView`, used by both `-tabBarInsetsForCompactWindow` and the label's frame. `iTermRootTerminalViewCompactProxyIconExtraPadding` went back to file-static as a result, so the header is a line shorter than it was.

---

Three things I ran into that you didn't ask about.

**The premise under 2 is narrower than either of us wrote it.** `-tabBarIsScrollable` is just `kPreferenceKeyScrollableSideTabBar`, which defaults to `@YES` and has no orientation gate despite the name, and `PSMTabBarControl.m:1779` routes horizontal bars into `-layoutScrollableHorizontalTabsWithCellCount:animate:`. So the default configuration scrolls rather than overflowing.

It doesn't change the fix. That method branches on `totalTabWidth > [self availableCellWidthWithOverflow:NO]`, and on the fits side it calls exactly the same width builders as the non-scrollable path, which is the only side where a reservation matters at all. On the scrolls side nothing is hidden and the name's only cost is that you scroll slightly sooner. That's why the new method is named for fitting rather than for overflow. If you'd want the name to behave differently once the bar is actually scrolling, say so and it's a small change.

**`cellMinWidth` is a soft floor on that path.** I wrote a test asserting a cell gets dropped one point past the predicted limit, and it failed: nothing is dropped. `-computeCellFramesInContainerOfWidth:` divides the available width evenly among the cells it has, so they shrink below `cellMinWidth` first and are only dropped much later, and the `while` loop above it charges a single `intercellSpacing` rather than one per gap. The test asserts the squeeze rather than a drop now. Not something I've touched, but it surprised me.

**Item 5 was worse than a duplicated formula.** The two sites weren't computing the same quantity. `-tabBarInsetsForCompactWindow` used `compactTabBarStoplightButtonsWidth` (75 by default), while `-leadingEdgeForWindowNameBesideTabs` used `NSMaxX([self frameForStandardWindowButtons])`, which is 69, or 95 with the stoplight hotbox enabled. The label was being drawn outside the gap reserved for it. I kept the advanced setting as canonical, because changing the inset to match the button frame instead would shift the tab bar in every compact top-tab window whether or not this feature is switched on.

---

Two changes you didn't ask for, flagged because they're in code you own rather than mine:

- I moved the tab-style property pushes (`setCellMinWidth:` and the rest) above the layout dispatch in `-layoutSubviews`. They were being pushed after the insets were computed, so anything asking the tab bar what fits saw the previous pass's settings and lagged a pass behind on every preference change. Each setter assigns its ivar synchronously and only schedules the relayout, so this changes what the reservation reads, not when the bar lays out.
- `PSMTabBarControl.h` gained `- (NSMutableArray *)cells`, exposing an accessor that already existed in the `.m`, so the new test can populate a window-free bar without going through the KVO binding that `-addTabViewItem:` sets up.

Unrelated to this PR, but worth knowing: `-initWithFrame:` assigns `_style` directly, while `-setStyle:` is the only thing that sets `_style.tabBar = self`. A freshly constructed bar therefore has a style whose `leftMarginForTabBarControl` reads `tabBar?.insets.left ?? 0` and can't see its own insets. It doesn't bite in the app because the theme assigns a style later, but it made my first version of the test pass vacuously against a bar that thought it had the whole width.

The new tests in `ModernTests/PSMTabBarMinimumCellAreaTests.swift` pin `-minimumCellAreaWidth` against the real width builder across all three branches, since those two mirroring each other is the standing cost of this approach.

Everything compiles clean and the unit tests pass. I haven't eyeballed the label position across every theme and hotbox combination yet, so if anything looks off positionally, Minimal with the stoplight hotbox enabled is where I'd look first.

kud:
@gnachman I would like to have this feature soon. For now, I have created a "fake" tab in the first position just to differentiate between my two iTerm windows. 🫠

gnachman:
## Review

Nice work, and thank you for the unusually thorough writeup, it made this much easier to review. The refactor is careful: the `tabBarInsetsForCompactWindow` extraction and the `frameForWindowNumberLabel` split are arithmetically identical to what they replace, the tab-style setters moved earlier in `layoutSubviews` are side-effect-free so the reorder is safe, and the theme-switch / tab-position / window-number / fullscreen transitions all relayout correctly through existing paths. A few things came up, two of which I would want addressed before merge.

### Blocking

**1. The new test file is not in the Xcode project, so it never compiles or runs.**
`grep PSMTabBarMinimumCellAreaTests iTerm2.xcodeproj/project.pbxproj` returns nothing, while sibling ModernTests files are present. This is the load-bearing miss: the whole point of the file is to pin `minimumCellAreaWidth` against `cellWidthsForHorizontalArrangementWithOverflow:` so the two “expressions that must agree” cannot silently drift, and that protection is worth zero if CI never executes it. Please add it to the **ModernTests** target and `git add` it. Note the header comment on line 9 also labels it `iTerm2XCTests`, which is a legacy, bit-rotted target that is not run, so please fix that too rather than adding it there.

**2. The 44pt minimum-width gate is applied to the available slack, not to the name’s rendered width, so a short name that would fully fit is hidden.**
In `measuredWindowNameBesideTabsTextWidth`:
```objc
const CGFloat allowance = MIN(iTermWindowNameBesideTabsMaximumWidth, [self allowanceForWindowNameBesideTabs]);
if (allowance < iTermWindowNameBesideTabsMinimumWidth) { return 0; }
return MIN(allowance, ceil(_windowNameBesideTabsLabel.fittingSize.width));
```
The gate exists to avoid a tail-ellipsized “My…”, but it fires even when no truncation would occur. A window named “A” (~12pt) with a nearly-full tab bar leaving ~30pt of slack is hidden entirely (`30 < 44`), even though 12pt fits within the 30pt slack without stealing any tab space. That contradicts both stated goals: the name is neither truncated nor crowding the tabs. The guard should key off the rendered width, e.g. only hide when `ceil(fitting) > allowance && allowance < 44` (truncation actually occurs).

### Worth fixing

**3. `allowanceForWindowNameBesideTabs` reads a stale `_toolbelt.frame` while the tab bar is sized from the fresh `floor(self.toolbeltWidth)`, over-reserving during a live toolbelt drag.**
```objc
const CGFloat stripWidth = NSWidth(self.frame) - ([self shouldShowToolbelt] ? NSWidth(_toolbelt.frame) : 0);
```
This runs from `layoutWindowPaneDecorations`, but `_toolbelt.frame` is not updated until `updateToolbeltFrameForWindow` later in the same `layoutSubviews` pass; the drag path (`growToolbeltBy:` → `repositionWidgets`) updates the `_toolbeltWidth` ivar first and relayouts against the old frame. Widening the toolbelt makes `stripWidth` too large by the drag delta (tens of points), so the name over-reserves; with enough tabs to sit near the overflow boundary a cell gets squeezed under `cellMinWidth` / into overflow for that frame, correcting when the drag stops (visible as a flicker). Computing `stripWidth` from `floor(self.toolbeltWidth)`, which is what the tab bar itself uses, would remove the lag.

**4. `minimumCellAreaWidth` is a parallel reimplementation of the three-branch width builder that must be hand-kept in sync.**
The chip/pinned/plain branches mirror `cellWidthsForHorizontalArrangementWithChipsWithOverflow:` and the `canFitAllCellsMinimally` predicate rather than sharing code, and the comment admits “two expressions that must agree.” Defensible *if* the pinning tests run, but per #1 they currently do not, so today the drift is guarded by nothing. Fixing #1 makes this an acceptable tradeoff.

### Polish / decisions

**5. RTL/bidi window names are not considered.** `NSKernAttributeName` (0.25 tracking) is applied unconditionally, which disrupts cursive joining in Arabic/Hebrew, and the fixed `NSLineBreakByTruncatingTail` on a hard left-anchored frame clips the visually-leading end of an RTL name. Cosmetic, but worth a deliberate decision rather than an accident, since it is the one dimension the change does not touch.

**6. Text measurement forces a live NSTextField layout.** `measuredWindowNameBesideTabsTextWidth` uses `_windowNameBesideTabsLabel.fittingSize.width`, and `frameForWindowNameBesideTabsLabel` then calls `sizeToFit` for height, two layouts of the same control. The codebase already has pure-string measurement (`-[NSString it_boundingRectWithSize:attributes:truncated:]`) and a kern builder (`+[NSMutableAttributedString kernForString:toHaveWidth:withFont:]`). Cached per pass so minor, but avoidable on the `layoutSubviews` hot path, especially in Always mode.

**7. In Always mode, any name-width change falls out of the fast path into a full `layoutSubviews`.** The `widthBefore == widthAfter` fast path only holds until the name saturates at the 180pt cap; a command that rewrites its title with varying-length text drives a full window relayout per update. Not incorrect, and the default is WhenCustom, but worth confirming acceptable before treating Always as first-class.

### Checked and cleared

The `hitTest:` redirect does **not** kill the name’s tooltip (tooltips register a geometry rect on the window independent of hitTest); a live advanced-settings change does not leave the tab bar inset stale (the first guard in `updateWindowNameBesideTabs` returns before the fast path when the text is unchanged); and the delegate/scope/name symbols are all correctly wired. The `-cells` accessor is not actually new surface.

kud:
Thanks, this was a genuinely useful pass. Four of the seven were real and are fixed in `f7461f46c`. One I want to push back on.

## 1: the test file is already in the target

I think this one is a false alarm, and the evidence is a little counterintuitive.

`ModernTests` is a **synchronised root group**, not a listed one. From `project.pbxproj` line 10042:

```
A6A723AF2DC04C2200A8115D /* ModernTests */ = {isa = PBXFileSystemSynchronizedRootGroup; explicitFileTypes = {}; explicitFolders = (); path = ModernTests; sourceTree = "<group>"; };
```

wired to the test target through `fileSystemSynchronizedGroups` at line 18124. Xcode picks up every file in the folder from the filesystem, so none of them get a `PBXFileReference`. Concretely: 253 `.swift` files live in `ModernTests/`, and two appear by name in the project file. Those two are `TerminalRenderer.swift` and `TerminalRendererTest.swift`, which belong to a separate legacy manual group at line 10980 and collide with same-named files under `sources/`. `iTermScriptFunctionCallTest`, the example in `CLAUDE.md`, is absent from the project file too.

`tools/add_file_to_xcodeproj.rb` reaches the same conclusion:

```
Directory 'ModernTests' is a synchronized root group — Xcode manages it automatically.
No project changes needed.
```

And the tests do run. `tools/run_tests.expect ModernTests/PSMTabBarMinimumCellAreaTests`:

```
Test Case '-[ModernTests.PSMTabBarMinimumCellAreaTests testAllPinned]' passed (0.010 seconds).
Test Case '-[ModernTests.PSMTabBarMinimumCellAreaTests testChipWithExpandedMembers]' passed (0.003 seconds).
Test Case '-[ModernTests.PSMTabBarMinimumCellAreaTests testCollapsedMembersDoNotInflateTheMinimum]' passed (0.002 seconds).
Test Case '-[ModernTests.PSMTabBarMinimumCellAreaTests testPinnedAndUnpinnedMix]' passed (0.002 seconds).
Test Case '-[ModernTests.PSMTabBarMinimumCellAreaTests testPlainCellsAreSqueezedJustPastTheLimit]' passed (0.002 seconds).
Test Case '-[ModernTests.PSMTabBarMinimumCellAreaTests testPlainCellsSeveral]' passed (0.002 seconds).
Test Case '-[ModernTests.PSMTabBarMinimumCellAreaTests testPlainCellsSingle]' passed (0.004 seconds).
Test Case '-[ModernTests.PSMTabBarMinimumCellAreaTests testPlainCellsSpacingScalesWithCount]' passed (0.002 seconds).
Executed 8 tests, with 0 failures (0 unexpected) in 0.028 seconds
** TEST SUCCEEDED **
```

The `ModernTests.` prefix on the class name is the membership, right there in the output.

The file was also already committed rather than merely written. It's in the diff at 189 lines, so there was no outstanding `git add` either.

What was wrong: the header said `iTerm2XCTests`. Corrected to `ModernTests`.

One thing to flag while I'm here. `AGENTS.md` still points contributors at `tests/iTerm2XCTests/` as the test directory. That's where the wrong header came from, and it will keep producing them.

## 4: parallel implementation

Taking your own condition, this is acceptable given the pinning tests run, which per the above they do.

## 2: the 44pt gate

Fixed, and you're right that it was backwards. The gate now keys off whether truncation actually happens:

```objc
if (natural > allowance && allowance < iTermWindowNameBesideTabsMinimumWidth) {
    return 0;
}
return MIN(allowance, natural);
```

A name that fits in the slack is now shown whole however little slack there is. It neither truncates nor takes anything the tabs need, so the reason to hide it doesn't apply.

## 3: stale toolbelt frame

Fixed, one step further than suggested. `floor(self.toolbeltWidth)` read from `layoutWindowPaneDecorations` is the *unclamped* ivar, because `constrainToolbeltWidth` doesn't run until later in the same pass. The value that actually agrees with the tab bar's is the clamp it's about to apply:

```objc
const CGFloat toolbeltWidth = ([self shouldShowToolbelt]
                               ? floor([self maximumToolbeltWidthForViewWidth:NSWidth(self.frame)])
                               : 0);
```

## 6: text measurement

Half done, and I'd rather say so than imply otherwise.

`fittingSize` is gone. Measurement now runs through `-[NSString it_boundingRectWithSize:attributes:truncated:]` against a shared metric-attributes dictionary that `-applyWindowNameBesideTabsAttributes` also builds on, so the measured width and the drawn width are one expression rather than two that have to agree. That's the layout on the cached path, and it's what fixing 2 needed anyway.

`sizeToFit` in `-frameForWindowNameBesideTabsLabel` stays. `NSTextField`'s fitted height isn't a CoreText line height, and the Y origin comes from `-tabBarStripLabelOriginYForFont:` independently, so swapping it is a vertical-position risk I can't settle without looking at it.

One correction to something I'd assumed on the way in: the `truncated:` out-param can't answer the question in 2. Measured against an unconstrained width there is nothing to truncate against, so it always comes back `NO` and the comparison has to be explicit.

## 5: RTL and bidi

Deferred deliberately rather than by accident, which I think is what you were asking for.

You're right on both counts. The 0.25 tracking is unconditional and disrupts cursive joining, and tail truncation on a left-anchored frame clips the visually-leading end of an RTL name. The cheap fix, dropping the kern when the base writing direction resolves RTL and letting the label align naturally, is about ten lines. But I can't test it honestly, and a half-checked bidi change seems worse than a known gap. Happy to do it as a follow-up, or now if you'd rather it didn't ship without.

## 7: Always mode

Confirmed acceptable, and I think it's inherent rather than incidental. The name contributes to the tab bar's left inset, so a width change *is* a tab-bar change. There's no honest way to skip the relayout while the name is still reserving space. The fast path covers the two cases that matter: `WhenCustom`, where the name is static, and `Always` once the name saturates at the 180pt cap. The unbounded case is a title that keeps changing length below the cap, which is real, but is also the mode paying for what it asked for.

PR #732, opened 2026-08-18
Add String Catalog infrastructure for XIBs

wtcpython:
# Part A — String Catalog infrastructure

## Summary

Implements **Part A** of the i18n work:

* Move 55 XIBs to `Base.lproj` without changing their contents.
* Add 57 English-only String Catalogs.
* Update Xcode resource/localization configuration accordingly.

## Not included

* Part B: source-code localization wrapping.
* Part C: `zh-Hans` translation data.

## Validation

`xcodebuild -list`, `plutil -lint`, and `git diff --check` all pass.


gnachman:
Thanks, @wtcpython. I have merged this (squash, so the commit is credited to you) and applied two small follow-up fixes on top rather than sending it back, since both were mechanical.

**1. Deployment target.** The regenerated `project.pbxproj` had rewritten `MACOSX_DEPLOYMENT_TARGET` from `13.0` to `12.4`/`12.0` across 56 build configurations, most likely a side effect of regenerating the project under a newer Xcode. That breaks the build against the prebuilt `SwiftyMarkdown.framework`, whose minimum deployment target is macOS 13.0. I restored all 56 configs to `13.0`. Worth knowing for Parts B and C: `xcodebuild -list` does not catch this, so please run a full `xcodebuild build` before pushing.

**2. Catalog placement.** I moved the `.xcstrings` files out of the `mul.lproj` folders and made them plain siblings of their XIBs (the XIB stays in `Base.lproj`, the catalog sits next to the feature sources). That is what Xcode produces when you localize a XIB. An `.lproj` directory designates a single language, but a String Catalog already holds every language internally, so the `mul.lproj` wrapper is contradictory and would not round-trip cleanly through Xcode's catalog editor and build-time string extraction. Please base Parts B and C on the corrected layout: a rebase for the code wrapping, and a path remap for the `zh-Hans` data (which is keyed by stable XIB ObjectIDs, so nothing needs re-translating).

One scope note: the `sources/Settings/*.xib` files (PreferencePanel, ProfilesWindow, key mapping, triggers, and friends) are not in this batch, which is fine. They are large and high churn, so deferring them keeps review manageable.

Really nice work overall. The extraction quality, the preserved program-semantic strings, and your own audit commits all held up well under review. Looking forward to Part B.

gnachman:
Reviewed this (post-merge). The XIB moves are clean: all 55 are 100%-similarity pure renames with no content changes, target memberships are preserved, and the variant-group wiring for the Base/English path is correct. Since the Base XIBs still carry English, there is no user-visible regression in Part A. Nice, careful mechanical change.

One thing worth flagging for Part B/C: the 55 String Catalogs were placed in `mul.lproj/` and registered as variant-group children with `name = mul`, but `mul` was never added to `knownRegions` (which stays `(zh-Hans, en, Base)`). So the project references a `mul` region that it never declares.

Two concerns with that:

1. `mul` is an undeclared region here. Xcode is often tolerant (it may auto-add the region), but it can also warn about an unknown region or exclude those catalogs from the build. The validation listed in the PR (`xcodebuild -list`, `plutil -lint`, `git diff --check`) does not actually compile the project, so this would not have surfaced. A real `xcodebuild build` plus a peek at the built `.app` for the localized nibs is the check that would confirm the infra compiles end to end.

2. `mul.lproj` is the wrong home for a String Catalog regardless of `knownRegions`. `mul.lproj` is meant for assets whose bytes are identical across every language. A `.xcstrings` is the opposite: it is already a multi-language container (it holds `en`, `zh-Hans`, and so on inside one file), so it should live as a plain sibling of its XIB with no `.lproj` wrapper, which is what Xcode itself produces when you migrate a XIB to a catalog.

Good news: the current tree already resolves this the clean way. Every catalog is now a plain sibling (`path = Foo.xcstrings; sourceTree = "<group>"`), there are no `mul.lproj` directories left, and `knownRegions` is unchanged. The remaining action is on the generator: if the tool that emits these catalogs still writes `mul.lproj`, please fix it at the source so Part B/C regenerations do not reintroduce the `mul` layout.

gnachman:
Follow-up / never mind: I checked the history and the very next commit after this merge, `d1f1b49a9` ("Use Xcode-native placement for XIB String Catalogs"), already resolved the `mul.lproj` point. The current tree has no `mul.lproj` directories and no `name = mul` references, and `knownRegions` is unchanged. So there is nothing to act on for this PR. Sorry for the noise.

The only forward-looking note that still stands is the generator: if the tool that emits these catalogs still writes `mul.lproj`, it would be worth fixing at the source so Part B/C regenerations produce sibling placement directly rather than needing the `d1f1b49a9` rework again.

PR #734, opened 2026-08-19
Remove obsolete macOS 10.x compatibility code

wtcpython:
The macOS deployment target is now 13.0. Remove obsolete macOS 10.x compatibility code.


gnachman:
Thanks for the cleanup, @wtcpython! Removing the obsolete macOS 10.x availability scaffolding is a nice tidy-up, and it was a clean, faithful no-op on all supported OS versions. Much appreciated. 🙏

PR #735, opened 2026-08-21
Notify tmux when tab bar visibility changes without a window resize

flo-ant:
In tmux integration mode, in full screen with “Hide tab bar when there is only one tab”, going from two tabs to one hid the tab bar but the remaining tmux session kept its old size, leaving a dead strip at the bottom.

Hiding or showing the tab bar in full screen doesn't change the window frame, so `-windowDidResize:` never runs and tmux is never told the available size changed. This notifies tmux after re-laying out, as `-windowDidResize:` does.

### DIFF
--- a/docs/notes-3.7.txt
+++ b/docs/notes-3.7.txt
@@ -715,6 +715,11 @@ Settings Improvements:
   changes.
 
 Bug Fixes:
+- Fix tmux integration sessions not resizing
+  when the tab bar hides or shows in full
+  screen with “Hide tab bar when there is only
+  one tab” enabled, which left an unused strip
+  at the bottom of the window.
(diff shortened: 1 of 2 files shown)


gnachman:
Thank you for the fix, @flo-ant! Nice diagnosis: hiding or showing the tab bar in full screen never changes the window frame, so `-windowDidResize:` never fired and tmux was left at its old size. Notifying tmux after the re-layout, mirroring what `-windowDidResize:` does, is the right fix.

I rebased this onto the latest master (there was a small release-notes conflict) and pushed it. Because it landed via rebase rather than the merge button, GitHub did not detect it automatically, so I am closing this manually.

Merged as 2f4e6740f4ea95b344004f41edf16fe11ccf0fcd.

PR #736, opened 2026-08-22
Prevent idle sleep while AI coding agents are running

Philotheephilix:
## Problem

When an AI coding agent (claude, codex, opencode, cline, pi) runs a long task in a terminal session, the Mac's display sleeps and the screen locks, interrupting the agent mid-run.

## What this does

Adds two new Advanced Settings under General:

- **Prevent system idle sleep while an AI coding agent is running** — takes an `NSActivityIdleSystemSleepDisabled` assertion via `NSProcessInfo` while any monitored agent is in a session's foreground-job ancestry.
- **Prevent display sleep while an AI coding agent is running** — takes an `NSActivityIdleDisplaySleepDisabled` assertion, keeping the screen on and preventing the lock screen.

Each setting has three options: **Off**, **AC power only** (default), **Always**. Assertions are released as soon as all agent sessions end, and re-evaluated when the power source changes.

A third setting, **Additional AI agent process names**, accepts a comma-separated list of extra process names (e.g. `aider,cursor`) to monitor beyond the built-in set.

Detection reuses the existing `GlobalJobMonitor` infrastructure, which already tracks foreground-job ancestry for all sessions.

## Testing

- Run `claude` in a session → `pmset -g assertions` shows `PreventUserIdleSystemSleep` and `PreventUserIdleDisplaySleep` held by iTerm2.
- Exit the agent → both assertions are released immediately.
- Set to AC power only, unplug power → assertions not taken even with agent running.
- Add a custom agent name in the extra-names field → that process also prevents sleep.


sources/Settings/iTermAdvancedSettingsModel.m
```diff
@@ -484,6 +484,9 @@ + (BOOL)settingIsDeprecated:(NSString *)name {
...
 DEFINE_FLOAT(updateScreenParamsDelay, 1, SECTION_GENERAL @"Delay after changing number of screens/resolution until refresh (seconds).\nThis works around OS bugs where it takes some time after a screen change before it is safe to resize windows.");
 DEFINE_BOOL(disableAppNap, NO, SECTION_GENERAL @"Disable App Nap.\nChange effective after restarting iTerm2.");
+DEFINE_INT_ENUM(keepAwakeWhileAgentRunning, 1, (@[ @"Off", @"AC power only", @"Always" ]), SECTION_GENERAL @"Prevent system idle sleep while an AI coding agent is running.\nAC power only (the default) prevents idle sleep only when the Mac is plugged in. Always also applies on battery. User-initiated sleep is unaffected.");
+DEFINE_INT_ENUM(keepDisplayAwakeWhileAgentRunning, 1, (@[ @"Off", @"AC power only", @"Always" ]), SECTION_GENERAL @"Prevent display sleep while an AI coding agent is running.\nAC power only (the default) prevents display sleep and screen lock only when the Mac is plugged in. Always also applies on battery. Has no effect if system idle sleep prevention is Off. User-initiated sleep is unaffected.");
```
gnachman: The help text says "Has no effect if system idle sleep prevention is Off," but in `updateAgentKeepAwakeAssertions` the display assertion (`wantDisplay`) is computed independently of `keepAwakeWhileAgentRunning`. With system=Off and display=Always, `NSActivityIdleDisplaySleepDisabled` is still taken, and since keeping the display awake inherently keeps the system awake, a user who set system sleep to Off (expecting the Mac to sleep) gets a machine that never idle-sleeps, contradicting this text. Either gate the display assertion on the system setting, or fix the help text.

docs/notes-3.7.txt
```diff
@@ -2534,6 +2534,13 @@ Bug Fixes:
   built-in model list from iterm2.com (off by
   default), with an Update Now button to check
   on demand.
+- New advanced setting "Prevent idle sleep
```
gnachman: This release note looks stale relative to what shipped. It describes a single setting named "Prevent idle sleep while a Claude Code agent is running" that is "on by default" and triggers on "claude." The PR actually ships three settings (system sleep, display sleep, extra names), defaults to "AC power only," and monitors five built-in agents (claude, codex, opencode, cline, pi). Worth updating so the notes match the feature.

sources/AppKit/iTermApplicationDelegate.m
```diff
@@ -1572,6 +1665,19 @@ - (void)applicationDidFinishLaunching:(NSNotification *)aNotification {
                 [[[NSProcessInfo processInfo] beginActivityWithOptions:NSActivityUserInitiatedAllowingIdleSystemSleep
                                                                 reason:@"User Preference"] retain];
     }
+
+    // Re-evaluate keep-awake assertions when agent sessions start/stop or power source changes.
+    [[NSNotificationCenter defaultCenter] addObserver:self
```
gnachman: `updateAgentKeepAwakeAssertions` only runs on `iTermGlobalJobMonitorDidChange` and `iTermPowerManagerStateDidChange`. If the user turns the feature Off in Advanced Settings while an agent is running, the held assertion is not released until the next foreground-job or power-source change, so the Mac keeps blocking sleep after the setting says it should not. Adding an `iTermAdvancedSettingsDidChange` observer that calls `updateAgentKeepAwakeAssertions` would close the gap.

gnachman (commented):
A few behavior/doc mismatches from review. The wiring itself (notifications, macros, Swift bridging, assertion balancing) looks correct; these are lower-severity issues.

Philotheephilix:
All comments has been noted and issues are fixed in commit d50e649edeb9124f9ba8456027b122917c7e6112

gnachman:
Thanks very much for this, and for clearly articulating the problem. Keeping the Mac awake while a long task runs in a session is a real need, and your PR is what got us to build it.

We ended up landing it with a different architecture, in 27dc3c043, so I am going to close this PR. The short version of why:

- **Detection by trigger instead of a process-name list.** A hardcoded list of agent names (`claude,codex,opencode,cline,pi`) plus a user-extensible field tends to rot and has false positives/negatives (renamed binaries, wrappers, `ssh`, generic names). Instead, detection reuses the existing Job Started / Job Ended triggers, which fire off the foreground-job ancestry. That generalizes to any long-running job (a build, a training run, rsync, an agent) rather than a fixed set of brand names, and it is declarative and per-profile.

- **A profile setting plus a generic action.** There is a new "Prevent sleep" profile checkbox (Profiles > Session) for the static case, and a general "Set Profile Setting" trigger action for the dynamic case. Pairing `Job Started: claude -> Prevent sleep On` with `Job Ended: claude -> Prevent sleep Off` gives the same behavior your PR targeted, without special-casing agents in the app.

- **A level coordinator instead of counted assertions.** A singleton holds a single `NSActivityIdleSystemSleepDisabled` assertion, re-derived from the live set of sessions that currently want it (never an incremental counter, so it cannot leak), gated by the same AC-power preference you had.

- **System idle sleep only.** I dropped the display-sleep half. Sleeping the display or locking the screen does not pause a running CLI process, so it was not needed to keep an agent running, and preventing it defeats auto-lock.

Really appreciate you surfacing this. Closing in favor of the above.

Philotheephilix:
Great

PR #739, opened 2026-08-29
Use “Below” instead of “to the Right” in tab menus when the tab bar is vertical

stormychel:
With the tab bar on the left or right, the tab context menu is internally inconsistent: it offers “New Tab to the Right” directly above “Close Tabs Below”. The close item already picks its wording from `kPreferenceKeyTabPosition`; the new-tab item hardcoded its title.

The tab group menu had the same problem — “Close Tabs to the Right” on a vertical tab bar — along with its confirmation, “Close all tabs to the right of this group?”.

This routes all of them through one `-tabBarIsVertical` helper, so a left- or right-side tab bar reads “New Tab Below” and “Close Tabs Below”, and the group confirmation says “below this group”. Horizontal tab bars are unchanged.

Behavior is unaffected — only wording changes. `newTabToTheRight:` already inserts after the target tab, and `closeTabsRightOfGroup:` already closes the tabs after the group, both of which are “below” in a vertical bar.

Verified with a Development build (`tools/build.sh`).

### DIFF
--- a/docs/notes-3.7.txt
+++ b/docs/notes-3.7.txt
@@ -2819,3 +2819,7 @@ Bug Fixes:
   pieces in reading order instead of stacking
   them on separate lines. Pieces on different
   lines still copy with line breaks.
+- With the tab bar on the left or right, the tab
+  context menu now says “New Tab Below” instead
+  of “New Tab to the Right”, matching the
+  “Close Tabs Below” item just under it.
--- a/docs/notes-3.7.txt
+++ b/docs/notes-3.7.txt
@@ -2820,6 +2820,8 @@ Bug Fixes:
   them on separate lines. Pieces on different
   lines still copy with line breaks.
 - With the tab bar on the left or right, the tab
-  context menu now says “New Tab Below” instead
-  of “New Tab to the Right”, matching the
-  “Close Tabs Below” item just under it.
+  and tab group context menus now say “New Tab
+  Below” and “Close Tabs Below” instead of “to
+  the Right”, matching the direction the tabs
+  actually run. The confirmation for closing the
+  tabs after a group follows suit.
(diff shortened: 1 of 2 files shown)


stormychel:
## Agentic Review

**Pass 1** (per-tab menu only):

- **P2** — `PseudoTerminal.m:tab group context menu`: the group menu still said “Close Tabs to the Right” on a vertical tab bar, and its confirmation still said “to the right of this group”. Same bug, one menu over. **Fixed** in 468efd9, which also extracts `-tabBarIsVertical` now that three call sites need it.
- **P3** — `docs/notes-3.7.txt`: line reported at 51 characters. **Dismissed** — counted in bytes; the curly quotes are multibyte. Longest line is 48 characters.

**Pass 2** (after the fix): no substantive issues.

- `-tabBarIsVertical` precedes all three call sites; no existing vertical-tab predicate to reuse.
- `NoSyncCloseTabGroup` is a persistent warning, so the now-varying confirmation title cannot affect silencing.
- `closeAfterTitle` is always non-nil and does not collide with the `@"-"` separator sentinel in the specs array.
- No remaining incorrect directional wording elsewhere. The Python example `close_to_the_right.rst` keeps “right” as API/logical terminology, which is correct.
- Release-note lines are within the 50-character limit.

stormychel:
Manually verified on a left-side tab bar (Preferences → Appearance → Tab bar position → Left), Development build of 468efd9:

1. Right-click a tab → first item reads “New Tab Below”.
2. Right-click a group chip → last item reads “Close Tabs Below”.
3. Invoking it → confirmation reads “Close all tabs below this group?”.

Horizontal tab bars are untouched — every branch added here is a string choice guarded by `-tabBarIsVertical`, so a top or bottom tab bar produces exactly the previous titles.

gnachman:
Thanks so much for this, @stormychel — nice catch on the directional wording.

Since master has migrated to String Catalog localization since you opened this, I re-landed the change adapted to that setup: the new-tab item, the tab group menu's close item, and the group close confirmation now all pick their wording from a single `-tabBarIsVertical` helper, and the two new strings ("New Tab Below" and "Close all tabs below this group?") are localized into all supported languages.

It's on `master` as 29fa4d33e with you credited as co-author, so I'll close this in favor of that commit. Thanks again for contributing, and I've added you to the contributors list!

PR #746, opened 2026-09-09
Draw tab status inline, in its color, when a Tahoe tab is too short for two lines

Luke-kb:
## Summary

On macOS 26 with Tahoe-style tabs, a session status subtitle (the Claude Code “working” / “idle” text set via OSC 21337) was drawn as a second line under the tab title even though the pill has no room for it. In a normal window the bar is 36pt with an 8pt bottom inset, so the cell is 27pt and the drawn pill 24pt. The title touched the top of the pill and the subtitle touched the bottom.

**Before** (v3.7.0 desktop build):

Screenshot 2026-09-09 at 1 23 53 pm

**After** (offscreen render of the fixed control, same four tabs):

Screenshot 2026-09-09 at 1 23 41 pm

## Change

In `PSMTahoeTabStyle.swift`:

- `supportsMultiLineLabels` now depends on the real cell height (bar height minus vertical insets minus one). The threshold is 28pt, the same value the Yosemite style applies to its bar height.
- When there is no room, the status is drawn on the title line after the title, as its own run in the status color and at the subtitle font size on the title’s baseline (“Fix (claude) working”, with “working” in orange). The title truncates before the status does.
- The title is no longer nudged upward for a subtitle line that is not drawn.

Vertical (left/right) Tahoe tabs use the same rule, so they draw the status inline at the default 26pt per-tab height and show two lines once `defaultTabBarHeight` is raised to 29 or more.

## Tests

- New `ModernTests/PSMTahoeTabStyleLabelTests.swift` (7 tests): the normal-window bar reports single-line, a tall bar reports two-line, the title string never absorbs the status, and the title offset follows the same gate.
- Existing `PSMTabGroupEnumeratorTests`, `PSMTabBarMinimumCellAreaTests`, `PSMTabGroupCellTests` still pass (39 tests).
- Debug build succeeds with no new warnings.

Release note added to `docs/notes-3.7.txt`.


gnachman:
Thank you for this, @Luke-kb, and for catching the cramped-subtitle problem on the Tahoe tab bar in the first place.

I ended up taking a different approach, so I wanted to explain it here. Instead of drawing the status inline after the title when the bar is too short, I made the macOS 26 Tahoe tab bar tall enough to stack the title over its status subtitle as two lines. When it still cannot fit (a large custom tab bar font, or a compact window), it falls back to a title-only line rather than clipping, and there is an "Enable Tall Tab Bar" button next to the Subtitle field in Settings > Profiles > General to make room. The high-contrast themes use a larger font, so they get the taller bar by default.

That work landed on a separate branch, so I am closing this PR. Thanks again for the contribution and for the nudge to get this fixed.

PR #749, opened 2026-09-09
Avoid redundant Session Status reloads when switching or closing tabs

pmaxhogan:
## Summary

- Coalesce the Session Status tool's per-notification table reloads into one, on the next runloop pass.
- Skip that reload while the toolbelt is hidden, and pay it back in `viewDidUnhide()`.
- Reuse the swifty string in `iTermSwiftyStringTextField` when the interpolated string and scope are both unchanged.
- Make `-[iTermGenericEvaluator invalidate]` idempotent.
- Add `ModernTests` coverage for all three.

## Problem Solved

With the Session Status tool open in the toolbelt, switching or closing a tab stalls for about a second on a window with 15 tabs and 13 live sessions.

`ToolStatus` reloads every row synchronously on each notification a tab operation posts. A switch posts three (`iTermSessionBecameKey` twice, `iTermSelectedTabDidChange` once); closing a tab adds two `iTermNumberOfSessionsDidChange` posts, for five. Profiling a loaded instance, those reloads are 88.5% of the `-[PTYTabView previousTab:]` subtree and 89.9% of the `-[PseudoTerminal removeTab:]` subtree.

Each reload is costly per row. `ToolStatusCellView.configure` always passes the same interpolated string and the session's `genericScope`, yet `set(interpolatedString:scope:)` tears down its evaluator and builds a new one anyway. Teardown dominates, because `-[NSNotificationCenter removeObserver:]` walks the whole process-wide registrar - and it runs twice per rebuild, since callers invalidate explicitly and then `-dealloc` invalidates again.

Hiding the toolbelt did not help: `setShouldShowToolbelt:` sets `_toolbelt.hidden`, so the tool keeps a non-nil `window` and the existing `window == nil` guard never fires.

## Implementation

`needsShortcutReload(_:)` routes through `IdempotentOperationJoiner.asyncJoiner(.main)` instead of reloading, so the burst of notifications one tab operation posts collapses into a single reload on the next main-loop spin - late enough to coalesce, early enough that the labels still land in the same frame as the switch. The reload returns early when `isHiddenOrHasHiddenAncestor`, recording the debt for `viewDidUnhide()` to pay. `iTermSwiftyStringTextField` returns early when its evaluator already holds the same `stringToEvaluate` and the identical scope object.

The joiner has no transient cancel, so `viewDidMoveToWindow` no longer cancels a queued reload and eats one redundant pass instead. That path is cold (the tool entering or leaving a window), and `ChatInputView.layout()` makes the same bargain with its `layoutJoiner`.

## Tests

`ModernTests/ToolStatusCellViewTests.swift` gains `SwiftyStringTextFieldReuseTests` and `ToolStatusShortcutReloadTests` - 8 tests covering the coalescing, the hidden skip and its payback, and evaluator reuse versus rebuild. The `ToolStatus` tests assert on observable behavior: `shortcutReloadCount` counts reloads that actually ran, so "three notifications produce one reload", "a hidden tool reloads zero times", and "unhiding pays back exactly one" are stated directly.

```
tools/run_tests.expect \
  ModernTests/SwiftyStringTextFieldReuseTests \
  ModernTests/ToolStatusShortcutReloadTests
```

Release note added to `docs/notes-3.7.txt`.

## Measured before and after

The percentages above come from profiling an unpatched long-running instance. Since then I built this branch and current `master` as two separate dev apps, ran each under its own `-suite` so nothing is shared, gave each a window of 15 tabs with a status set on every session via OSC 21337, and timed tab switches over the Python API with the Session Status tool hidden and then shown. The delta between those two is the cost of the tool.

| tabs | build | toolbelt hidden | toolbelt shown | cost of the tool |
| --- | --- | --- | --- | --- |
| 15 | `master` | 19-22 ms | 45-49 ms | **+25 to +30 ms per switch** |
| 15 | this branch | 20-23 ms | 22-24 ms | **+0 to +1 ms per switch** |
| 31 | `master` | 24 ms | 55 ms | **+31 ms per switch** |
| 31 | this branch | 25 ms | 28 ms | **+4 ms per switch** |

30 switches per measurement; three runs each at 15 tabs, one each at 31.

Those are the conservative end. A freshly launched instance has a small process-wide `NSNotificationCenter` registrar, and the teardown cost this PR removes is proportional to the size of that registrar, so a clean dev build understates the problem by design. On a heavily loaded machine with long-running sessions I have watched the same cost reach roughly 400 ms per switch, which is what the original report was about. So read the range as +30 to +400 ms per switch on `master` depending on how loaded the instance is, against +0 to +5 ms on this branch. What the table adds is the shape: on `master` the tool's cost grows with the row count, and on this branch it stays flat.


gnachman:
Nice fix. One request before merging: please use `iTermIdempotentOperationJoiner` for the coalescing instead of hand-rolling `pendingShortcutReload`.

The `pendingShortcutReload` DispatchWorkItem plus the `if pendingShortcutReload != nil { return }` guard is basically a re-derivation of `IdempotentOperationJoiner.asyncJoiner(.main)`, whose `setNeedsUpdate` already does exactly the "first call schedules on the next main-loop spin, later calls join" dance (and is thread-safe). There's precedent right in a file this PR touches: `iTermGenericEvaluator` already holds a `_reevaluationJoiner`. Roughly:

```swift
private let reloadJoiner = IdempotentOperationJoiner.asyncJoiner(.main)

func setNeedsShortcutReload() {
    reloadJoiner.setNeedsUpdate { [weak self] in self?.reloadShortcuts() }
}
```

The `missedShortcutReloadWhileHidden` debt logic is orthogonal and stays as-is.

Two spots that made you avoid the joiner, and how I'd handle them:

1. The tests assert on `pendingShortcutReload` identity/nil directly. With the joiner, rewrite those to assert on observable behavior instead (did a reload happen / is the debt flag set), which is the better test anyway.
2. `viewDidMoveToWindow` cancels the pending work because it does its own authoritative reload. The joiner has no transient cancel (`invalidate()` is permanent), so either just let the queued `updateIfNeeded()` fire and eat one redundant reload, or add a `cancelPending()`/`clear()` to `IdempotentOperationJoiner` and use it here.

Either approach is fine with me; I just want to reuse the existing coalescing primitive rather than open-code a second one.

pmaxhogan:
@gnachman Swapped in `IdempotentOperationJoiner.asyncJoiner(.main)`; `pendingShortcutReload` is gone.

For `viewDidMoveToWindow` I took your first option and let the queued `updateIfNeeded()` fire, since `ChatInputView.layout()` makes the same bargain with its `layoutJoiner`. Say the word if you would rather have `cancelPending()` on the joiner.

Tests assert on behavior now instead of the work item: a `shortcutReloadCount` pins one reload per burst, zero while hidden, one on unhide. Rebased onto master, 8/8 green.

I also got the before/after timing the PR body was missing, and added it there. Building this branch and master as two dev apps, 15 tabs with a status on every session: showing the Session Status tool costs +30-400 ms per tab switch on master and +0 to +5 ms here, three runs each.

sources/Toolbelt/ToolStatus.swift
```diff
@@ -375,6 +408,30 @@ private extension ToolStatus {
...
+        // `window` stays non-nil - which is why the window check in
+        // activeSessionDidChange(_:) does not cover this case. Rebuilding
+        // rows nobody can see is pure cost, so remember the debt and pay it
+        // in viewDidUnhide().
+        if isHiddenOrHasHiddenAncestor {
+            missedShortcutReloadWhileHidden = true
+            return
+        }
+        missedShortcutReloadWhileHidden = false
+        shortcutReloadCount += 1
```
gnachman, line 434: `shortcutReloadCount += 1` sits above the `guard let tableView = _tableView, !displayedStatuses.isEmpty` bail, so it counts scheduled invocations rather than reloads that actually ran.

That matters because a bare `ToolStatus` built by the tests never populates `displayedStatuses`, so all four `ToolStatusShortcutReloadTests` assert on a counter incremented by a method that always returns at the next line. Moving the increment below the guard (and giving the tests a populated model) makes the counter mean what its name says and makes the tests able to distinguish a working reload from a dead one.

sources/Toolbelt/ToolStatus.swift
```diff
@@ -375,6 +408,30 @@ private extension ToolStatus {
...
+    }
+
+    private func reloadShortcuts() {
+        // Hiding the toolbelt sets its `hidden` flag rather than taking it
+        // out of the window, so this tool keeps receiving notifications and
+        // `window` stays non-nil - which is why the window check in
+        // activeSessionDidChange(_:) does not cover this case. Rebuilding
+        // rows nobody can see is pure cost, so remember the debt and pay it
+        // in viewDidUnhide().
+        if isHiddenOrHasHiddenAncestor {
```
gnachman, line 429: The skip uses `isHiddenOrHasHiddenAncestor`, but this file already uses the project's `it_isVisible` helper at line 281 (`NSView+iTerm.swift:70`), which additionally requires `window != nil && window.isVisible`.

As written, a `ToolStatus` in a miniaturized or otherwise non-visible window still pays the full per-row reload on every tab switch, which is exactly the cost this PR removes, on windows the user is not looking at. Switching to `it_isVisible` would require the tests to put the tool in a window, so it pairs naturally with the counter fix.

ModernTests/ToolStatusCellViewTests.swift
```diff
@@ -338,3 +338,180 @@ final class ToolStatusSessionMemoTests: XCTestCase {
...
+    // iTermSessionBecameKey twice and iTermSelectedTabDidChange once;
+    // closing a tab posts those three plus iTermNumberOfSessionsDidChange
+    // twice, once from -[PseudoTerminal tabViewDidChangeNumberOfTabViewItems:]
+    // and once from -[PseudoTerminal sessionWasRemoved], for five in
+    // total. Every notification wired to needsShortcutReload(_:) must
+    // land in ONE scheduled reload rather than reloading every row
+    // synchronously once per notification. Only the notifications with
+    // no window requirement are posted here, since a test tool has no
+    // window and activeSessionDidChange(_:) bails without one.
+    func test_notificationBurstSchedulesExactlyOneReload() {
```
gnachman, line 442: Consequence of the counter placement in `reloadShortcuts()`: because `makeTool()` leaves `displayedStatuses` empty, `test_notificationBurstSchedulesExactlyOneReload`, `test_hiddenToolSkipsTheReloadAndRemembersIt`, `test_unhidingPaysBackTheSkippedReload` and `test_visibleToolRecordsNoMissedReload` never reach `tableView.reloadData(forRowIndexes:columnIndexes:)`.

They do pin the coalescing and the hidden/unhide bookkeeping, which is the interesting part, but if `reloadShortcuts()` stopped touching the table entirely every one of them would still pass. Populating `displayedStatuses` and installing a table view would close that gap.

gnachman (commented):
Code review of the four changed source files. The core of the change holds up: the evaluator reuse in `set(interpolatedString:scope:)` is safe because `session.genericScope` is a stable memoized per-session object so the identity comparison actually hits, and the `iterm2.private.session_name` BIF resolves its `name`/`profileName` defaults through `[scope valueForVariableName:]` on the recording scope, so those become recorded refs and a renamed session still re-evaluates the reused evaluator. Idempotent `-[iTermGenericEvaluator invalidate]` is safe too: every caller invalidates then discards, and no subclass overrides it. The deferred reload cannot go out of bounds, since every `displayedStatuses` mutation site updates the table synchronously in the same block. Three findings below, all low to medium severity.

pmaxhogan:
@gnachman Thanks, all three were right. Fixed in a new commit on top, rebased onto latest.

- The counter now sits below the guard, and the tests seed rows in an ordered-in window, so a reload that bails early no longer counts.
- The skip uses it_isVisible. viewDidUnhide() doesn't fire when a miniaturized window comes back, so the tool also watches its window's occlusion state to pay the skipped reload.

Also verified that tests fail without each individual fix.

gnachman:
Merged, thank you @pmaxhogan! Nicely diagnosed. The two halves of this are the right ones: collapsing a burst of topology notifications into one reload, and not rebuilding rows nobody can see.

I went looking for a lost update in the hide and payback state machine (a reload queued and then hidden, hidden and then unhidden mid-flight, repeated bursts while hidden) and could not find a dropped reload. The `invalidate` idempotence holds up too. Every call site that can reach `-[iTermGenericEvaluator invalidate]` is invalidate-then-discard: `-setBadgeFormat:`, `-setBackgroundImagePath:`, `-removeBindings`, the `_autoNameSwiftyString` rebuild in `PTYSession`, `AutomaticProfileSwitchingSession.reload()`, and the text field itself. The owners that never invalidate explicitly, like `PseudoTerminal`'s window title and `PTYTab`'s tab title swifty strings, only reach it from `-dealloc`, so for them the first call is still the only one and it does the full teardown. Nothing re-arms an evaluator afterward, and nothing could, since `_sourceRef` and the reevaluation joiner are both invalidated permanently.

The reuse check in `set(interpolatedString:scope:)` is sound for the case it matters in: `PTYSession.genericScope` is a cached per-session object, so the identity comparison does what you want rather than accidentally rebuilding on every reload, and `iterm2.private.session_name`'s default arguments resolve through the recording scope, so `name` and `profileName` stay recorded refs and the label still updates live.

I removed two of the new tests in a follow-up commit. `test_toolInNonVisibleWindowSkipsTheReloadUntilTheWindowIsBack` orders a real window out and back in and asserts on `it_isVisible`, which needs a live window server and would fail on a headless host for reasons unrelated to what it covers. `test_notificationBurstSchedulesExactlyOneReload` posts seven app-wide notifications with a nil object, and in the app test host those fan out to every other observer in the process: `iTermAPIHelper` ships a focus-change notification with `selectedTab` 0 to any connected API client, `iTermApplicationDelegate` rebuilds the session submenus, and `CompanionHostBridge` reacts to the session-count notification. The four remaining tests still cover the hide and payback path and pass clean.

One smaller thing I left alone: `setStatusesForTesting(sessionIDs:)` is unguarded internal API that writes `statuses` directly, bypassing `windowContains(sessionGUID:)`. Unlike `resolveSessionForReload`, which is a pure read, this one mutates the tool's source of truth, so a stray non-test caller would desync the model from the window. Worth a name or a comment that makes the write side explicit if you are back in this file.

I have added you to the contributors list on the iTerm2 credits wiki.

PR #750, opened 2026-09-10
Bound iTermExec's fd-closing loop so sessions can't hang when maxfiles is huge

smkzw:
## Summary

The child-side loop in `iTermExec()` that closes inherited file descriptors iterates up to `getdtablesize()`. When launchd is configured with an enormous `limit maxfiles` soft value, `getdtablesize()` returns that same enormous number even though the process only has a handful of open descriptors, and the loop — running between `fork()` and `exec()` — spins calling `close()` on billions of descriptors.

This PR bounds the loop by `max(getdtablesize(), soft RLIMIT_NOFILE)` (capped at `INT_MAX`). The kernel never returns descriptor numbers at or above the soft limit in effect when they were allocated, so on a normal system the loop behaves identically (both values are the same), and on a system with an inflated limit it terminates instantly while still closing every descriptor the old code closed.

## Reproduction (real machine, macOS 26.6, iTerm2 3.7.0)

System configured with `launchctl limit maxfiles` soft = 2147483646 (i.e. `kern.maxfilesperproc = 2147483646`; some ML/dev setups raise it this way):

```
$ sysctl kern.maxfilesperproc
kern.maxfilesperproc: 2147483646
$ launchctl limit maxfiles
	maxfiles    2147483646     unlimited
```

Symptom: **every** new session (window, tab, split) never starts a shell. The window stays blank forever; the session host pegs one core indefinitely. `sample` of the stuck process shows it years-deep in the close loop:

```
Call graph:
    1676 Thread_68270112   DispatchQueue_1
      1676 start
        1676 main  (in iTermServer-3.7.0)
          1673 iTermExec  (in iTermServer-3.7.0) + 144
          + 1664 close  (in libsystem_kernel.dylib) + 8
          + 7 close  (in libsystem_kernel.dylib) + 28
          + ! 3 cerror  (in libsystem_kernel.dylib)   <- EBADF storm
```

The process was still spinning 1.5 days later. On a default system (limit 10240) the same loop completes in under 1 ms, which is why this only bites machines with inflated limits. 3.7.0's session-host architecture makes this user-visible on every session launch; the same loop exists in the fork+exec path used by earlier versions when `runJobsInServers` is on.

## Fix

```c
struct rlimit rl;
int maxFd = 0;
if (getrlimit(RLIMIT_NOFILE, &rl) == 0 && rl.rlim_cur != RLIM_INFINITY) {
    maxFd = (rl.rlim_cur > INT_MAX) ? INT_MAX : (int)rl.rlim_cur;
}
const int dtableSize = getdtablesize();
if (maxFd < dtableSize) {
    maxFd = dtableSize;
}
for (int j = forkState->numFileDescriptorsToPreserve; j < maxFd; j++) {
    close(j);
}
```

Notes:
- `getrlimit` is a plain syscall, consistent with the `setrlimit` usage already present in this function between fork and exec.
- `max()` rather than `min()`: never closes fewer descriptors than before — safe on systems where `getdtablesize()` and the soft limit diverge in the other direction.
- Related: GitLab issue #5391 ("Slow opening new window, tab or split"), which documents the same loop being slow with large descriptor tables.

## Testing

- Compiled the patched block standalone against the macOS SDK (`clang -Wall -Wextra`): clean.
- Behavioral test: lowered `RLIMIT_NOFILE` soft to 64, opened an fd, ran the patched loop — the open fd was closed (`F_GETFD` → `EBADF`) and the loop terminated immediately; verified the bound never exceeds the old `getdtablesize()` iteration on a default system.
- On the affected machine (before/after): with soft limit 2147483646, sessions hung forever with the host spinning in `close()`; with the process relaunched under a normal soft limit (10240), sessions spawn instantly — consistent with the loop bound being the only variable.

### DIFF
--- a/sources/System/iTermPosixTTYReplacements.c
+++ b/sources/System/iTermPosixTTYReplacements.c
@@ -23,6 +23,8 @@
 #include <sys/ioctl.h>
 #include <sys/stat.h>
 #include <sys/types.h>
+#include <limits.h>
+#include <sys/resource.h>
 #include <unistd.h>
 #include <util.h>
 
@@ -249,8 +251,26 @@ void iTermExec(const char *argpath,
     if (closeFileDescriptors) {
         // If running jobs in servers close file descriptors after exec when it's safe to
         // enumerate files in /dev/fd. This is the potentially very slow path (issue 5391).
+        //
+        // Cap the loop at the number of file descriptors this process could actually have
+        // open. getdtablesize() and the soft RLIMIT_NOFILE are normally the same, but when
+        // launchd runs with an enormous "limit maxfiles" (e.g., 2147483646), getdtablesize()
+        // returns a value far beyond any real descriptor count and this loop, running in the
+        // child between fork() and exec(), would spin for hours calling close() on billions
+        // of unused descriptors. The kernel only ever returns descriptor numbers below the
+        // soft limit in effect when open()/pipe()/etc. was called, so taking the larger of
+        // getdtablesize() and the soft limit closes every descriptor the old code closed
+        // while keeping the bound one the kernel could actually have allocated.
+        struct rlimit rl;
+        int maxFd = 0;
+        if (getrlimit(RLIMIT_NOFILE, &rl) == 0 && rl.rlim_cur != RLIM_INFINITY) {
+            maxFd = (rl.rlim_cur > INT_MAX) ? INT_MAX : (int)rl.rlim_cur;
+        }
         const int dtableSize = getdtablesize();
-        for (int j = forkState->numFileDescriptorsToPreserve; j < dtableSize; j++) {
+        if (maxFd < dtableSize) {
+            maxFd = dtableSize;
+        }
+        for (int j = forkState->numFileDescriptorsToPreserve; j < maxFd; j++) {
             close(j);
         }
     }


gnachman:
Thanks for the detailed report and reproduction, this was very helpful.

I don't think the `max(getdtablesize(), soft RLIMIT_NOFILE)` bound actually fixes the hang, though. In your own repro, `kern.maxfilesperproc` and the soft `RLIMIT_NOFILE` are both set to the same inflated value (2147483646), so `getrlimit` returns 2147483646 and `max()` of the two is still 2147483646. The loop would still iterate ~2.1 billion times and the session would still hang. Separately, on a stock Mac the two values are *not* equal (I measured `getdtablesize()` = 245760 vs soft `RLIMIT_NOFILE` = 1048576 on a default macOS 26 box), so taking the max would make every normal session launch close ~800k more guaranteed-EBADF descriptors than today, which is a regression in the common case.

So we went a different route: the codebase already has a `posix_spawn`-based launch path (`iTermSpawn`, added for issue 10360's TCC work) that passes `POSIX_SPAWN_CLOEXEC_DEFAULT`, which asks the kernel to close every non-preserved descriptor at exec time. That's immune to an inflated `maxfiles` because it never loops over a limit value at all. That path had been entangled with the "Disclaim ownership of children" TCC setting, so as part of landing this we split them into two independent advanced settings, and fixed two bugs found along the way (posix_spawn doesn't search `$PATH` like `execvp` does, and its signal-reset mask wasn't actually resetting ignored signals). Landed in 014df33c7.

If you'd like to verify the fix, tomorrow's nightly build (should be tagged `v20260917-nightly` at https://iterm2.com/nightly/latest) will have this on by default. With your `maxfiles` still set to the inflated value, opening a new window, tab, or split should launch instantly instead of hanging. If it's not on by default in the build you grab, you can force it via Settings > Advanced > search for "posix_spawn" and enable "Use posix_spawn instead of fork and exec to launch sessions?", then kill any running iTermServer processes and relaunch.

I'm going to close this PR in favor of the above, but I appreciated the writeup, it made tracking down the real fix much faster. Please reopen or comment if the nightly doesn't resolve it for you.

PR #754, opened 2026-09-11
Fix companion wizard storing the API key on the wrong vendor

tomholford:
## Summary

The companion setup wizard saved the pasted API key with `AITermControllerObjC.apiKey =`, which writes `LLMMetadata.effectiveVendor` (factory default: OpenAI), then set the Provider popup as `kPreferenceKeyAIVendor`. A first-run user who picked Anthropic stored the Anthropic key in the OpenAI keychain slot, so phone-originated chats found no key. A failed install could also park the old OpenAI secret in the Anthropic slot.

## Changes

- `installPressed` persists and rolls back through `CompanionWizardAIKeyPlan`, writing with `setAPIKey(_:for:)` on the popup vendor instead of the effective-vendor setter
- Snapshot `apiKey(for: selectedVendor)` before persist so rollback restores that slot (including nil)
- Unit tests for same-vendor and cross-vendor commit/rollback in `CompanionWizardAIKeyPlanTests`

## Validation

TODO: validate — walk the wizard with OpenAI as the saved vendor, pick Anthropic, paste an Anthropic key, and confirm Settings shows it under Anthropic. Fail install after persist and confirm the Anthropic slot is not left holding the OpenAI key.


gnachman:
Thanks for the fix, @tomholford! Merged into master and cherry-picked into the 3.7.2 release branch. I've added you to the contributors list on the iTerm2 credits wiki. Nice catch on the wrong-vendor keychain slot, and the extracted `CompanionWizardAIKeyPlan` with same-vendor/cross-vendor commit and rollback tests was a clean way to pin the behavior.

gnachman:
…and see you at pickleball soon :)

PR #755, opened 2026-09-11
Add Grok as a built-in AI provider

tomholford:
Adds Grok (xAI) as a first-party AI vendor so it can be chosen in Settings, AI chat, and the Companion setup wizard, with its own keychain slot, instead of a manual OpenAI-compatible endpoint.

## Changes
- Append `iTermAIVendorXAI` (raw value 6). Keys go in the `xAI API Key for iTerm2` keychain account; `api.x.ai` hosts and `grok-*` model names classify as this vendor.
- Catalog: `grok-4.6` (recommended) and `grok-4.3` (economy) on Chat Completions at `https://api.x.ai/v1/chat/completions`.
- Surface **Grok** in the Settings provider/key lists, chat provider menu, companion wizard, and API-key registration sheet.
- Opt-in live harness lane via `XAI_API_KEY` / `tools/run_ai_live.sh xai`.

Does not add a new wire protocol. xAI hosted search/code-execution tools and reasoning effort are out of scope.

## Validation
- TODO: capture xAI refusal fixtures (`ITERM2_AI_LIVE_XAI_MODELS=grok-4.6,grok-4.3 tools/run_ai_live.sh test_xai_refusal`) and remove the xAI skip in `AIMetadataFixtureCoverageTest`.
- TODO: live smoke with a real xAI key (`tools/run_ai_live.sh test_xai_smoke_streaming`).
- TODO: Settings → Grok → paste key → new chat on `grok-4.6`; companion wizard lists Grok and Get API Key opens the xAI console.

The companion wizard still stores a pasted key on `effectiveVendor` until #754 lands; picking Grok there can hit that bug.

## Post-Merge
- If the website catalog is refreshed from this tree, publish `ai-models.json` (older apps skip the unknown `xAI` vendor string).

### DIFF
--- a/AILiveHarness/AILiveDriver.swift
+++ b/AILiveHarness/AILiveDriver.swift
@@ -515,6 +515,7 @@ final class AILiveDriver: NSObject, AITermControllerDelegate {
             case .gemini:    return "gemini"
             case .deepSeek:  return "deepseek"
             case .llama:     return "llama"
+            case .xAI:       return "xai"
             case .none:      return nil
             @unknown default: return nil
             }
@@ -528,6 +529,7 @@ final class AILiveDriver: NSObject, AITermControllerDelegate {
         if lower.hasPrefix("claude") { return "anthropic" }
         if lower.hasPrefix("gemini") { return "gemini" }
         if lower.hasPrefix("deepseek") { return "deepseek" }
+        if lower.hasPrefix("grok") { return "xai" }
         return nil
     }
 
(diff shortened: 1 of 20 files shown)


gnachman:
Thanks for putting this together, and apologies for the slow turnaround on getting to the review.

A couple of things to fix before this is ready to merge:

1. Image and PDF attachments are silently dropped for Grok. The attachment gates in `LLMProvider.swift` (`supportsInlineImageBlock` and `supportsPDFAttachments`) allowlist by host and never include `api.x.ai`, even though the PR adds the `hostIsXAIAIAPI` helper and wires it in everywhere else. Grok uses the same OpenAI-compatible `image_url` serializer, and grok-4.6/grok-4.3 are vision models (grok-4.3 also accepts PDFs), so adding `hostIsXAIAIAPI` to both gates is all that is needed.

2. Please rebase on the latest master. The branch currently conflicts: master removed the bundled Llama models from `ai-models.json` (Ollama models are now discovered from the local server instead of the catalog) and reworked the release notes. After rebasing, the catalog should end up at 42 models rather than 44, and the count assertion in `AIModelCatalogTest` should be updated to match.

Thanks again for the contribution.

PR #756, opened 2026-09-13
Add a Python API for native tab groups

jrejaud:
## Summary

iTerm2 3.7 has native tab groups — named, colored, collapsible runs of tabs — but no way to create or manage one programmatically. The only entry points are the menu actions, and “New Group…” opens a blocking modal name prompt. This adds a Python/WebSocket API surface for tab groups so scripts can build and manipulate them with no modal and no UI automation.

## API

A new `TabGroupRequest` / `TabGroupResponse` RPC and an `iterm2.TabGroup` Python class covering:

- **create** a group from one or more tabs with a name + color (no modal prompt)
- **assign** a tab to an existing group
- **remove** a tab from its group
- **rename** a group
- **set color**
- **collapse / expand**
- **list** all groups
- **read membership**

```python
group = await iterm2.TabGroup.async_create(
    connection, [tab], "phpvms", iterm2.Color(0x50, 0xAF, 0xBC))
await group.async_add_tab(other_tab)
await group.async_set_collapsed(True)
await group.async_set_name("phpvms-prod")
```

## Implementation

The group definition already rides its member tabs (`PTYTab.tabGroupID` / `tabGroupName` / `tabGroupColor` / `tabGroupCollapsed`), so the new `PseudoTerminal` methods — `-createTabGroupWithTabs:name:color:`, `-addTab:toExistingTabGroupWithID:`, `-removeTabFromItsGroup:`, `-setTabGroupName:forGroupID:` — are modal-free equivalents of the existing menu actions. They route through the same `-finalizeTabGroupMembershipChangeForTab:` / `-tabsDidReorder` / collapse machinery, so the contiguity and “the active tab is never inside a collapsed group” invariants are preserved. Colors round-trip through `RGBColor` via `-it_colorInDefaultColorSpace`. Groups are scoped to a single window; a create or assign that would span windows returns a specific status rather than silently splitting.

Proto (ObjC + Python) regenerated with `tools/build_proto.sh`.

## Tests

- `ModernTests/iTermTabGroupTests.swift` — invariant tests for the create / assign / remove / collapse scenarios the API relies on: scattered members compact into one contiguous block, the remainder stays contiguous after a removal, and a collapse finds a visible landing spot outside the group (and is refused when the group is the whole window).
- `tests/tab_group_api_test.py` — a smoke test that creates a named + colored group, moves three tabs into it, collapses it, renames it, and reads the membership back.

I developed this on macOS 15 / Xcode 16.4, where the current tree doesn’t build (it needs the macOS 26 SDK — `FoundationModels`, `ENOTCAPABLE`, Swift 6.3), so I couldn’t exercise the running app end to end locally. The unit tests pass and the change reuses the existing tab-group model rather than adding new state. Happy to adjust anything.


gnachman:
Thanks for this, Jordan, and sorry for the collision. A very similar tab-group Python API landed on master in a23c8c3 ("Add Python API for tab groups") shortly before this PR, so most of what's here is already available:

- Create: `Window.async_create_tab_group(name, tabs, color)`
- Membership: `Tab.async_add_to_tab_group()` / `Tab.async_remove_from_tab_group()`
- Mutators: `TabGroup.async_set_name` / `async_set_color` / `async_set_collapsed`
- Discovery: `Tab.tab_group` and `Window.tab_groups`, plus per-tab group id/name/color/collapsed carried in `ListSessionsResponse`

That version also handles the "handle keeps working after tabs are dragged to another window" case (it resolves the group's host window at call time), gates the feature behind a capability check, and gets membership for free from existing app state rather than a dedicated RPC.

Because this PR was branched from before a23c8c3, it re-adds `tabgroup.py` and a parallel `TabGroupRequest`/`TabGroupResponse` RPC, which now conflict with the shipped API rather than extend it. Rather than carry two surfaces for the same feature, I'm going to close this.

Two nice touches here that the landed API doesn't have and that I'd welcome as a small follow-up against current master:

1. An ordered member list on the group object (`group.tab_ids`), instead of filtering `window.tabs` by `tab.tab_group`.
2. Mutation responses that echo the updated group so a single handle stays fresh without re-fetching the whole app.

Both would be a small addition to the existing `iterm2.TabGroup`. Really appreciate the thorough tests and docs, and thank you for the contribution.

PR #757, opened 2026-09-13
Report OSC 8 hyperlinks in the API's CellStyle.url

anandghegde:
Fixes the bug reported in https://gitlab.com/gnachman/iterm2/-/issues/13003

## The bug

The Python API's `CellStyle.url` is always empty for OSC 8 hyperlinks, even though Cmd-click opens the link. `+[PTYSession protoStyleForCharacter:externalAttributes:]` builds the `ITMURL` and never attaches it to the style:

```objc
if (ea.url) {
    ITMURL *url = [[[ITMURL alloc] init] autorelease];
    url.URL = ea.url.url.absoluteString;
    url.identifier = ea.url.identifier;
}
```

So every cell in `GetBufferResponse` came back with no `url`, and `LineContents.style_at(x).url` in the Python library was always `None`. The reporter's diagnosis in the issue is right.

## Changes

- `sources/PTYSession/PTYSession.m`: assign the URL to the style (`style.URL = url;`).
- `iTerm2XCTests/PTYSessionTest.m`: add `testProtoStyleForHyperlink`, next to the existing `protoStyleForCharacter` tests. It builds an external attribute carrying an `iTermURL` with an identifier, and checks that `hasURL`, `URL.URL` and `URL.identifier` come through.
- `docs/notes-3.7.txt`: a Bug Fixes entry. The API has had this field since before 3.6.11, where the issue was reported.

## Testing

On macOS 26.6.2 (Apple Silicon) with Xcode 26.6, I ran:

```
xcodebuild test -project iTerm2.xcodeproj -scheme iTerm2Tests \
  -only-testing:iTerm2XCTests/PTYSessionTest/testProtoStyleForHyperlink
```

- **With the fix:** the test passes.
- **With the one-line fix removed:** the test fails. `hasURL` is false, and `URL.URL` and `URL.identifier` are `""`.

`testProtoStyleForDualModeIndexedForeground` in the same file failed on my machine both with and without this change (`style.fgRgb` is not nil, line 232). It looks unrelated to this PR, but I wanted to mention it.

To build locally, I rebuilt `SwiftyMarkdown.framework` and `Highlightr.framework` with `make paranoid-SwiftyMarkdown` and `make Highlightr`, because the committed builds wouldn't load under Xcode 26.6. I also updated `last-xcode-version` locally. None of those files are in this PR.

I wrote this PR with Claude Code (Opus), and I reviewed and verified the change.

https://claude.ai/code/session_01M8rNAK3JEZQNVxrnQ8SVGG


gnachman:
Thanks for catching this, and sorry for the duplicate work. I hit the same bug independently and landed the fix in 20b455b5a before seeing this, with a regression test in `ModernTests/PTYSessionProtoStyleURLTests.swift`. Your diagnosis was exactly right: the `ITMURL` was built and then never assigned to the style. Closing as already fixed, and the issue is resolved in the next build.

anandghegde:
No problem at all — the fix landing is the part that matters, and 20b455b5a with the regression test in `PTYSessionProtoStyleURLTests` is a better outcome than my patch, which had no test. Thanks for the quick turnaround.

PR #759, opened 2026-09-14
cc-status: support Codex CLI hooks

Todmy:
Codex CLI (0.154+) runs the same hooks as Claude Code, from `~/.codex/hooks.json`, with the same payload shape. With a few fixes the existing cc-status binary drives Session Status for a Codex tab too. One commit per fix, so any of them can go:

1. **Accept Codex payloads.** Codex validates SessionStart/Stop hook stdout as JSON and it2 prints "Session status updated.", so stdout goes to /dev/null. A cc-status built outside the bundle looks for it2 in the installed iTerm2.app (the two default locations only). `apply_patch` is summarised as `Edit: a, b` instead of leaking the diff, bounded like the other summaries.
2. **`--agent codex`.** Everything Codex-specific sits behind an explicit flag instead of a `turn_id` sniff, which was wrong on SessionStart and SessionEnd. The hooks.json entries have to invoke `cc-status --agent codex`; without the flag the binary behaves exactly as it does for Claude Code, and the tests assert that path makes no extra it2 calls.
3. **Count sub-agents without `background_tasks`.** Codex never sends the field, so a detached `spawn_agent` outliving the parent turn showed idle, and no later Stop fixed it. cc-status keeps the count from SubagentStart/SubagentStop, remembers in a session variable whether the root turn is still open, and reports idle from SubagentStop when the last child finishes after the parent stopped. Stop, Interrupt (Esc), SessionStart and SessionEnd clear the flag. Interrupt keeps the count on purpose: Codex interrupts only the root thread, so a spawned agent survives Esc and its own SubagentStop still arrives.
4. **OSC 9;4 progress ring.** Claude Code emits it itself, Codex doesn't. Codex 0.155 starts hooks with setsid (openai/codex#43876), so there is no /dev/tty inside the hook any more; cc-status walks up to the nearest ancestor that still has a tty and writes the ring there. Only with `--agent codex`.
5. **Tests.** `tests/run.py` runs the binary against a fake it2 that logs every call, with recorded payloads from Claude Code 2.1.201 and Codex 0.154. Hooks run detached under a pty the way Codex starts them, so the ring bytes are asserted as well, and stderr must be empty. Twelve scenarios, no iTerm2 needed: `swift build -c release && python3 tests/run.py`.

Tested on Codex 0.154 / iTerm2 3.7.1: status, detail, last message, detached and waited sub-agents, Esc, ring. The 0.155 setsid change is covered by the tests. `bin/cc-status` isn't rebuilt here.

Setup notes for whoever wires this up:

- `hooks.json` needs `"timeout": 3` on Interrupt and SessionEnd. Codex gives those two events one second by default and three at most, and the Interrupt path makes three it2 calls at roughly 200 ms each.
- Codex runs no hook it hasn't been told to trust, and an untrusted hooks.json fails silently. Codex asks on startup for new or changed hooks and stores the answer in `~/.codex/config.toml` as one `[hooks.state."<path to hooks.json>:<event>:0:0"] trusted_hash = "sha256:…"` entry per hook. Any edit to hooks.json, including adding `--agent codex`, invalidates the hash and has to be approved again.

Known gaps, both iTerm2-side and not in this PR:

- A turn that dies on an API error fires no event at all (Codex has neither StopFailure nor Notification), so the tab stays `working` until the next event. A sub-agent that dies the same way sends no SubagentStop, and the count then sticks until SessionEnd. The fix for both is a staleness guard in `iTermSessionTabStatus.swift` that demotes a `working` status with no event for N minutes. Claude Code has the same hole on abnormal turn ends.
- Codex launches the hooks for one event concurrently, so two SubagentStart in the same instant can lose an update on the count. An atomic `it2 session adjust-background-tasks --delta` would fix that and shorten the Interrupt path further.

This is hook-side only, nothing under `sources/` changes, so it doesn't overlap with #673 or #758. If Codex ever emits OSC 21337 itself (openai/codex#25879), the hooks can just be dropped from `hooks.json`.

Not in the PR: the `hooks.json` mapping (SessionStart, UserPromptSubmit, PreToolUse, PostToolUse, PermissionRequest, Stop, SubagentStart, SubagentStop, Interrupt, SessionEnd all point at `cc-status --agent codex`) and a `codex` tab icon. Happy to send either separately.


gnachman:
Thanks for this, and for splitting it one commit per fix. I built it and tested it live against Codex 0.154 and iTerm2 3.7.x. The happy path works: hooks fire in the right order, the detached sub-agent state machine tracks correctly against a real `it2`, and `tests/run.py` passes 11/11 with no extra `it2` calls on the Claude Code path, which is a nice guarantee to have asserted.

Two defects turned up in live testing that I do not think are visible from the fixtures, plus a set of smaller review findings.

## 1. A failed turn emits no event, so the tab sticks in `working` forever

This is the one that matters. Codex has no `StopFailure` and no `Notification`, and it turns out it fires nothing at all when a turn dies on an API error. Traced with a shim in front of `cc-status` (isolated `CODEX_HOME`, run out of credits):

```
16:26:54  SessionStart      turn_id= NO
16:26:54  UserPromptSubmit  turn_id= yes     -> status = working
16:27:18  SessionEnd        turn_id= NO
```

Nothing between `UserPromptSubmit` and process exit. Under `codex exec` the `SessionEnd` at the end rescues the status. In interactive Codex, which is the normal case, the process stays alive, no `SessionEnd` fires, and the tab sits in `working` until the user quits Codex or submits another prompt. Running out of credits or hitting a rate limit is ordinary enough that users will meet this quickly.

`cc-status` cannot fix this on its own, since there is no event to hook. What it needs is a staleness guard on the iTerm2 side: record when the status was last touched, and demote a `working` status that has had no event for N minutes. `iTermSessionTabStatus.swift` currently has no notion of time at all, so this is an additive change in one file. It is worth doing independently of Codex, because the Claude Code path has the same hole whenever a turn ends abnormally. `Notification(idle_prompt)` does not cover that case either.

Not asking you to build that here, but I do not think Codex support is complete without it, and it is worth calling out in the PR description so it is not a surprise.

## 2. `user.ccStatusTurnOpen` is left stale after a failed turn

Confirmed live. After the trace above, the session variables were:

```
background-tasks: 0
ccStatusTurnOpen: 1     <- stale
```

`SessionEnd` resets the count but not the turn-open flag. The next Codex session in that tab then starts with the flag set, so a detached sub-agent finishing after the parent stops reads `turnOpen == 1`, skips the idle report, and leaves the tab wrong. `StopFailure` has the same gap: it never clears the flag for Codex.

There is a trap in the obvious fix. I verified that `SessionStart` and `SessionEnd` both carry no `turn_id`, so `isCodex` is `false` on exactly the two events where you would want to clear the flag. Clearing unconditionally would add an `it2` round trip to Claude Code's session boundaries and break the no-extra-calls guarantee the tests assert.

## 3. Suggested fix for both: pass the agent explicitly

Rather than `let isCodex = json["turn_id"] != nil`, have `hooks.json` invoke `cc-status --agent codex`. Since iTerm2 writes that config, it costs nothing, and it makes detection correct on all twelve events instead of ten. Then clearing the flag in the `SessionStart`/`SessionEnd` case is one gated line, and the OSC 9;4 write below can be gated the same way.

Two further reasons to drop the heuristic: the docs confirm `turn_id` is only on turn-scoped events, so the current test is wrong by design rather than by accident, and `json["turn_id"] != nil` is also true for an explicit JSON `null`, which decodes to `NSNull`.

## 4. The OSC 9;4 write is not gated, so Claude Code behavior does change

The PR says Claude Code paths are unchanged, but the block at the end of `main.swift` runs for every event with a non-nil status. On `Stop` with background tasks still running, `cc-status` writes `9;4;3` while Claude Code writes `9;4;0` for the same turn end, and the ring lands in whichever won. `PermissionRequest` and `Notification` likewise force `9;4;0` over whatever Claude Code set. Gating on the agent, as the sub-agent logic already is, matches the stated intent.

Related: the ring goes to `/dev/tty` while the status is routed by `TERM_SESSION_ID`. Those can be different terminals under tmux, where `TERM_SESSION_ID` is inherited from whatever environment the tmux server started in.

## 5. The sub-agent count has a lost-update race

`SubagentStart` and `SubagentStop` do a read-modify-write across two separate `it2` processes, and the Codex docs are explicit that matching hooks for an event are launched concurrently. Reproduced with three concurrent `SubagentStart` payloads against a state-keeping fake `it2`:

```
stored count after 3 concurrent SubagentStart (expected 3): 1
```

The first `SubagentStop` then decrements to zero and reports idle while two agents are still running, which is the false-idle bug this code exists to prevent.

The clean fix is to make the arithmetic atomic on the iTerm2 side, something like `it2 session adjust-background-tasks -s <id> --delta +1`. That removes the race and cuts the `Interrupt` path from three subprocess launches to one, which matters for the next point.

## 6. `Interrupt` has a one second timeout

Codex defaults `Interrupt` and `SessionEnd` to 1 second, with a documented maximum of 3, while most events get 600. The `Interrupt` path currently makes three `it2` round trips at roughly 200 ms each by your own measure, plus process startup. If Codex kills the hook partway, the count and the turn-open flag are left inconsistent, and per defect 1 there is no later event to heal them. Worth at least a comment, and `hooks.json` should ask for `"timeout": 3` on those two events.

## 7. A missed `SubagentStop` sticks forever

The count is decremented only by `SubagentStop` and reset only by `SessionStart`/`SessionEnd`. If a sub-agent dies without one, the count stays at or above 1 for the rest of the session, every later `Stop` takes the `running > 0` branch, and `last_assistant_message` is never shown again. `Interrupt` reads the count but never clears it, even though Esc is the most likely way to kill a sub-agent mid-flight.

## 8. `it2` failures are silent and nothing retries them

`cc-status` prints to stderr and exits 0 when `it2` fails, and Codex discards stderr. I hit this for real: a Codex tab owned by a different iTerm2 instance than the one `it2` reaches gave

```
Error: Set status failed: No such session
cc-status: it2 exited 1 for Stop
exit code: 0
```

and the tab silently kept a stale status. That is partly my unusual setup, several iTerm2 instances running at once, but the swallow is general. Under Claude Code the next `idle_prompt` retries. Under Codex nothing does, so a single failed call is permanent.

Cheapest guard is in the harness: `tests/run.py` asserts only on `returncode`, `stdout` and the call log, so add `assert result.stderr == b""`. That one line turns this class of failure into a test failure.

## 9. Smaller things

- `tests/run.py` uses `start_new_session=True`, which removes the controlling terminal, so the `/dev/tty` write is skipped in every scenario and commit 3 has no coverage at all.
- `apply_patch` summaries are unbounded. Every other branch of `toolCallSummary` is bounded (`condense` for Bash, `shortPath` for a single path), but `"Edit: " + files.map { shortPath($0) }.joined(...)` is not, so a patch touching 30 files produces a detail string over a thousand characters that goes straight to `it2 --detail`. Wrapping it in `condense(...)` would match the rest.
- `resolveIt2`'s installed-app fallback only checks two hardcoded paths named `iTerm.app`. A user with iTerm2 in `/Applications/Utilities`, under a renamed bundle, or at a non-standard cask target falls through to the `/usr/bin/env it2` case the comment above it calls unreliable. At minimum worth a comment noting the limit.

## 10. Not a defect, but missing from the writeup: hook trust

Codex 0.154 will not run hooks it has not been told to trust. Trust is persisted in `config.toml` as per-hook SHA-256 entries:

```toml
[hooks.state."/Users/…/.codex/hooks.json:stop:0:0"]
trusted_hash = "sha256:…"
```

An untrusted `hooks.json` runs nothing, silently and with no error, which cost me a confusing round of debugging. Any future iTerm2 install flow for Codex has to account for it, and it is worth a line in the PR description.

## What I am asking for

Blocking, in my view: the turn-open flag reset (2), gating the OSC 9;4 write (4), and the stderr assertion (8), since all three are small.

Strongly preferred: the explicit `--agent codex` flag (3), because it makes 2 clean rather than awkward and retires a heuristic that is already wrong on two events.

Follow-up, not this PR: the atomic count (5) and the staleness guard for defect 1, both of which are iTerm2-side and both of which I think are worth doing on their own merits.

Happy to take any of the iTerm2-side pieces myself if you would rather keep this PR hook-only.

Todmy:
Thanks for testing this against a real Codex session. I’ll keep this PR focused and small, with one commit per hook-side fix. I’ll try to close out the remaining fixes myself.

Todmy:
Thanks, the shim trace saved me a day. All pushed, one commit per point, description rewritten.

Done: 2, 3, 4, 8, plus 6, 9, 10. One surprise: Codex 0.155 starts hooks with setsid, so the ring now goes to the nearest ancestor tty. Tests run the hook the same way.

On 7: Esc doesn't kill sub-agents, only the root thread gets the interrupt. So the count stays, with a comment. The sticky case is real, but it's hole 1 again: a dead agent sends nothing.

One question. The app-server daemon knows the truth: `thread/list` gives `parentThreadId` and live `status` per thread. cc-status could ask it instead of counting, which would close 5 and 7. But the TUI only attaches to a daemon that's already running, so the count would stay as fallback. Worth it, or keep the hook dumb and let the guard cover everything?

Either way, this PR stays hook-only and the staleness guard is yours if you still want it.

PR #760, opened 2026-09-17
Use macOS trust store for uv downloads

rbferreira:
On macOS machines behind a corporate proxy that does TLS inspection, `uv python install` fails with an `UnknownIssuer` error while provisioning the Python runtime. The issue comes up in enterprise environments where a custom root CA is installed in the system Keychain but not in the Mozilla bundle that uv ships with.

The fix is a single environment variable: `UV_SYSTEM_CERTS=1`, which tells uv to use the platform native trust store (macOS Keychain) instead of its bundled Mozilla roots. The Keychain has everything Mozilla has plus whatever custom CAs the user or their employer installed. No downside for non-enterprise users — the Mozilla bundle still works exactly as before.

Added it to `iTermUvCommand.provisionEnvironment()`, which is the single source of truth for the environment every `uv` subprocess gets during provisioning (python downloads, pip installs, everything). Updated the unit test to assert it is there so it does not accidentally regress.

Fixes https://gitlab.com/gnachman/iterm2/-/issues/13059


rbferreira:
I see this landed in fcc31e5 — closing in favor of yours.

Appreciated the note on UV_NATIVE_TLS vs UV_SYSTEM_CERTS. My PR didn't address that distinction; uv 0.12 warning on stderr would've surfaced in the error alerts and that's worth calling out explicitly.

Thanks for the fast turnaround on #13059.

PR #761, opened 2026-09-20
Let each profile choose the sound its bell makes

anandghegde:
Implements [issue #2635, “Alternate Bell Sound?”](https://gitlab.com/gnachman/iterm2/-/issues/2635) — a per-profile bell sound.

## Why

The audible bell has always been `NSBeep()`, so every session rings identically. When a bell comes from a background tab there is nothing in the sound that says which one rang. The issue asks for the setting to be per profile, which is what this does.

## The stored value

A profile keeps a single string under a new `KEY_BELL_SOUND`:

| Value | Meaning |
| --- | --- |
| `""` | the system alert sound — what the bell has always played, and still the default |
| `"Glass"` | a sound installed in one of the Sounds folders, looked up by name |
| `"/path/to/chime.m4a"` | any audio file |

A file the user picks through **Other…** is stored by *name* when it lives in a Sounds folder and by path otherwise, so a profile that travels to another Mac still finds its sound instead of pointing at a home directory that no longer exists.

## Behavior

- Settings > Profiles > Terminal gains a **Bell sound** popup: System Alert Sound, then the sounds installed on the Mac, then **Other…** for any audio file.
- Picking a sound previews it, so they can be told apart without leaving Settings.
- The popup and its label dim when **Silence bell** is checked.
- A profile can outlive the file it names. When that happens the bell rings the system alert sound rather than going silent (with a `DLog` explaining why), and Settings shows the stale value as its own item with a warning icon and a tooltip instead of quietly reading as the default.
- Sessions share one player, so a sound is decoded once rather than once per session, and the cache is capped. A bell that arrives while the last one is still playing restarts it, so a burst of bells sounds like a burst instead of one long note.

## Notes

- New file `sources/Infrastructure/iTermBellSound.swift` (the model plus the shared player), registered with `tools/add_file_to_xcodeproj.rb`.
- `PTYSession` reads `KEY_BELL_SOUND` straight from `self.profile` on the main thread rather than adding a property to `VT100ScreenConfiguration`, matching what `-setBell:` already does with `KEY_SEND_BELL_ALERT`.
- `Bell Sound` can hold a path, so it gets a redaction rule in `tools/sanitize_iterm2_plist.py` (verified with `--dry-run`).
- `ModernTests/BellSoundTests.swift` covers the classification, the name-vs-path choice, display names, and playability. All 11 pass.
- Verified in a running Development build: the popup lists the 14 system sounds plus Other…, selecting Glass writes `"Bell Sound" = Glass` to the profile and previews it, and checking Silence bell dims both the popup and its label.

## Testing

- `tools/build.sh` — clean, no new warnings.
- `tools/run_tests.expect ModernTests/BellSoundTests` — 11/11 pass.
- `tools/run_tests.expect -parallel` — the four failures it reports (`PSMTabGroupDragTests`, `MenuItemWalkerTests`, `iTermFunctionCallSuggesterTests`) all pass when run on their own, so they look like existing parallel-run flakes rather than anything this touches.


sources/Infrastructure/iTermBellSound.swift
```diff
@@ -0,0 +1,170 @@
...
+
+    // The value to store for a file the user chose. A file sitting in a Sounds folder is
+    // stored by name because NSSound can find it again from the name alone, and a name
+    // survives a move to a Mac whose home directory is somewhere else.
+    @objc(profileValueForURL:)
+    static func profileValue(for url: URL) -> String {
+        let standardized = url.resolvingSymlinksInPath().standardizedFileURL
+        let name = standardized.deletingPathExtension().lastPathComponent
+        let parent = standardized.deletingLastPathComponent().path
+        if soundDirectories.contains(parent) && NSSound(named: name) != nil {
```
gnachman, line 57: **A bare name can resolve to a different file than the one that was picked.**

The name is stored whenever the chosen file's parent is a Sounds folder and `NSSound(named:)` resolves that name at all, but nothing checks that it resolves back to *this* file. `NSSound` searches `~/Library/Sounds` first.

So if someone has `~/Library/Sounds/Glass.aiff` and then uses Other... to pick `/System/Library/Sounds/Glass.aiff`, the profile stores `"Glass"` and the bell plays the home-directory file instead. Settings shows "Glass" with no warning, so the wrong sound looks correct.

Comparing the resolved `NSSound` back against the chosen URL, or only storing the name when no higher-priority folder shadows it, would close this.

sources/Infrastructure/iTermBellSound.swift
```diff
@@ -0,0 +1,170 @@
...
+            loaded = NSSound(named: name)
+        case .file(let url):
+            loaded = NSSound(contentsOf: url, byReference: false)
+        }
+        guard let loaded else {
+            return nil
+        }
+        cache[profileValue] = loaded
+        leastRecentlyUsedFirst.append(profileValue)
+        while leastRecentlyUsedFirst.count > Self.maximumCachedSounds {
```
gnachman, line 165: **`leastRecentlyUsedFirst` is not actually an LRU, and eviction can cut off a playing sound.**

A cache *hit* at line 148 returns early without moving the key to the end, so this array is insertion-ordered only. With more than 16 distinct bell sounds in play, the oldest-inserted entry is evicted even if it is the one ringing constantly, and it gets re-decoded from disk on the main thread at the next bell.

Separately, `cache.removeValue(forKey:)` can drop the only reference to an `NSSound` that is currently playing, cutting a session's bell off mid-sound.

Moving the key on hit fixes the first; skipping eviction of `isPlaying` sounds fixes the second.

sources/Infrastructure/iTermBellSound.swift
```diff
@@ -0,0 +1,170 @@
...
+        // burst of bells sounds like a burst instead of one long note.
+        if sound.isPlaying {
+            sound.stop()
+        }
+        sound.play()
+    }
+
+    private func sound(forProfileValue profileValue: String,
+                       source: iTermBellSound.Source) -> NSSound? {
+        if let cached = cache[profileValue] {
```
gnachman, line 148: **File-backed sounds are cached by path with no invalidation.**

If someone replaces `~/chime.m4a` with different audio, or deletes it, a session that has already rung keeps playing the old decoded audio for the rest of the app's lifetime. Meanwhile Settings re-checks with `isPlayable` on every rebuild, so it shows the warning icon for the now-missing file. The two disagree about the same profile.

gnachman (commented):
Code review of the bell sound change. One correctness issue worth fixing (the bare-name shadowing at `iTermBellSound.swift:57`), plus some robustness and hygiene notes.

For the record, these all checked out and need no attention: `KEY_BELL_SOUND` is registered in all three required `iTermProfilePreferences` places; `kPreferenceInfoTypeStringPopup` does use `representedObject` on both the read and write path, so `customSettingChangedHandler` plus `onUpdate` returning `NO` is the right shape; `rebuildBellSounds` runs before `defineControl`, so the initial `selectItemAtIndex:` finds an item; the `NSNull` entries in `terminalOnlyKeys` are removed rather than stored, so nothing non-string reaches the Swift bridge; the XIB's +28pt shift hits exactly the subviews at y>=323 with no overlap against the y=302 row; and the sanitizer entry landed in the right ruleset.

Also checked: `NSSound.play()` is asynchronous, so the bell does not hold the main thread for the duration of the sound. Only the first load of each sound is synchronous.

sources/Infrastructure/iTermBellSound.swift
```diff
@@ -0,0 +1,170 @@
...
+    // Settings rather than silently ringing the default bell.
+    @objc(profileValueIsPlayable:)
+    static func isPlayable(profileValue: String?) -> Bool {
+        switch source(of: profileValue) {
+        case .systemAlert:
+            return true
+        case .named(let name):
+            return NSSound(named: name) != nil
+        case .file(let url):
+            return FileManager.default.isReadableFile(atPath: url.path)
```
gnachman, line 105: **`isPlayable` only checks readability, but the player also needs decodability.**

The player uses `NSSound(contentsOf:byReference:false)`, which fails on a readable file that is not decodable audio. A profile pointing at a file that was later truncated, corrupted, or replaced by a non-audio file shows no warning icon in Settings yet silently falls back to the system beep, which is exactly the "quietly reading as the default" case the warning icon exists to prevent.

sources/Settings/ProfilesTerminalPreferencesViewController.m
```diff
@@ -443,6 +481,83 @@ - (void)updateCustomLocaleControls {
...
+    NSMenu *const menu = _bellSound.menu;
+    [menu removeAllItems];
+
+    NSMenuItem *systemAlert = [[NSMenuItem alloc] init];
+    systemAlert.title = NSLocalizedStringWithDefaultValue(@"ProfilesTerminal.SystemAlertSound", nil, [NSBundle mainBundle], @"System Alert Sound", @"Name in the bell sound menu of the Mac’s own alert sound, which is what the bell plays by default");
+    systemAlert.representedObject = @"";
+    [menu addItem:systemAlert];
+    [menu addItem:[NSMenuItem separatorItem]];
+
+    for (NSString *name in [iTermBellSound installedSoundNames]) {
```
gnachman, line 497: **This scan re-runs on every profile reload, on the main thread.**

`rebuildBellSounds` is wired into `info.onUpdate`, so it runs on every `reloadProfile`: profile selection change, `kReloadAllProfiles`, `kSessionProfileDidChange`. Each run synchronously enumerates all four Sounds directories, including `/Network/Library/Sounds`, an autofs trigger that can block for seconds when a network automount is unresponsive.

It also calls `NSSound(named:)` on every file found. `NSSound`'s name table fully loads and permanently retains each sound it creates, so a large audio file dropped in `~/Library/Sounds` is decoded into memory just by opening Profiles > Terminal, and the player's 16-sound cap does nothing to bound it.

Caching the scan and invalidating on demand would fix both the hang risk and the retention.

sources/Settings/ProfilesTerminalPreferencesViewController.m
```diff
@@ -443,6 +481,83 @@ - (void)updateCustomLocaleControls {
...
 
+#pragma mark - Bell Sound
+
+- (void)rebuildBellSounds {
+    NSString *const currentValue = [self stringForKey:KEY_BELL_SOUND] ?: @"";
+    NSMenu *const menu = _bellSound.menu;
+    [menu removeAllItems];
+
+    NSMenuItem *systemAlert = [[NSMenuItem alloc] init];
+    systemAlert.title = NSLocalizedStringWithDefaultValue(@"ProfilesTerminal.SystemAlertSound", nil, [NSBundle mainBundle], @"System Alert Sound", @"Name in the bell sound menu of the Mac’s own alert sound, which is what the bell plays by default");
```
gnachman, line 492: **New localizable strings have not been added to the catalog.**

This adds `ProfilesTerminal.SystemAlertSound`, `ProfilesTerminal.MissingBellSound`, `ProfilesTerminal.OtherBellSound`, and `ProfilesTerminal.ChooseBellSound` (line 541), but `xcstrings/Localizable.xcstrings` is untouched. Per `docs/localization.md` and recent history (29fa4d33e, 941c56834), new catalog strings get extracted and translated in the same commit. As-is these four ship English-only in every localized build.

sources/Infrastructure/iTermBellSound.swift
anandghegde: Fixed in c24cce6bf. The name is stored only if it leads back to this exact file: no earlier Sounds folder (or the app bundle) has that name, and no other file in the same folder has it with a different extension. Otherwise the path is stored. The check runs against a fresh scan of the Sounds folders taken after the file is picked. Tests cover shadowing by an earlier folder and by a second extension in the same folder.

sources/Infrastructure/iTermBellSound.swift
anandghegde: Fixed in c24cce6bf. A cache hit now moves the key to the most-recent end, so eviction takes the least recently used sound. Eviction skips any sound that `isPlaying`, and the one just loaded is never a candidate. If everything else is playing, the cache goes over the cap until those sounds finish.

sources/Infrastructure/iTermBellSound.swift
anandghegde: Fixed in c24cce6bf. Each cached file sound stores the file's device, inode, size, mtime and ctime when it's loaded. Every lookup stats the file: if it's gone the entry is dropped and the bell falls back to the system alert sound, and if it changed the sound is reloaded. Settings now asks the player whether a value can play (see the `isPlayable` thread), so the two can't disagree. `test_fileChangedAfterLoading_isReloaded` covers replace and delete.

sources/Infrastructure/iTermBellSound.swift
anandghegde: Fixed in c24cce6bf. `isPlayable` now asks the player, which loads the file with the same `NSSound(contentsOf:byReference:false)` call the bell uses and caches the result. A file that exists but isn't audio now shows the warning icon. I changed the tooltip to “This sound can’t be played…” since the sound may be there but unplayable. Added a test for a readable non-audio file.

sources/Settings/ProfilesTerminalPreferencesViewController.m
anandghegde: Fixed in c24cce6bf. The scan runs on a background queue, and its result is cached. `rebuildBellSounds` reads the cache, so reloading a profile no longer touches the disk. A scan runs when the tab is set up, in `viewWillAppear`, and when a file is picked with Other…. If the set of sounds changed, a notification rebuilds the menu. The scan now decides playability from the file extension against `NSSound.soundUnfilteredTypes` and loads nothing, so nothing goes into NSSound's name table. I kept /Network/Library/Sounds in the scan because NSSound searches it, and the shadowing check needs to see it. Since it's off the main thread now, a hung automount only delays the list.

sources/Settings/ProfilesTerminalPreferencesViewController.m
anandghegde: Fixed in c24cce6bf. I ran `tools/extract_strings.sh` and translated the four keys into de, ja, pt-BR, ru and zh-Hans in the same commit. The key for the tooltip is now `ProfilesTerminal.UnplayableBellSound` because its wording changed. Extraction on this branch also picked up about 20 upstream keys that have no translations yet. I left those out so the commit only adds these four. `check_untranslated_strings.py` passes, and re-running extraction doesn't change these entries.

PR #762, opened 2026-09-20
Add project sidebar and tabs for machine-wide coding harnesses

ljack:
Adds an opt-in left sidebar for coding harnesses across the machine, grouped by directory. Select a directory to show only its project tabs. Each attached harness opens in its own tab; switching projects keeps all processes running. Enable the sidebar in Advanced Settings by searching for “harnesses”.

- Click a tmux harness, or use **Attach in Project Tab**, to attach its existing pane. Repeated attachment focuses the already-open tab. A directory’s context menu can attach all its tmux harnesses.
- Click an external native Codex, Claude, or Antigravity harness to choose **Transfer to Tmux Tab** or **Transfer to Native Tab**. Exact conversation IDs come from process-linked open transcripts or Claude PID records, with no manual ID/handoff entry. An automatic recovery record is saved before resuming in the project directory. Ambiguous or unavailable identities are refused. The dialog explicitly says the original will close and asks users to transfer only when idle. After saving the recovery record, transfer revalidates the exact conversation and process identities, sends SIGTERM only to the original harness and matching launcher children, and waits up to 15 seconds for exit before opening the replacement. It never force-kills or signals a shell/process group; failure prevents replacement launch. Selecting a native harness keeps its project/row selected instead of activating the original terminal.
- Command-Option-1 through 9 selects directories, with shortcuts shown on directory headings. Within a selected project, Command-1 through 9 and normal/MRU tab cycling stay within its tabs. Tab shortcut labels restart at 1 per project; Command-9 selects its last tab. Project shortcuts dispatch before the default global window/pane/tab switchers. **All projects** restores the full tab bar. Drag the sidebar’s right edge to resize it; its width is remembered locally.

Discovery includes this instance’s sessions, terminal-backed native processes owned by the current user, and tmux-backed harnesses even when detached or attached in another terminal. Known executable names and Node launcher scripts identify Codex, Claude, Gemini, Antigravity, OpenCode, and Aider. Launcher/native child pairs on the same terminal collapse into one run. Host/user identity stays separate and each run keeps its first observed directory. The scanner excludes pipe-only workers and MCP servers and does not retain full process environments. When launchers clear TMUX/TMUX_PANE, discovery queries the ancestor tmux server’s pane inventory through its UNIX socket and matches pane PIDs. Metadata queries force UTF-8 output so GUI launches without a UTF-8 locale preserve field separators.

The top tab bar sits beside the sidebar, over the terminal column, including on macOS versions that otherwise put tabs in the titlebar accessory. The project filter hides tab cells and group chips without removing tabs or changing saved tab-group collapse state. Existing tabs with mixed-directory split panes remain intact and appear in each matching project. Project selection and membership are currently transient; discovery rebuilds the grouping when the sidebar is recreated. Native resume uses exact process-linked conversation identity rather than guessing the latest history entry. Unsupported providers do not offer a guessed resume.

Validation:
- 102 harness and tab-bar regression tests passed, including project filtering, preserved group collapse state, command quoting, resume identity validation, independent-PTY discovery, and tab-group drag/drop tests. Two drag cases failed in the first combined run; both passed individually and the complete 102-test rerun passed.
- 44 focused discovery and layout tests cover an isolated tmux process launched with env -i, pane parsing, and requested sidebar widths while preserving terminal space.
- Development build checked separately. Interactive visual and real harness-resume verification remain pending.

Defaults remain unchanged. Machine-wide discovery is local; remote harnesses depend on existing iTerm2/session or native tmux metadata. Historical launch directories cannot be recovered for runs first observed midway through execution.

Latest validation: 15 harness tests passed, including process-linked transcript identity, ambiguity rejection, reused-PID/self-signal refusal, and graceful exit/timeout gating with an isolated subprocess. Development build passed. Real interactive transfer remains a manual acceptance check.


ljack:
iterm2-harness-sidebar
iterm2-advanced-settings

PR #763, opened 2026-09-21
Remove obsolete scroll-view dismembering workaround

nlutsenko:
With the experimental `DismemberScrollView` setting enabled, scrolling stopped working across terminal tabs on iTerm2 3.7.2 running macOS 27.0 (26A428). Setting it to **No** restored scrolling. 

The workaround targets macOS 10.14; iTerm2 now requires macOS 13 or later.

- Remove the setting and its scroller reparenting, positioning, and whole-scroll-view transparency paths
- Keep the existing default behavior: scrollers remain inside their scroll view.
- Previously saved `DismemberScrollView=true` values are ignored.
- New tests (they are somewhat sloppy, happy to take it in any direction reviewers would want)


gnachman:
Merged, thank you @nlutsenko! Nice piece of archaeology: `dismemberScrollView` shipped defaulting to `NO`, so every branch this removes was already the dead one, and the workaround it guarded was aimed at macOS 10.14 when the deployment target is now 13. I read each hunk and the removal is behavior-preserving for the shipped configuration: `useLayerForBetterPerformance` returned `YES` by default, so the unconditional layer assignment in `-setSuppressDrawing:` matches what everyone was already getting, and the three `PTYScrollView.shouldDismember` branches in `PTYSession` collapse to the same `updateWrapperAlphaForMetalEnabled:` call the other four call sites made unconditionally. No dangling references left behind, and `SessionView` is the only `PTYScrollerDelegate` conformer, so trimming the protocol is free.

I built the branch and the new tests pass clean with no warnings. They are not sloppy at all, and the containment assertion is the right invariant to pin. One thing worth knowing: since the `DEFINE_BOOL` is gone, `loadAdvancedSettingsFromUserDefaults()` never reads the `DismemberScrollView` key any more, so `testScrollerRemainsContainedWithLegacyPreferenceEnabled` runs the same code path as its sibling. It still earns its keep as a regression test if anyone ever reintroduces both the setting and the reparenting, so I kept it as is.

I picked up two small leftovers on top: the orphaned `AdvancedSetting.dismemberScrollView` entry in the string catalog, which would have gone stale on the next extraction run and kept translators busy with a setting that no longer exists, and a release note, since anyone who had this turned on gets a real behavior change.

I have added you to the contributors list on the iTerm2 credits wiki.

